PR #5, opened 2016-07-18
Master

laughing099:
Hi, your works are great. 
I am a  Master in china, i Want a State Machine/hybrid automata described in Modelica 3.3.
so, when it will be released?
I am looking forward to your reply, thanks.


dietmarw:
This is just a mirror that does not accept any pull requests.

Issue #2157, opened 2017-02-06
ComplexBlocks.Sources.LogFrequencySweep is not complex


JoeRiel:
Why is the LogFrequencySweep in the ComplexBlocks package?  Its output is real. Shouldn't it be in Blocks.Sources?

tbeu:
This requires a conversion script.

beutlich:
This requires a conversion script.

See #1850 for the corresponding ticket that added this block.

AHaumer:
The block is mainly used in the ComplexBlock - Examples.

AHaumer:
It should go rather to a Utilities package of ComplexBlocks ...

AHaumer:
As far as I remmebr, the block can't be moved now due to backwards compatibility,

beutlich:
> As far as I remmebr, the block can't be moved now due to backwards compatibility,

Exactly. The milestone is set to MSL_next-MAJOR-version for that reason.

christiankral:
I will move `ComplexBlocks.Sources.LogFrequencySweep` to `Blocks.Source`, as it is a signal source: this is where it belongs to.

beutlich:
Havea look at https://github.com/modelica/ModelicaStandardLibrary/pull/2158/files. Might save you some typing.

PR #2158, opened 2017-02-07
Move LogFrequencySweep from ComplexBlocks to Blocks

beutlich:
refs #2157: Move LogFrequencySweep from ComplexBlocks to Blocks

### DIFF
--- a/Modelica/Blocks/Sources.mo
+++ b/Modelica/Blocks/Sources.mo
@@ -1964,6 +1964,23 @@ a flange according to a given acceleration.
 </html>"));
   end KinematicPTP2;
 
+  block LogFrequencySweep "Logarithmic frequency sweep"
+    extends Modelica.Blocks.Interfaces.SO;
+    import Modelica.Constants.eps;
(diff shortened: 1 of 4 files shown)


tbeu:
This only was a test of the GitHub Reviewer feature.

MartinOtter:
This is a non-backwards compatible change (moving a block from one library to another one). This means that existing models using this block will no longer work. As long as we have not yet decided that conversion scripts can be used, we should not perform such changes (a pre-requisite is that the tool vendors support the conversion annotations introduced in Modelica 3.4).

beutlich:
> This is a non-backwards compatible change (moving a block from one library to another one). 

For that reason, the milestone is set to [MSL_next-MAJOR-version](https://github.com/modelica/Modelica/milestone/4), which is defined as the next (**not yet specified**) major version which is non-backward compatible and (might) need a conversion script.

beutlich:
Clsoing after @AHaumer's https://github.com/modelica/Modelica/issues/2157#issuecomment-355080017.

Issue #2079, opened 2017-01-15
Parameter enableNoise depends per default on settings in GlobalSeed


modelica-trac-importer:
**Reported by tobolar on 13 Oct 2016 08:59 UTC**
Since the availibility of multiple noise signals generally applies for a complete model where they are used, the parameter `enableNoise` in Modelica.Blocks.Interfaces.PartialNoise should be per default dependent on `globalSeed.enableNoise` from Modelica.Blocks.Noise.GlobalSeed

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/2079

modelica-trac-importer:
**Modified by beutlich on 13 Oct 2016 20:40 UTC**

beutlich:
@tobolar Can you please provide a PR for this enhancement. Thank you very much.

tobolar:
Sorry, but what does PR stand for? [http://www.abbreviations.com/PR](http://www.abbreviations.com/PR)

xie-dongping:
@tobolar

PR stands for pull request in GitHub.
https://help.github.com/articles/about-pull-requests/

tobolar:
I see. Then I will prepare PR.

beutlich:
Resolved both in master and maint/3.2.2.

PR #2160, opened 2017-02-08
Fixed issue #2079 - enableNoise in PartialNoise depends on GlobalSeed

tobolar:
Blocks.Interfaces.PartialNoise: parameter enableNoise depends per default on enableNoise from Modelica.Blocks.Noise.GlobalSeed (issue #2079)


beutlich:
Thank you!

PR #2163, opened 2017-02-09
Fix icon for Electrical.Digital

sjoelund:
Current icon has the lines in the wrong part of the annotation.

The error was found when making OMEdit report failure to interpret icon annotations. The change was tested in OMEdit and looks reasonable (please review; I can add it to maint/3.2.2 once accepted into the master). The source of the error seems to be bf5ac67e (@otter), which says it added an icon for Digital (but there seems to have already been an icon there).


sjoelund:
Actually, there seems to be one line in there that I am unsure if it belongs or not. I'll the icon and one without the line I am uncertain about.

As it looks in the PR:
digital-all

What it should look like?
digital-cleaned

jespermattsson:
I think the extra line might be meant to represent a digital signal.

tbeu:
Point to @MartinOtter instead of otter.

beutlich:
Please remove that ternary signal line added by bf5ac67e02868aa348d402527ec13f395399dc99 and reveal the block icon.

sjoelund:
So the second image, then? Basically reverting bf5ac67 and instead of adding the signal fixing the error in the icon.

beutlich:
LGTM.

beutlich:
Also cherry-picked in maint/3.2.2 branch.

Issue #386, opened 2017-01-14
Animation missing in MultiBody.Forces.SpringDamperSeries


modelica-trac-importer:
**Reported by otter on 22 Jul 2010 10:19 UTC**
The component MultiBody.Forces.SpringDamperSeries does not yet have animation 

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/386

modelica-trac-importer:
**Modified by otter on 5 Aug 2010 21:37 UTC**

modelica-trac-importer:
**Comment by dietmarw on 6 Sep 2012 11:57 UTC**
These tickets don't rely on conversion scripts, hence rescheduling for the next major release

modelica-trac-importer:
**Comment by beutlich on 29 Apr 2014 06:29 UTC**
@dietmarw: Should Milestone be set back to MSL_withConversionScript?

modelica-trac-importer:
**Changelog removed by beutlich on 29 Apr 2014 06:29 UTC**

modelica-trac-importer:
**Comment by otter on 13 Dec 2015 08:05 UTC**
Do not yet have time to provide an animation. Feel free to propose an animation object and send a patch

beutlich:
> Feel free to propose an animation object and send a patch

@MartinOtter Done by #2183. Please check the proposed animation enhancements for  SpringDamperSeries and SpringDamperParallel.

beutlich:
Resolved in master by #2183.

PR #2183, opened 2017-02-20
Add animation for SpringDamperSeries and SpringDamperParallel

beutlich:
See #386 for the corresponding issue.

@MartinOtter Please add your review.


beutlich:
@MartinOtter What do you think about the added visualization components?

beutlich:
@MartinOtter When will you be able to deal with it?

Issue #2000, opened 2017-01-15
Modelica.Fluid.Examples.DrumBoiler: incorrect unit on output p_S


modelica-trac-importer:
**Reported by dag on 24 May 2016 11:25 UTC**
The top-level output `p_S` in Modelica.Fluid.Examples.DrumBoiler.DrumBoiler has incorrect unit. The sensor value (in Pa) is converted to Bar using a gain-block. However, the unit is not adjusted so the unit remains Pa although the value is scaled.

The minimum update would be to add the appropriate unit on the output of `Pa2bar `with a modifier

```
y(unit="bar")
```
I think an even better change would be to delete the components `Pa2bar `and `K2degC` altogether, because the outputs from the corresponding sensors already have the displayUnits we want.

 

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/2000

modelica-trac-importer:
**Comment by rfranke on 18 Jun 2016 10:06 UTC**
Good point. It is not an option to just remove the conversion blocks, because the global outputs are intended for model export, where the actual values must be scaled appropriately, not just the display. 

Think of the use case that one wants to build a controller around the DrumBoiler model, e.g. using Modelica.Blocks. It is very unlikely that the controller signals would be scaled in Pascal or Kelvin anywhere in the world.

The dedicated Celsius.FromKelvin K2degC is working correctly. We might add dedicted `toBar` and `fromMW` blocks to MSL. Where?

Alternatively: Modelica sensors should support unit conversion of the signal values that can be modified for the application at hand. An imperial application would most likely want to use psi and degF.

modelica-trac-importer:
**Modified by rfranke on 18 Jun 2016 10:15 UTC**

beutlich:
@rfranke Would Modelica.SIunits.Conversions.to_bar be an option?

DagBruck:
@beutlich The minimum change that fixes the problem is to add the `y(unit="bar")` modifier.

henrikt-ma:
Hej!

Tack för senast.

Märkte du att Hans placerade dig i gruppen som skulle styra upp tolkningen av arrowSize?
https://trac.modelica.org/Modelica/ticket/1894 <https://trac.modelica.org/Modelica/ticket/1894>

[...]

PR #2197, opened 2017-03-03
Fix unit of Real output signal p_S

beutlich:
* Fix unit of Real output signal p_S
* Improve support for tools that not not feature unit propagation by explicitly providing units for Real outputs
* See #2000


beutlich:
@rfranke What do you think about the proposed unit changes?

beutlich:
@rfranke Now, that one month passed, what do you think about the proposed unit changes?

PR #2213, opened 2017-03-21
Remove pretected element from record

miscco:
The record kc_flatTube_IN_con contained the protected element 

```
        type TYP =
            Modelica.Fluid.Dissipation.Utilities.Types.HTXGeometry_flatTubes;
```
This is illegal according to Modelica specification section 4.6. as records can only contain public sections. As TYP is never used in the recrod simply delete it.


Modelica/Fluid/Dissipation.mo
```diff
@@ -1136,10 +1136,6 @@ Calculation of the mean convective heat transfer coefficient <b> kc </b> for the
         "Input record for function kc_flatTube and kc_flatTube_KC"
         extends Modelica.Icons.Record;
 
-      protected
-        type TYP =
-            Modelica.Fluid.Dissipation.Utilities.Types.HTXGeometry_flatTubes;
-
       public
```
sjoelund: The public keyword also seems unnecessary here.

beutlich:
For me it looks rather odd that this bug is not present in [FluidDissipation >= v1.1.7](https://github.com/ClaRaLibrary/FluidDissipation/blob/master/HeatTransfer/HeatExchanger/kc_flatTube_IN_con.mo) which means, the two repositories ([Modelica.Fluid.Dissipation](https://github.com/modelica/Modelica/blob/master/Modelica/Fluid/Dissipation.mo) and [FluidDissipation](https://github.com/ClaRaLibrary/FluidDissipation)) are already out of synchronized maintenance.

Since we do not have a modularized MSL it would be good to know if it is intended that bug fixes on FluidDissipation are ported to Modelica.Fluid.Dissipation and vice versa. @wischhusen Can yo please comment on the maintenance strategy. Thanks a lot!

Modelica/Fluid/Dissipation.mo
miscco: That is correct, but I wanted to just have the minimal diff for the discussion. Otherwise everything in the function would be red due to indentation

PR #2214, opened 2017-03-22
Simplify conditional assignments in Fluid/Dissipation.mo

miscco:
An assignment of the type
```
   = if condition then true else false;
```
is trivially equivalent to 
```
  = condition
```


beutlich:
Thank you for your pull request which got merged to the master branch.

Issue #2052, opened 2017-01-15
No header(s) for ModelicaExternalC library


modelica-trac-importer:
**Reported by efredriksson on 31 Aug 2016 15:04 UTC**
Currently there is no ModelicaRandom.h header file with function declarations (for the functions in ModelicaRandom.c) to use with the ModelicaExternalC library.

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/2052

modelica-trac-importer:
**Comment by anonymous on 31 Aug 2016 17:37 UTC**
Confirmed. And so do the other files, Strings, FFT, Internal. I guess it is missing for historical reasons where the C files where included. Not sure if header files are really needed.

modelica-trac-importer:
**Comment by efredriksson on 1 Sep 2016 08:09 UTC**
Replying to [comment:1 anonymous]:
> Confirmed. And so do the other files, Strings, FFT, Internal. I guess it is missing for historical reasons where the C files where included. Not sure if header files are really needed.

You would only need function declarations for your interface (functions you are going to call from Modelica code), so it is probably not needed for Internal.

It is of course possible to generate these function declarations from your Modelica code. But it is still nice to have the function declarations in a .h file, C compilers will then check that the 2 function declarations are the same.

modelica-trac-importer:
**Comment by anonymous on 1 Sep 2016 08:15 UTC**
A Modelica tool cannot both generate the prototype and include a .h file. When using `Include="..."`, it might not be a prototype, but an inline static function instead.

modelica-trac-importer:
**Comment by beutlich on 2 Sep 2016 11:59 UTC**
Since there is nothing to do, I am closing this ticket.

modelica-trac-importer:
**Comment by efredriksson on 5 Sep 2016 13:05 UTC**
Replying to [comment:4 beutlich]:
> Since there is nothing to do, I am closing this ticket.

What is the motivation for that there is nothing to do?

modelica-trac-importer:
**Comment by anonymous on 5 Sep 2016 20:34 UTC**
ModelicaExternalC consists of FFT, Internal, Random and Strings - all w/o header.
ModdelicaIO consists of IO - with header.
ModelicaMatIO consists of MatIO - with header.
ModelicaStandardTables consists of StandardTables - with header.

Missing headers are consistenst within ModelicaExternalC itself, but not for all four C libs.

[...]

Issue #2008, opened 2017-01-15
Make ModelicaIO.h usable from a C++ environment


modelica-trac-importer:
**Reported by ottot on 10 Jun 2016 10:03 UTC**
The functions in ModelicaIO.h should be marked as `extern "C"` when compiled with C++ compiler, see attached patch.

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/2008

modelica-trac-importer:
**Comment by hansolsson on 13 Jun 2016 11:13 UTC**
We should avoid having defines with generic names such as "EXTERN" since that is likely to cause conflicts. 

Either have no define (requires a bit more) - or use a common prefix, such as the existing "MODELICA_".

modelica-trac-importer:
**Comment by beutlich on 13 Jun 2016 13:04 UTC**
There already is EXTERN defined in ModelicaMatIO.h.

modelica-trac-importer:
**Comment by hansolsson on 13 Jun 2016 13:23 UTC**
Replying to [comment:2 beutlich]:
> There already is EXTERN defined in ModelicaMatIO.h.

Ah, yes, we should thus rename both of them. If necessary with another ticket.
We are lucky that we have NOT stumbled on this yet, but e.g. f2c uses "Extern" and not "EXTERN" (with completely different definition).

modelica-trac-importer:
**Comment by beutlich on 13 Jun 2016 13:31 UTC**
Right. Will rename to MODELICA_EXTERN (with context of this ticket).

modelica-trac-importer:
**Modified by beutlich on 15 Jun 2016 15:53 UTC**

modelica-trac-importer:
**Comment by beutlich on 23 Jun 2016 20:40 UTC**
Hans: Should we instead utilize the MODELICA_EXPORT macro which alreads is used in ModelicaFFT.c, ModelicaInternal.c, ModelicaRandom.c, ModelicaStrings.c and even in ModelicaIO.c.

It is
```c
#if !defined(MODELICA_EXPORT)
#   define MODELICA_EXPORT
#endif
```

Proposal: Should we
* rename EXTERN to MODELICA_EXPORT in ModelicaMatIO.h,
* add MODELICA_EXPORT in ModelicaStandardTables.h,
* remove MODELICA_EXPORT from ModelicaIO.c and add it to ModelicaIO.h instead?

This way it will be consistent throughout the C Sources.

[...]

PR #2216, opened 2017-03-24
Restructure ModelicaExternalC sources

beutlich:
* Add headers for ModelicaExternalC sources
* Use MODELICA_EXPORT consistently (in header declaration only) and set it to `extern "C"` in case of C++
* Rename EXTERN to MATIO_EXTERN in ModelicaMatIO.h
* See #2008 and #2052 

@HansOlsson I kindly ask for your review (since you contributed in the above-mentioned tickets). Thank you!

One minor issue I see is that the static function linking is removed now. For that reason I also removed the respective comment on `MODELICA_EXPORT`. But since tool vendors need to provide libraries anyway, it should be fine this way.

@sjoelund FYI and testing with OMC. Thank you!


beutlich:
@HansOlsson Any feedback so far?

PR #2220, opened 2017-03-26
Fix argument order in ModelicaIO.c

adrpo:
The order of the arguments is wrong in the read*Table calls!


beutlich:
Thank you for reporting and the PR. Fixed differently by 65547e6f8115e11f87ff5125d2d9a78093f19cb9.

PR #2255, opened 2017-05-30
Generalize Multiplex/DeMultiplex for arbitrary number of connector signals

beutlich:
Similar as the MultiSum block the new Multiplex and DeMultiplex blocks feature an arbitrary number of connector signals, which is automatically updated by tools supporting the connectorSizing annotation.

Once this pull request gets merged, the Multiplex2, Multiplex3, Multiplex4, Multiplex5, Multiplex6, DeMultiplex2, DeMultiplex3, DeMultiplex4, DeMultiplex5, and DeMultiplex6 blocks could be declared obsolete. I will file an extra pull request for this later.

Credits go to @dietmarw for pointing to the connectorSizing annotation and the MultiSum block.

### DIFF
--- a/Modelica/Blocks/Interfaces.mo
+++ b/Modelica/Blocks/Interfaces.mo
@@ -172,6 +172,7 @@ Connector with one input signal of type Integer.
 Connector with one output signal of type Integer.
 </p>
 </html>"));
+
   connector RealVectorInput = input Real
     "Real input connector used for vector of connectors" annotation (
     defaultComponentName="u",
@@ -231,8 +232,8 @@ Integer input connector that is used for a vector of connectors,
 for example <a href=\"modelica://Modelica.Blocks.Interfaces.PartialIntegerMISO\">PartialIntegerMISO</a>,
 and has therefore a different icon as IntegerInput connector.
 </p>
-
 </html>"));
+
   connector BooleanVectorInput = input Boolean
     "Boolean input connector used for vector of connectors" annotation (
     defaultComponentName="u",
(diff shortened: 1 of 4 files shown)


beutlich:
Actually, Multiplex2, Multiplex3, Multiplex4, Multiplex5, Multiplex6, DeMultiplex2, DeMultiplex3, DeMultiplex4, DeMultiplex5, and DeMultiplex6 cannot be removed since the Support vector signals, which is not the cas for the new generalized Mux/Demux blocks.

dietmarw:
I added a fix (thanks to @HansOlsson) so that the new Mux and Demux blocks now also support vector signals. In addition I've added test models as this is required for any new model added to the MSL.

beutlich:
@MartinOtter What do you think about this new enhancement?

tbeu:
@MartinOtter Any comments on this enhancement?

beutlich:
@MartinOtter I updated this pull request (=PR) with an additional capitalization fix.

beutlich:
@MartinOtter Do you want additional change to set Multiplex2, Multiplex3, Multiplex4, Multiplex5, Multiplex6, DeMultiplex2, DeMultiplex3, DeMultiplex4, DeMultiplex5, and DeMultiplex6 obsolete, since the proposed blocks supersede them?

MartinOtter:
I would  not mark  them now as obsolet but keep them for a while to get experience with the blocks with arbitrary signals (how this works in the various tools). Furthermore, sometimes it might be simpler for a user to just use a mux/demux block with two (fixed) connectors.

Issue #2316, opened 2017-09-01
Add Integrator block with external reset trigger


beutlich:
Unfortunately there is no resettable integrator block in Modelica.Blocks.Continuous. Possible options are
* reset to zero
* reset to y_start
* reset to previously hold value
* consider upper and lower limit (as [LimIntegrator](https://github.com/modelica/Modelica/blob/master/Modelica/Blocks/Continuous.mo#L95-L201))

@tobolar I am aware of [LimIntegratorReset](https://github.com/modelica/Modelica_LinearSystems2/blob/master/Modelica_LinearSystems2/WorkInProgress/Controller/LimIntegratorReset.mo), however this new block has no external reset trigger (yet) and it is not decided if it will be released one day.

PR #2319, opened 2017-09-07
Add optional reset to Integrator and LimIntegrator

beutlich:
See #2316.

### DIFF
--- a/Modelica/Blocks/Continuous.mo
+++ b/Modelica/Blocks/Continuous.mo
@@ -5,9 +5,13 @@ package Continuous "Library of continuous control blocks with internal states"
   import Modelica.SIunits;
   extends Modelica.Icons.Package;
 
-  block Integrator "Output the integral of the input signal"
+  block Integrator "Output the integral of the input signal with optional reset"
     import Modelica.Blocks.Types.Init;
     parameter Real k(unit="1")=1 "Integrator gain";
+    parameter Boolean use_reset = false "=true, if reset port enabled"
+      annotation(Evaluate=true, HideResult=true, choices(checkBox=true));
+    parameter Boolean use_set = false "=true, if set port enabled and used as reinitialization value when reset"
+      annotation(Dialog(enable=use_reset), Evaluate=true, HideResult=true, choices(checkBox=true));
 
     /* InitialState is the default, because it was the default in Modelica 2.2
      and therefore this setting is backward compatible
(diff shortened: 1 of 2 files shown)


beutlich:
The one question I have is if it fits in Modelica.Blocks.Continuous?

beutlich:
@modelica/map-lib-write Any feedback?

beutlich:
Yes, I implemented Reset&Hold behaviour. Which option seems preferable?

1. Add Reset behaviour only.
1. Add Reset&Hold behaviour (as is), but improve documentation. The user still can use Modelica.Blocks.Logical.Change (and similar blocks) to model plain Reset behaviour.
1. Add new Boolean parameter hold and have der(y)=if localReset and hold then 0 else k*u

HansOlsson:
I would prefer to only add reset. 
It seems possible to "hold" by using Modelica.Blocks.Logical.Switch for the input (connected to a constant 0) - which makes "hold" sort of orthogonal from "reset".

Combining reset and hold with a set-input seems to be complicated: if the set-input is time-varying it seems as if the "held" value should change to match that; that is not implemented and seems unnecessarily complicated.

beutlich:
Implementing only the Reset behaviour makes it possible to simply __extend__ a IntegratorWithReset from Integrator and put it anywhere else in the library (which was impossible with the Hold behaviour on board).

beutlich:
I removed the hold part of the resettable integrator and added 3 simple test models of a self-resetting integrator, a reset&hold integrator and a limiting resettable integrator. OK for you to merge?

HansOlsson:
> I removed the hold part of the resettable integrator and added 3 simple test models of a self-resetting integrator, a reset&hold integrator and a limiting resettable integrator. OK for you to merge?

That seems ok, except  I couldn't find the version above.

beutlich:
> That seems ok, except I couldn't find the version above.

What do you mean by "version above"? My previous commit using reset+hold? That commit was removed by `git commit --amend` and `git push --force`.

HansOlsson:
> What do you mean by "version above"? My previous commit using reset+hold? That commit was updated by  git commit --amend  and  git push --force .

Yes, and now I see it, and it seems ok. 

I was just confused that there was no new commit listed for this.

beutlich:
OK, thank you for your review @HansOlsson.

beutlich:
I compared Buildings.Utilities.Math.IntegratorWithReset vs. Modelica.Blocks.Continuous.Integrator in the example model Buildings.Utilities.Math.Examples.IntegratorWithReset and can confirm, that both integrators give identical results.

MartinOtter:
A slight drawback is that the icon of the "set" connector is at the place of %name, so the instance name and the connector icon overlap. It would be better to place the "set" connector icon to the "left", above the RealInput u connector.

beutlich:
> A slight drawback is that the icon of the "set" connector is at the place of %name

It also happens for gains with some more digits.
int

Actually I wanted the set and reset input connectors to match the vertical "reset" line by intention.

MartinOtter:
The strings "%name" and "k=%k" could be "right justified" a little bit "left" to "reset". In this case it would be guaranteed that no overlapp occurs

beutlich:
The "%name" is inherited from Modelica.Blocks.Icons.Block. It has the same (horizontal) width from -150 to 150 as the "k=%k" string. Not that easy to change then.

Issue #2307, opened 2017-08-22
Final modifier inheritance in Electrical.Analog sources


beutlich:
It is
```mo
model RampVoltage "Ramp voltage source"
  parameter SI.Voltage V(start=1) "Height of ramp";
  parameter SI.Time duration(min=Modelica.Constants.small, start=2) "Duration of ramp";
  extends Interfaces.VoltageSource(redeclare Modelica.Blocks.Sources.Ramp
    signalSource(final height=V, final duration=duration));
  // ...
end RampVoltage;
```
i.e. signalSource.height is redeclared with a final modifier. However, it is not the case in StepVoltage, e.g.,
```mo
model StepVoltage "Step voltage source"
  parameter SI.Voltage V(start=1) "Height of step";
  extends Interfaces.VoltageSource(redeclare Modelica.Blocks.Sources.Step
  signalSource(height=V));
  // ...
end StepVoltage;
```
Is there a special reason or is this just an inconsistency?

MartinOtter:
The correct way is to use "final" in a modifier for parameters. However, I guess that there are many places in MSL where final is not used.

christiankral:
The changes of a0d8187286dfdb6b2a0379e11b3e78e8e48b92db broke quite some models of the MSL. 

Example ``Modelica.Electrical.Analog.Semiconductors.HeatingDiode``:

The original implementation 
``extends Modelica.Electrical.Analog.Interfaces.ConditionalHeatPort(useHeatPort=true);``
was replaced by 
``extends Modelica.Electrical.Analog.Interfaces.ConditionalHeatPort(final useHeatPort=true);``, so the heat port is always on  -- which is not the intention of the model.

When running the test example ``Modelica.Electrical.Analog.Examples.HeatingRectifier`` then Dymola complains that the final parameter ``heatPort`` gets overwritten again by
`` Modelica.Electrical.Analog.Semiconductors.HeatingDiode HeatingDiode1(
      useHeatPort=true) ...;``
which is definitely not legal. 

I founrd some more broken models:
- Modelica.Electrical.Machines.Examples
- Modelica.Magnetic.FundamentalWave.Examples.BasicMachines

beutlich:
Do you recommend to revert a0d8187286dfdb6b2a0379e11b3e78e8e48b92db because of the trouble with useHeatPort?

Furthermore, there are some ConditionalHeatPort that take a parametric T_Ref and other that take a literal temperature of 293.15K.

[...]

Issue #1771, opened 2017-01-15
Wrong combiTimeTable output when startTime < first table point


modelica-trac-importer:
**Reported by filip&#x2e;jorissen on 1 Sep 2015 14:10 UTC**
When combiTimeTable.startTime is smaller than the first timestamp of a table then the wrong output values are returned. Consider the attached example, which produces the desired results and then set combiTimeTable.startTime to -1e7. Results are not the same for these two cases.

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/1771

modelica-trac-importer:
**Comment by beutlich on 1 Sep 2015 14:52 UTC**
If combiTimeTable.startTime is -1e7 then combiTimeTable.y![1] extrapolates to constant 1 since 1e7 is greater than combiTimeTable.t_max = 2049095. This is the expected behaviour, right?

modelica-trac-importer:
**Comment by filip&#46;jorissen on 1 Sep 2015 15:36 UTC**
It seems I misinterpreted the implementation of the model. 'startTime' has description "Output = offset for time < startTime". This does not seem to be correct? It's rather something like "Output = table(time-startTime)", i.e. it shifts all outputs in time. Based on this I was able to sort out the bug I was experiencing. If my interpretation is correct I suggest to adjust the description of 'startTime' and possibly rename the parameter too.

Filip

modelica-trac-importer:
**Comment by beutlich on 1 Sep 2015 18:24 UTC**
It's rather like "output = offset + if time < startTime then 0 else table(time-startTime)". I guess you did expect "output = offset + if time < startTime then 0 else table(time)". Thus indeed the table is shifted in time. At least it is documented.

Most of the source blocks have this parameter startTime, thus it is not simple thing to change.

beutlich:
[...]

PR #2329, opened 2017-09-27
Introduce T_operational parameter to resistors / conductors

beutlich:
T_ref is used both as reference temperature (for the linear temperature dependency of the resistance/conductance) and as operational temperature (if no conditional heat port is available). With the introduction of T_operational the confusion can be avoided by a clean separation of the two temperatures.

Furthermore, parameter T can now be modified as final. Setting T_operational=T_ref by default enables the backward compatibility.

This pull request is the prerequisite for a clean solution of #2307 (final modifier in extends clause) and #2318 (multiphase).

### DIFF
--- a/Modelica/Electrical/Analog/Basic.mo
+++ b/Modelica/Electrical/Analog/Basic.mo
@@ -37,11 +37,12 @@ package Basic "Basic electrical components"
     parameter SI.Resistance R(start=1)
       "Resistance at temperature T_ref";
     parameter SI.Temperature T_ref=300.15 "Reference temperature";
+    parameter SI.Temperature T_operational=T_ref "Operational temperature if useHeatPort = false" annotation(Dialog(enable=not useHeatPort));
     parameter SI.LinearTemperatureCoefficient alpha=0
       "Temperature coefficient of resistance (R_actual = R*(1 + alpha*(T_heatPort - T_ref))";
 
     extends Modelica.Electrical.Analog.Interfaces.OnePort;
-    extends Modelica.Electrical.Analog.Interfaces.ConditionalHeatPort(T=T_ref);
+    extends Modelica.Electrical.Analog.Interfaces.ConditionalHeatPort(final T=T_operational);
     SI.Resistance R_actual
       "Actual resistance = R*(1 + alpha*(T_heatPort - T_ref))";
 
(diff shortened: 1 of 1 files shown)


beutlich:
See #1771 for a similar issue where a parameter (startTime) was used for two distinct functionalities, the start time and a time shift. This was resolved by backward compatible introduction of a new parameter shiftTime.

christiankral:
Example to be broken by new implementation
```modelica
model TestResistor
  Modelica.Electrical.Analog.Basic.Resistor resistor(final T=343.15) annotation (Placement(transformation(extent={{-10,-10},{10,10}})));
  Modelica.Electrical.Analog.Basic.Ground ground annotation (Placement(transformation(extent={{-20,-40},{0,-20}})));
equation 
  connect(resistor.p, ground.p) annotation (Line(points={{-10,0},{-10,-20}}, color={0,0,255}));
end TestResistor;
```

beutlich:
> This implementation ist not backwards compatible.

Right. I'd rather consider this kind of existing user models with `resistor(final T=343.15)` as illegal now, which can happen with tools getting more Modelica compliant and ibraries getting cleaner in implementation and interpretation. Thus, we can 
* tell the user about his previously correct but now wrong final parameter setting (we did such things in the past, see e.g. [doc.modelica.org](http://doc.modelica.org/help/Modelica_UsersGuide_ReleaseNotes.html#Modelica.UsersGuide.ReleaseNotes.Version_3_2_2) and search for "non-backwards")
* skip the final modification in the pull request for the next minor release only and add it later for the next major release (where conversion scripts are allowed).

> A very easy solution is to change the actual doc string of the original implementation from "Fixed device temperature if useHeatPort = false" to "Operational temperature if useHeatPort = false"

I'd say this introduces even more confusion as we stil have "Resistance at temperature T_ref" for parameter R.

christiankral:
>I'd say this introduces even more confusion as we stil have "Resistance at temperature T_ref" for parameter R.

I do not see the confusion. 

```modelica
parameter SI.Resistance R(start=1) "Resistance at temperature T_ref";
```
This describes exactly what it is: The resistance R refers to the reference temperature T_ref. The resistor model may then be operated at operating temperature T (if no heat port is connected).

AHaumer:
I've attached an explanation about the temperatures. 
I don't get the idea why we should use "fixed" when extending the CondtionalHeatPort?
I also do not understand why a user is not allowed to set the op. temperature T of a resistor "fixed"?
[TemperatureDependency.pdf](https://github.com/modelica/Modelica/files/1346538/TemperatureDependency.pdf)

kristinmajetta:
I do not see the problem at all. I think we have to see the users perpective. In my opinion for a user it is absolutely clear, that T_ref is the reference temperature used for calculating the value of the resistance when the resistor is not working at reference temperature and that on the other hand T_ref is used as operational temperature when no heat port is used.

PR #2338, opened 2017-10-02
Removes note about an old MSL dependency

dietmarw:
Since Fluid.Dissipation is part of MSL this is no longer necesssary.


beutlich:
See #2340 for the general issue on the maintenance and update strategy of Fluid.Dissipation.

PR #2339, opened 2017-10-02
Update README.md

agerwitz:


dietmarw:
This is basically a question of style and not an error as such. I prefer the version as is.

PR #2341, opened 2017-10-04
Rewrite Modelica.Electrical.MultiPhase.Functions.numberOfSymmetricBaseSystems

beutlich:
Rewrite numberOfSymmetricBaseSystems as a non-recursive one-liner (that is auto-inlined)

### DIFF
--- a/Modelica/Electrical/MultiPhase.mo
+++ b/Modelica/Electrical/MultiPhase.mo
@@ -2704,20 +2704,7 @@ The back transformation matrix can be used to determine the time phasors from th
       input Integer m=3 "Number of phases";
       output Integer n "Number of symmetric base systems";
     algorithm
-      // Init number of base systmes
-      n := 1;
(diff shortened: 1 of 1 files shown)


christiankral:
Please be patient, I am considering your proposal, see https://trac.openmodelica.org/OpenModelica/ticket/4550

sjoelund:
I fixed one issue in OpenModelica; the remaining issue is with the model: `log(m/2)/log(2)` for `m=4` is not exactly 1.0 depending on computation order (we do `1/log(2)`=1.442695040888963 and we end up with 0.999999999997 instead of 1.0; `integer()` rounds that down).

sjoelund:
I talked a bit with @casella and I guess log((n+0.2)/2)/log(2) would fix any rounding errors. Not at a computer to check it though.

beutlich:
Right. No point to merge it then.

PR #2344, opened 2017-10-06
Make use of groupImage to help understanding block parameters

dietmarw:
This commit makes use of the new `groupImage` annotation.
It simply places the already present image from the documentation side by side
with the parameters. 
This simplifies the usage and understanding of parameter's meaning to a large degree.


beutlich:
Nice idea. I would prefer to have PNGs with an alpha channel in order to better support transparency on non-white backgrounds. See Modelica.Magnetic.FluxTubes.Examples.MovingCoilActuator.Components.PermeanceActuator.N and [MovingCoilActuator_dimensions.png](https://github.com/modelica/Modelica/blob/master/Modelica/Resources/Images/Magnetic/FluxTubes/Examples/MovingCoilActuator/MovingCoilActuator_dimensions.png) for an example.

sjoelund:
What about black backgrounds?

dietmarw:
I've made the PNGs now with transparent backgrounds. I can not think that black backgrounds are relevant.

beutlich:
Looks poor on black background.

permeanceactuator1

dietmarw:
Right. So white background would be the safe (but uglier) choice. I'll see what the LO @MartinOtter would prefer. I can easily revert the transparency again.

beutlich:
Anyway, transparency should be consistent then troughout the MSL.

dietmarw:
I agree that it has to be consistent. But I don't think we need to change all images right now but start with all images that are used as `groupImage`.

HansOlsson:
With non-transparent image background it will look somewhat bad on all backgrounds in parameter-dialog - except the one corresponding to the image background.
With transparent image background it will look good - except when background in the parameter-dialog corresponds to the foreground color(s) of the image.

A reasonable statement would be that we say that the transparent images are assumed to work with light backgrounds. Based on that tools that want to use a dark background in the parameter-dialog could:

- Make the image background white (which will be the same as non-transparent image background)
- Create a complementary image by swapping b&w

Color-images will be more complicated

MartinOtter:
White background of the image will always work, independent of the background of the parameter menu, even if it is black. I would therefore propose to use the first version from Dietmar with white background. 

Furthermore, a ticket should be filled for the Modelica Language to discuss how images should be provided. If transparent images should be supported, then there should be a clean process how to it so that it works for all potential backgrounds.

sjoelund:
@MartinOtter Note that the same issues apply for example for diagram/icon layers. I believe all/most vendors choose a bright background there since the default line and text colour is black in the spec, but the dialog is usually a box from the desktop environment / OS.

dietmarw:
FYI I've reverted the transparent PNGs again.

HansOlsson:
> @MartinOtter Note that the same issues apply for example for diagram/icon layers. I believe all/most vendors choose a bright background there since the default line and text colour is black in the spec, but the dialog is usually a box from the desktop environment / OS.

And in the documentation layer as well. Most of the images that Dietmar modified were used in the documentation as well as the parameter-dialog.

beutlich:
> FYI I've reverted the transparent PNGs again.

OK. Will file a new MSL issue soon to have it non-transparent everywhere then.

MartinOtter:
Nice improvement

PR #2352, opened 2017-10-11
add missing each in Modelica.Media.Examples.TestOnly.MixIdealGasAir

FrenkelJ:
Because medium2.X is a vector the modification of fixed attribute needs each prefix.

### DIFF
--- a/Modelica/Media/package.mo
+++ b/Modelica/Media/package.mo
@@ -2921,7 +2921,7 @@ points, e.g., when an isentropic reference state is computed.
         p(start=1.0e5, fixed=true));
       Medium.BaseProperties medium2(
         T(start=300.0, fixed=true),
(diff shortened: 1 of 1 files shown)


thorade:
You could run the following commands from a Linux terminal to find a couple of more candidates:

```
mkdir temp
cd temp
wget -r -np -nd --accept=err https://libraries.openmodelica.org/branches/master/Modelica_trunk/files/
grep "missing" *.err
```

FrenkelJ:
@thorade Thats a good idea, would it be possible for you to great a pull request from each valid candidate.

sjoelund:
We have https://libraries.openmodelica.org/branches/master/ModelicaTest_trunk/files/ as well if you are interested.

PR #2353, opened 2017-10-11
add missing each in Modelica.Blocks.Math.mo

thorade:
to Modelica.Blocks.Math.mo
as discussed in #2352 

search for `missing` here:
https://libraries.openmodelica.org/branches/master/Modelica_trunk/files/Modelica_trunk_Modelica.Blocks.Examples.Rectifier6pulseFFT.err


thorade:
Sidenote: Probably `abs` is not a good variable name.

Issue #2247, opened 2017-04-25
Questions on Modelica.Mechanical.Rotational.Components.ElastoBacklash2


UlfNordstrom:
Hi,

We have a customer who has asked for some details on the ElastoBacklash2 implementation that I cannot answer. Can you help explain (the odd numbering below is due to additional questions already answered)?

-------
Q1) tau_d
In both ElastoBacklash and ElastoBacklash2, you can see "tau_d = d*w_rel;"
But in the info layer, it says "if phi_rel - phi_rel0 < b/2 then tau_d = 0;"
Can you explain?

Q4) ElastoBacklash equation

--- ElastoBacklash ---
    tau = if abs(b) <= bEps then
             tau_c + tau_d
          else if phi_diff > bMax then
             smooth(0, noEvent(if tau_c + tau_d <= 0 then 0 else tau_c + min(tau_c, tau_d)))  <-- 
          else if phi_diff < bMin then
             smooth(0, noEvent(if tau_c + tau_d >= 0 then 0 else tau_c + max(tau_c, tau_d)))  <--
          else 0;

--- ElastoBacklash2 ---
    tau = if abs(b) <= bEps then 
             tau_c + tau_d
          else if phi_diff > bMax then 
             (if tau_c + tau_d <= 0 then 0 else tau_c + tau_d)  <--
          else if phi_diff < bMin then 
             (if tau_c + tau_d >= 0 then 0 else tau_c + tau_d)
          else 0;

Is "tau_c + min(tau_c, tau_d)" and "tau_c + max(tau_c, tau_d)" correct?
Please explain the exact difference between the two equations.
Please explain the reason why they differ.

Q6) Difference between ElastoBacklash and ElastoBacklash2
I understand the difference between the two is "tau_c + min(tau_c, tau_d)" and "tau_c + max(tau_c, tau_d)" (shown as Q4)
Is my understanding correct?

-------

Best regards,
Ulf

beutlich:
1dd9dde5a5dd19b159d7182327f6d22bc1ba309f introduced ElastoBacklash2 for MSL v3.2.2.

beutlich:
Q1 is addressed by #2357. Q4 and Q6 are support questions and may be also raised at [Stack Overflow](https://stackoverflow.com/questions/tagged/modelica).

MartinOtter:
Q4: The goal is to guarantee that tau is continuous when entering/leaving the backlash region (for simplicity ignoring smooth(..) below):

[...]

PR #2357, opened 2017-10-14
Match computation and documentation of tau_d

beutlich:
refs #2247, Q1: Match computation and documentation of tau_d


beutlich:
**Commented by @MartinOtter**

Bode codes (the original one and the proposed change) are correct. The original code is correct because:
- tau_d is a protected (auxiliary) variable.
- tau =  if abs(b) <= bEps then ... else if phi_diff > bMax then ... else if phi_diff < bMin ... else 0;
This means that tau = 0 in the backlash region, even tau_d is not zero in the backlash region.

I do not see a reason to change a correct model.

PR #2358, opened 2017-10-14
Fix typo steep->step

beutlich:
Shall it read step instead of steep?


MartinOtter:
"steep" (steil) is the intention, not "step" (Sprung). So the original text is correct.

beutlich:
Thanks for clarifaction.

PR #2359, opened 2017-10-14
Avoid default Real parameter evaluation

beutlich:
* Remove Evaluate=true annotation for Real parameters
* Keep Evaluate=true annotation for Boolean and enumeration parameters
* See https://stackoverflow.com/q/21957303/8520615

@thorade FYI

### DIFF
--- a/Modelica/Fluid/Sources.mo
+++ b/Modelica/Fluid/Sources.mo
@@ -144,21 +144,17 @@ with exception of boundary pressure, do not have an effect.
       annotation(Evaluate=true, HideResult=true, choices(checkBox=true));
     parameter Medium.AbsolutePressure p = Medium.p_default
       "Fixed value of pressure"
-      annotation (Evaluate = true,
-                  Dialog(enable = not use_p_in));
+      annotation (Dialog(enable = not use_p_in));
     parameter Medium.Temperature T = Medium.T_default
       "Fixed value of temperature"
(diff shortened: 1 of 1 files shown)


MartinOtter:
There are other "Evaluate=true" parametesr in the MSL, especially in the MultiBody library. There are good reasons for them, so do not remove them.

beutlich:
> There are other "Evaluate=true" parametesr in the MSL, especially in the MultiBody library. There are good reasons for them, so do not remove them.

Sure, joint axis should always be evaluated for perfromance reasons.

Note, this pull requests touches only Modelica.Fluid based on the original question on Stack Overflow.

casella:
@beutlich, I fully agree with your PR.

@HansOlsson, in principle you are right, but in practice I don't think this would make a big difference in most (if not all) cases. BTW, the original post on stackoverflow explicitly mentioned the need of changing m_flow without recompiling, which is absolutely reasonable.

In case one really wants extreme optimization and doesn't care about changing parameter values without recompiling, it is always possible to activate some tool options like 'Evaluate all parameters', which is well known to often speed up simulation and/or improve convergence. In fact, it would be nice to be able to evaluate all parameters except a subset selected via the GUI 😃 

@MartinOtter in this case there are no structural consequences of removing the Evaluate annotation, except for the minor issue mentioned by @HansOlsson.

PR #2364, opened 2017-10-14
Add missing C_default to PartialMedium

### LINKED ISSUE #2121: C_default not defined in Modelica.Media.Interfaces.PartialMedium
**Reported by rubenbaetens on 24 Nov 2016 09:41 UTC**
Modelica.Media.Interfaces.PartialMedium defines T_default, h_default, X_default and p_default, but no C_default. It would be better to add C_default = fill(0, Medium.nC) to the MSL so that all is consistent. As such, users can use C_default as the default start value and the default concentration for Boundary_pT, and can reassign this value in the medium.

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/2121

beutlich:
Close #2121.

@rubenbaetens FYI

### DIFF
--- a/Modelica/Fluid/Interfaces.mo
+++ b/Modelica/Fluid/Interfaces.mo
@@ -545,7 +545,7 @@ the boundary temperatures <code>heatPorts[n].T</code>, and the heat flow rates <
       "Start value of mass fractions m_i/m"
         annotation (Dialog(tab="Initialization", enable=Medium.nXi > 0));
       parameter Medium.ExtraProperty C_start[Medium.nC](
-           quantity=Medium.extraPropertiesNames)=fill(0, Medium.nC)
+           quantity=Medium.extraPropertiesNames) = Medium.C_default
       "Start value of trace substances"
         annotation (Dialog(tab="Initialization", enable=Medium.nC > 0));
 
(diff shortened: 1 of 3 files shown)


MartinOtter:
Bode codes (the original one and the proposed change) are correct. The original code is correct because:
- tau_d is a protected (auxiliary) variable.
- tau =  if abs(b) <= bEps then ... else if phi_diff > bMax then ... else if phi_diff < bMin ... else 0;
This means that tau = 0 in the backlash region, even tau_d is not zero in the backlash region.

I do not see a reason to change a correct model.

MartinOtter:
To make this more clear. The code to be implemented should be:

```
     if abs(b) <= bEps then
        tau_c = c*phi_diff;
        tau_d = d*w_rel;
        tau   = tau_c + tau_d;
     elseif phi_diff > bMax then
        tau_c = c*(phi_diff - bMax);
        tau_d = d*w_rel;
        tau   = smooth(0, noEvent(if tau_c + tau_d <= 0 then 0 else tau_c + min(tau_c,tau_d)));
     elseif phi_diff < bMin then
        tau_c = c*(phi_diff - bMin);
        tau_d = d*w_rel;
        tau   = smooth(0, noEvent(if tau_c + tau_d >= 0 then 0 else tau_c + max(tau_c,tau_d)));
     else
        tau_c = 0;
        tau_d = 0;
        tau   = 0;
     end if;
```

This means that tau_c and tau_d are the unmodified spring and damper torques and the modification of the force law takes place only in "tau". The implementation with tau_d = d*w_rel reflects this view.

Issue #2320, opened 2017-09-07
Implement License change decision to change to BSD-3-Clause


hubertus65:
- [x] Add LICENSE file containing standard BSD-3-Clause text to this repository (for easy visibility of the license at GitHub)
- [x] Change license comments to BSD-3-Clause in C-Sources
- [x] Change license comments to BSD-3-Clause in README.md
- [x] Remove classes that declare ModelicaLicense2 (and accept breaking backward compatibility)
- [x] Remove links to ModelicaLicense2
- [x] Change license comments to BSD-3-Clause in Modelica files

beutlich:
Can we please also decide on all copyright disclaimers in Modelica files and C-sources?! As we discussed it in Praha it turned out the current copyright disclaimers are invalid w.r.t. the MA.

hubertus65:
You mean copyright disclaimers claiming copyright by the MA, since we do not have a contributors license agreement?

beutlich:
Exactely. The ones like [here](https://github.com/modelica/Modelica/blob/master/Modelica/Resources/C-Sources/ModelicaStandardTables.c#L3).

hubertus65:
Hmmm. I am not certain about that. If anyone legally worked as a private person with this, and the person itself would say this was done on behalf of the MA, it would be correct. This certainly holds for academics in a number of countries. I would not touch this and rather spend effort on the CLA, which would make it correct.

beutlich:
@modelica/map-lib-write I created the new branch issue2320-license (based on master) where all of the above subtasks should go before they will be merged.

hubertus65:
What do you mean by "Remove classes that declare ModelicaLicense2 (and accept breaking backward compatibility)"? Do you think there are classes that were intentionally put under Modelica LIcense2 and can't be changed to BSD?

beutlich:
> What do you mean by "Remove classes that declare ModelicaLicense2 (and accept breaking backward compatibility)"? 

My intention was to completely remove the classes Modelica.UsersGuide.ModelicaLicense2 and ModelicaServices.UsersGuide.ModelicaLicense2 (instead of setting it obsolete). Since these classes are documentation-only classes, the case of breaking the backward compatibility should be acceptable. Let's see what the project manager @AHaumer thinks about it.

dietmarw:
I agree that removing the licence classes is an acceptable non-backward compatible change. I can not think any models *depending* on it.

beutlich:
[...]

PR #2369, opened 2017-10-16
Workaround for dummy usertab function

sjoelund:
This is a workaround for GCC/LLVM optimizing the function call to
usertab which assumes that usertab will not be relocated at runtime.
Providing both a user-defined usertab and the dummy usertab object in
the same executable is possible on Linux, but confuses the C-compilers.

The GCC/autotools projects have been updated, but since VisualStudio
does not include the dummy usertab, the new source-file is not included
for VisualStudio.

Note: There is a cleaner way to implement handling the usertab function.
By using dlopen/etc when calling the table constructor, we could query
for a usertab function at runtime and not need it during compile-time
(making it easier to work with pre-compiled ModelicaStandardTables).
That would also enable the possibility to have two different usertab
functions (passing the name of the function to the constructor) and only
keep the current (a bit inconvenient) method of Modelica tools needing
to only sometimes provide a dummy usertab function.

@beutlich without these changes, ModelicaStandardTables only works properly with the `-O0` flag if dummy usertab is provided (getting either illegal pointers or empty strings right after calling usertab). I would much rather have a `dlopen`-based approach sometime in the future.

### DIFF
--- a/Modelica/Resources/BuildProjects/VisualStudio2005/ModelicaStandardTables.vcproj
+++ b/Modelica/Resources/BuildProjects/VisualStudio2005/ModelicaStandardTables.vcproj
@@ -275,6 +275,10 @@
 				RelativePath="..\..\C-Sources\ModelicaStandardTables.c"
 				>
 			</File>
+			<File
+				RelativePath="..\..\C-Sources\ModelicaStandardTablesUsertab.c"
+				>
+			</File>
 		</Filter>
 		<Filter
 			Name="Header Files"
--- a/Modelica/Resources/BuildProjects/autotools/Makefile.am
+++ b/Modelica/Resources/BuildProjects/autotools/Makefile.am
@@ -4,7 +4,7 @@ libModelicaIO_la_SOURCES             = ../../C-Sources/ModelicaIO.c
 libModelicaIO_la_LIBADD              = libModelicaMatIO.la
 libModelicaMatIO_la_SOURCES          = ../../C-Sources/ModelicaMatIO.c
 libModelicaMatIO_la_LIBADD           = libzlib.la @LIBZLIB@ @LIBHDF5@
-libModelicaStandardTables_la_SOURCES = ../../C-Sources/ModelicaStandardTables.c
+libModelicaStandardTables_la_SOURCES = ../../C-Sources/ModelicaStandardTables.c  ../../C-Sources/ModelicaStandardTablesUsertab.c
 libModelicaStandardTables_la_LIBADD  = libModelicaMatIO.la
 
 # If the OS does not have zlib available, compile it and include it together with libModelicaStandardTables
--- a/Modelica/Resources/BuildProjects/gcc/Makefile
+++ b/Modelica/Resources/BuildProjects/gcc/Makefile
@@ -8,7 +8,8 @@ INC = -I"../../C-Sources/zlib"
 TARGETDIR = linux64
 
 TABLES_OBJS = \
-	ModelicaStandardTables.o
+	ModelicaStandardTables.o \
+	ModelicaStandardTablesUsertab.o
 
 MATIO_OBJS = \
 	ModelicaMatIO.o
--- a/Modelica/Resources/C-Sources/ModelicaStandardTables.c
+++ b/Modelica/Resources/C-Sources/ModelicaStandardTables.c
@@ -5697,11 +5697,3 @@ static READ_RESULT readTable(_In_z_ const char* fileName, _In_z_ const char* tab
     return NULL;
 #endif /* #if !defined(NO_FILE_SYSTEM) */
 }
-
-#if defined(DUMMY_FUNCTION_USERTAB)
-int usertab(char* tableName, int nipo, int dim[], int* colWise,
-            double** table) {
-    ModelicaError("Function \"usertab\" is not implemented\n");
-    return 1; /* Error */
-}
-#endif /* #if defined(DUMMY_FUNCTION_USERTAB) */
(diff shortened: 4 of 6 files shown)


beutlich:
Yes, this is relevant. We have something similar in SimulationX introduced, called dummyusertab.c instead of your ModelicaStandardTablesUsertab.c. ModelicaStandardTables.c is compiled with undefined DUMMY_FUNCTION_USERTAB.

```c
/* Implementation of user defined function for table computation
   in the Modelica Standard Library:

   The following #define is available.

    IGNORE_DUMMY_FUNCTION_USERTAB: Ignore this dummy function "usertab"
                                 : if it already is defined by user code
*/

#include "ModelicaUtilities.h"

#if !defined(IGNORE_DUMMY_FUNCTION_USERTAB)
int usertab(char* tableName, int nipo, int dim[], int* colWise, double** table) {
	ModelicaError("Function \"usertab\" is not implemented\n");
	return 1; /* Error */
}
#endif
```
This way, usertab is compiled by default, i.e. if IGNORE_DUMMY_FUNCTION_USERTAB is not defined. If there is a custom usertab.c from a user model/library, it will lead to double symbol ustertab, which is avoided by forced linking (and taking care which symbol survives) or by defining IGNORE_DUMMY_FUNCTION_USERTAB at compile time.

What do you think? Does it make sense?

P.S. I can already smell the merge conflicts from #2320 due to the BSD-2-clause -> BSD-3-clause change.

sjoelund:
Your solution makes sense, but I assume you then always link in a usertab function in your code (or check for this library manually)? We only link it if Tables is used and don't really treat any library special. Perhaps we should just add a version using dlopen/etc since that should be a lot simpler for all tool providers to handle. Anyway, the current MSL with usertab in the same file as the tables implementation doesn't work well at all (same implementation in a separate file works a little bit better).

beutlich:
We only link to usertab (coming from either cutom usertab.c or dummyusertab.c) if ModelicaStandardTables is to be compiled.

sjoelund:
Do you think it would be worth it for me to add a commit using dlopen? I don't want to require special logic in the compiler just for usertab (and especially not workarounds for Visual Studio, MinGW, Linux, BSD, and OSX linkers to deal with duplicate symbols just for this case).

beutlich:
Hm, how do you handle it w/o duplicated symbols then? I there is some file, say usertab.c defining symbol usertab you must not compile the dummy function usertab? How can you decide?

sjoelund:
If dlsym returns NULL, give an error that usertab wasn't provided at runtime. On Windows, I think it's GetModuleHandle(NULL) to access symbols in the process itself. On Linux, you dlopen(NULL). So Tables would contain no usertab and you never need to provide it if you run a platform with dynamic linking

beutlich:
That is correct, but confusing me more and more. If you compile ModelicaStandardTables you need a usertab symbol resolved. Where does it come from? Either dummy one or user one (or both).

sjoelund:
If we add some changes, usertab would not be needed to compile or link ModelicaUserTables. I would resolve it at runtime in the constructor when a table requesting usertab is called.

beutlich:
I thought the desired feature of the usertab is to resolve the table at compile time. But I believe @MartinOtter and his colleagues were not that happy with this feature and the complicate usage.

sjoelund:
If it's necessary to resolve the symbol at compile-time, usertab probably should have been an external object so you could pass its pointer to the table constructor (some NULL pointer if there is no usertab), but this would not be backwards compatible. But yes, I guess @MartinOtter could clarify if we could use dynamic loading of the usertab (it would still be linked statically with the executable; just loaded at runtime in order to be easier to use). The error messages to the user I believe would be the same because right now we always provide a dummy usertab (so you don't get linker errors if they forgot to add a usertab function).

sjoelund:
It seems having the `usertab` function linked with the main executable and dlopen causes weird behaviour. First of all, --export-dynamic has to be used to provide the symbol. Second of all, linking ModelicaStandardTables with the executable causes the executable's global variables to re-initialize and main to execute again. I guess the most convenient way to pass usertab to the tables would then be through ExternalObject.

sjoelund:
I updated the PR slightly. Made a weak linking reference for the dummy usertab and added the missing ModelicaUtilities include.

beutlich:
The weak linking is helpful. Did not know it.

Do you think you can incorprate the IGNORE_DUMMY_FUNCTION_USERTAB and comment such that I can test it in SimulationX?

sjoelund:
I guess to test it you just change `defined(DUMMY_FUNCTION_USERTAB)` as it is in MSL to `!defined(IGNORE_DUMMY_FUNCTION_USERTAB)`? It's only weak linkage for GCC/clang as well...

sjoelund:
Any update on this? I would like to merge it soon.

beutlich:
@sjoelund Is this still relevant for you?

sjoelund:
Yes, it's still relevant.

sjoelund:
One possible alternative fix might be to use clang/gcc pragmas to force no optimizations on the usertab call.

beutlich:
It is the MinGW test that fails now. Would also be possible to exclude MinGW from your pragmas.

sjoelund:
So I guess 1ba0781 works o MinGW? I don't have a Windows machine to test it on

beutlich:
> So I guess 1ba0781 works o MinGW?

Confirmed. 7b6af0ca442af0543654ab6a10ebbaf8cf8f8fb9 fixes the same issue for Cygwin.

beutlich:
Argh, this does not work on darwin, neither gcc nor clang (Apple LLVM version 8.1.0 (clang-802.0.42)): https://travis-ci.org/tbeu/ModelicaStandardTables/jobs/379729126#L2521 and https://travis-ci.org/tbeu/ModelicaStandardTables/jobs/379729127#L2580. Other than disabling it, I read https://stackoverflow.com/q/36319000/8520615 but there is no workaround given.

beutlich:
7fe9cd3cac2124ea949480436af489a0f3b95d82 disables clang/gcc weak linking on darwin.

PR #2372, opened 2017-10-16
Issue2320 license

jbgohl:
Updates per ticket #2320 to change the license from Modelica License 2 to BSD 3-Clause.


MartinOtter:
I do not think that this change can be just made as you did it because there are legal considerations. The correct way is that all providers of signifcant content to MSL have to sign first a CCLA (similar to the FMI CCLA) and in this CCLA the BSD3 clause is stated. Only afterwards, the license texts should be replaced in the MSL.

beutlich:
@MartinOtter No worries. This was not merged to master but to issues2320-license branch. I am about to prepare a pull request to merge issues2320-license in to master where all the legal issues can be discussed.

beutlich:
According to https://trac.modelica.org/Modelica/attachment/wiki/ElectronicVoting/MSLLicense/PollResults.png the MSL contributors @ECClauss, @HansOlsson and @rfranke are missing, (where @ECClauss agreed in a private mail - just after voting - to also go for the 3-Clause BSD).

PR #2390, opened 2017-11-10
Add bivariate table block with vectorized inputs and vectorized output

### LINKED ISSUE #2321: Add bivariate table block with n inputs and n outputs
The Modelica.Blocks.Tables.CombiTable2D extends from Modelica.Blocks.Interfaces.SI2SO. However, users asked several times for a CombiTable2D with n inputs and n outputs, i.e. a new block **CombiTable2Dn** extending from Modelica.Blocks.Interfaces.MI2MO.

See aso #1153.

beutlich:
The common declarations, equations and graphics between CombiTable2D and CombiTable2DVectorized moved to a new base class CombiTable2DBase.

Close #2321.

TODO: 

- [x] Test models for ModelicaTest @beutlich


beutlich:
> Seems good.

Thank you for your approval.

> A major concern is that there is only one table for the vectorized case and they will thus share "last" pointer.

You are right. This is another case where internal states of external objects (here: a cached index of last interval) is (kind of) problematic. We could introduce GetLastIndex/SetLastIndex functions but I doubt the performance will suffer al lot, given the benchmark analysis I sent to you by private mail on May 24, 2017.

beutlich:
TODO for me: Test models for ModelicaTest

beutlich:
Still thinking about the name:

- CombiTable2DVectorized would match AssignClock and AssignClockVectorized from Modelica_Synchronous
- CombiTable2Dv would match CombiTable1D and CombiTable1Ds (v for vector and s for scalar)
- CombiTable2Dn was my first idea

@MartinOtter Please advise and review this pull request.

PR #2412, opened 2018-01-02
Resize %name of adaptor models (Icon view)

christiankral:
The %name string is too small considering the icon guidelines at `Modelica.UsersGuide.Conventions.Icons`


christiankral:
Referred to wrong change

PR #2419, opened 2018-01-07
Unify location of %name in icon view of Blocks.Interfaces

christiankral:
The ```%name``` location and color shall be unified according to ```Modelica.UsersGuide.Conventions.Icons```


tobolar:
@beutlich A new question arises, why are those blocks (e.g. Modelica.Blocks.Interfaces.PartialBooleanMISO or Modelica.Blocks.Sources.RadioButtonSource) scaled in that way. I have no idea.

PR #2421, opened 2018-01-09
 Mechanics.MultiBody: Improved Diagram layout of examples to better fit the grid

tobolar:


beutlich:
It would be better if you push your changes to separate branch (instead of master). In TortoiseGit you can simply enable the "new branch" checkbox in the commit dialog to do so.

PR #2437, opened 2018-01-22
Bug fix for ticket #2377

hubertus65:
Fixes gas constant for table based media to be correct unit. No name changes, they would not be fine for the next minor release.


hubertus65:
Simple change considered 1 review enough.

Issue #2446, opened 2018-01-25
Summarize terms to be used in Modelica.UsersGuide


christiankral:
We had a discussion on the terms multiphase/multi-phase/multi phase/polyphase in #2290 and #1865. We recently discussed the term electromagnetic/electro magnetic in https://github.com/christiankral/Modelica/commit/a85aa55833e3abde2188ab8725a85431c9b312f4. 

To me it makes therefore sense to somewhere write down the proposed and approved terms in the documentation of the MSL. One place to put this information, sorted by domains (e.g. electrical, magnetic, mechanical, etc.). 

I propose to collect and summarize all these terms in either ```Modelica.UsersGuide.Conventions.ModelicaCode.Naming``` or put it in a new class ```Modelica.UsersGuide.Conventions.ModelicaCode.NamingTerms```. In this place we summarize terms, e.g.

Terms to be used | Terms NOT to be used 
--------------- | ------------------------
polyphase | multi phase, multi-phase, multiphase 

The goal were then to step by step implement a consistent use of the terms to be used.

christiankral:
For now everything is done.

PR #2447, opened 2018-01-28
Included documentation package on terms and spelling

christiankral:
See #2446

### DIFF
--- a/Modelica/package.mo
+++ b/Modelica/package.mo
@@ -1073,6 +1073,65 @@ For parameters, connectors, as well as inputs and outputs of function automatic 
 </html>"));
     end Documentation;
 
+    package Terms "Terms and spelling"
+      extends Modelica.Icons.Information;
+
+      class Electrical "Electrical terms"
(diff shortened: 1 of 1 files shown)


ceraolo:
I love Oxford dictionaries. 
But for electricity-related topics the official source is the International Electrotechnical Commission (IEC ) which publishes the International Electrotechnical Vocabulary, with, in turn is available to everyone from the Internet.
So In case of mismatch,  IEC is the official technical reference.
According to this, 
magnetomotive is ok (IEV 121-11-60)

christiankral:
OK, let's then switch to Electropedia.

christiankral:
However, when using Electropedia links I prefer not show search results, since the corresponding links may not work one day they change the way of querying. Instead I will use technical application terms which also demonstrate the use of the term.

ceraolo:
I quoted IEV to find "official" terminology which should be used in MSL.
I agree that it is not wise to use links to IEV in the MSL documentation.

Issue #2395, opened 2017-11-16
Fixed initial condition on a derivative in PartialGenericHysteresis


eshmoylova:
There is a fixed initial condition on a variable that is a derivative of an output:
https://github.com/modelica/Modelica/blob/f1628f6a8b9362f65d65abfd0c16a149ecf66dce/Modelica/Magnetic/FluxTubes.mo#L7650

Why is this restriction there? It makes the models valid only for a limited number of situations when the derivative of this variable is 0. Consider the following model:
```mo
model tellinen
   Modelica.Blocks.Sources.Sine S1(amplitude=10^5, freqHz=1, offset=0, phase=0, startTime=0) annotation(Placement(transformation(origin={-20.0,100.0},extent={{-20.0,-20.0},{20.0,20.0}},rotation=0)));
   Modelica.Magnetic.FluxTubes.Basic.Ground G1 annotation(Placement(transformation(origin={50.0,30.0},extent={{-20.0,-20.0},{20.0,20.0}},rotation=0)));
   Modelica.Magnetic.FluxTubes.Sources.SignalMagneticPotentialDifference signalSource1 annotation(Placement(transformation(origin={50.0,100.0},extent={{-20.0,20.0},{20.0,-20.0}},rotation=-90)));
   Modelica.Magnetic.FluxTubes.Shapes.HysteresisAndMagnets.GenericHystTellinenHard Core1(A=0.10e-3, Br=1.2, Hc=500000, K=1, M=10*Core1.Hc^(-1), d=0.5e-3, includeEddyCurrents=false, l=.1, sigma=10000000, useHeatPort=false) annotation(Placement(transformation(origin={150.0,100.0},extent={{-20.0,-20.0},{20.0,20.0}},rotation=-90)));
equation
   connect(S1.y, signalSource1.V_m) annotation(Line(points={{2.0,100.0},{32.0,100.0}},color={0,0,127},smooth=Smooth.None));
   connect(G1.port, signalSource1.port_n) annotation(Line(points={{50.0,50.0},{50.0,80.0}},color={255,127,0},smooth=Smooth.None));
   connect(signalSource1.port_p, Core1.port_p) annotation(Line(points={{50.0,120.0},{50.0,150.0},{150.0,150.0},{150.0,120.0}},color={255,127,0},smooth=Smooth.None));
[...]

PR #2461, opened 2018-02-08
Fixes issue #2395

eshmoylova:
Removed the initial condition on the derivative that causes issue #2395.


christiankral:
To put it the other way around: In case you keep the start value ```start=0```, I am fine to approve the proposal, since it will then only fix a design flaw.

eshmoylova:
I would be OK with keeping `start=0` and removing only `fixed=false`.

PR #2507, opened 2018-04-23
Replace outdated Documentation and ReleaseNotes Templates with generation script and template

beutlich:
* The documentation template is actually totally deprecated. For once we don't generate the PDF any more.
* A new release notes template will be added later once I have finished the new scripts for that.


dietmarw:
It would probably be good to *replace* the templates with your scripts in this PR. That way the transition is more obvious.

PR #2517, opened 2018-04-26
Remove all occurences of valign="top" from Documentation

dietmarw:
The attribute `valign="top"` has no effect on any of the tables and therefore just adds noise to the HTML code which is then also harder to read.

Also the attribute should not be added to `td` elements but rather defined via styles if really necessary.


dietmarw:
I am aware that there are some additional usages of `valign=bottom|middle` those should probably also be reevaluated but that takes more time than is worth it really. The main thing is that the User's Guide no longer mentions them and that the top ones are removed.

PR #2520, opened 2018-04-27
Using green toned colors in Mechanics.Translational

tobolar:


tobolar:
@beutlich You are absolutely right. I would wait for a week or two and when no changes regarding color selection will be suggested I would update the picures and documentation where needed.

tobolar:
I have found two more pictures to be adapted:
- Lib-Translational.png
- LineForceWithTwoMasses1.png

Issue #1578, opened 2017-01-15
Strategy on handling obsoleted/deprecated classes


modelica-trac-importer:
**Modified by dietmarw on 3 Oct 2014 07:42 UTC**
Currently the MSL contains almost 90 classes marked as obsolete/deprecated and with a notice that they will be removed soon. 

Now in my 9 years of following the MSL development I can not recall an actual removal having happened even once. So the question is what is the MA's strategy on handling obsolete/deprecated classes. I think I know the answer which is there is no strategy. So let me suggest the following:

 1. All classes that are deprecated/obsolete should be visually marked at least (by extending from `Modelica.Icons.ObsoleteModel`) from the **next** minor version on (E.g., obsoleted in MSL3.1.x --> extends from ObsoleteModel from at least MSL 3.2.x on).
 2. All classes that are deprecated/obsolete should be **removed** two minor versions later (E.g., obsolete model in MSL3.1.x will no longer be present in MSL3.3.x)
 3. The information is clearly and openly publicised and also deprecated/obsoleted classes should be clearly listed in the release notes of MSL together with the clear information **when** they will be removed. Best in its own section for each release version. Library developers will thus be able to update their own libraries in due time.
 4. When removed a conversion script (even though we still only have the unofficial Dymola variant script language available) should take care of renames where possible.

Following those steps no-one should really be able to complain about the MSL breaking their library. If it does then clearly the library developer have not tested their library with the MSL (at all I would say) and we as MA can not take care of the QA of other libraries. As it is the MA is struggling with its own QA enough as it is.

<sub>PS: \*Could any native English speaker (I'd accept AE aswell ;-) ) explain the correct usage of _obsolete_ vs. _deprecate_. I'm really struggling here to use the right one and so does the MSL as it seems. Thanks!\*</sub>

----

**Modified by dietmarw on 3 Oct 2014 07:40 UTC**
Currently the MSL contains almost 90 classes marked as obsolete/deprecated and with a notice that they will be removed soon. 

Now in my 9 years of following the MSL development I can not recall an actual removal having happened even once. So the question is what is the MA's strategy on handling obsolete/deprecated classes. I think I know the answer which is there is no strategy. So let me suggest the following:

[...]

PR #2532, opened 2018-05-16
Extend obsolete classes from ObsoleteModel

beutlich:
This is for preparation of #1578. All obsolete models should finally be marked by the ObsoleteModel icon.


beutlich:
I double checked Electrical and only found Modelica.Electrical.Machines.BasicMachines.Components.BasicTransformer

dietmarw:
I would say it might be worth the effort to make use of the `obsolete` annotation just to make users aware of it. Remember the next time the obsolete annotation becomes relevant would be the release **after** the next major release.

PR #2539, opened 2018-05-22
Decrease height in Modelica.Fluid.Examples.Tanks.TanksWithOverflow

### LINKED ISSUE #2060: Modelica.Fluid.Examples.Tanks.TanksWithOverflow does not respect variable bounds
**Reported by agnes&#x2e;ramle on 12 Sep 2016 11:24 UTC**
The model Modelica.Fluid.Examples.Tanks.TanksWithOverflow does only initialize if the max and min bounds are ignored and the initial values are outside of these bounds. 
An example of a variable with this problem is `overflow.port_a.p`.

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/2060

beutlich:
As proposed by @chria.
Close #2060.


casella:
Sounds good

casella:
Closes #2060, sorry if I first closed the PR without merging.

PR #2546, opened 2018-05-24
Revert "Remove all occurences of valign=top (#2517)"

beutlich:
This reverts commit 5591ebcedcb68b55b1c1fec4044dcb070475c9ae.

As stated by https://github.com/modelica/ModelicaStandardLibrary/pull/2517#issuecomment-391805726 there are tables where valign="top" has an effect. As it is rather tedious to figure out all tables with (long or explicit) line breaks, I'd rather propose to revert the changes of #2517.

### DIFF
--- a/Modelica/Blocks/Continuous.mo
+++ b/Modelica/Blocks/Continuous.mo
@@ -825,30 +825,30 @@ blocks inside the PID controller are initialized according to the following tabl
 </p>
 
 <table border=1 cellspacing=0 cellpadding=2>
-  <tr><td><strong>initType</strong></td>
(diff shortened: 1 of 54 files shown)


dietmarw:
Actually I like to explore better alternatives here. `valign` is not legal in HTML5 and we should be forward thinking. Even though the HTML version to use is not specified we should better use CSS and ideally in one central place.

beutlich:
Sure. CSS is the better and OO style way to go for. Unfortunately, nothing is yet specified and we would need to wait for [mo:#1893](https://trac.modelica.org/Modelica/ticket/1893) and new version of Modelica specification.

PR #2573, opened 2018-06-01
Perform syntax check using Travis CI

### LINKED ISSUE #2572: Add syntax checks for pull requests
I think it would be useful to have syntax checking setup to verify that all pull requests are at least a little bit sane. We could do this using some free service like Travis, or I could offer to build using Jenkins on the openmodelica.org servers since a syntax checking job is quite low on resources (the only advantage is pulling docker images faster since we cache them; travis has the advantage of being reasonably available even if it queues for some time and takes a little while to pull docker images it is reasonable if you base them on the same OS).

The things that would change is that there needs to be a file somewhere (`/Jenkinsfile` or `/ travis.yml`) in each branch that should be checked. 

The following is an example that would parse a single library (the old BioChem library in this case) using OpenModelica v1.12.0 and moparser:
[Jenkinsfile.txt](https://github.com/modelica/ModelicaStandardLibrary/files/2061303/Jenkinsfile.txt)
[travis.txt](https://github.com/modelica/ModelicaStandardLibrary/files/2061655/travis.txt) (skipped creating the mos-script for checking; that could either be a separate file or in the travis.yml).

sjoelund:
This uses OpenModelica v1.12.0 and Maplesoft's moparser to perform syntax
checking of the library. The OpenModelica job returns error on warnings
when it detects mismatching uses-annotations (which is not foolproof),
bad package.order, etc. The job could be enhanced to perform additional
checks or calculate some statistics such as number of classes in the
future.

close #2572

### DIFF
--- a/.travis.mos
+++ b/.travis.mos
@@ -1,0 +1,18 @@
+setModelicaPath(".");
+loadModel(ModelicaTest);
+loadModel(ModelicaTestOverdetermined);
+loadModel(ObsoleteModelica3);
+
+(numMessages,numError,numWarning):=countMessages();
+print(getErrorString());
+if numError<>0 or numWarning<>0 then
+  exit(1);
+end if;
+
+names := {typeNameString(cl) for cl in getClassNames(sort=true)};
+b := sum(names) == "ComplexModelicaModelicaServicesModelicaTestModelicaTestOverdeterminedObsoleteModelica3";
+
+if b and numError==0 and numWarning==0 then
+  exit(0);
+end if;
+exit(1);
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+language: generic
+services:
+  - docker
+before_install:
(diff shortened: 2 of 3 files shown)


sjoelund:
Note that the check is failing because there is an inconsistency in the library already (I assume that ObsoleteModelica3 3.2.3 should use MSL 3.2.3):
```
ObsoleteModelica3.mo:  annotation (uses(Modelica(version="3.2.2")),
```

beutlich:
Can you please update your PR to fix the uses annotation in ObsoleteModelica3.mo.

sjoelund:
There. Success

sjoelund:
I should probably mention that the two docker runs share the same working directory, which is not ideal especially if more checks are added later.

thorade:
I assume $PWD is a password, configured in Travis? Then the script will not work in forks, correct?
Could it be set from the script, such that forks can also run the tests (if the fork owner enables Travis)?

sjoelund:
It's not a password; it is the working directory.

PR #2582, opened 2018-06-03
FluidHeatFlowEnhancements

AHaumer:
Enhancements: output variables in examples and unified isolatedPipe and heatedPipe to ease further maintenance (isolatedPipe and heatedPipe marked obsolete, use Pipe with conditional heatPort instead)

### DIFF
--- a/Modelica/Thermal/FluidHeatFlow.mo
+++ b/Modelica/Thermal/FluidHeatFlow.mo
@@ -1735,8 +1735,7 @@ Block generating the sum of two ramps.
   package Components "Basic components (pipes, valves)"
     extends Modelica.Icons.Package;
 
-    model Pipe "Pipe with/without heat exchange"
-
+    model Pipe "Pipe with optional heat exchange"
       extends Interfaces.Partials.TwoPort;
       parameter Boolean useHeatPort = false "=true, if HeatPort is enabled"
         annotation(Evaluate=true, HideResult=true, choices(checkBox=true));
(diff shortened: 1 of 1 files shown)


AHaumer:
Well, how to proceed? 
Switch back to before merge of FluidHeatFlowDesignFlaw? 
Posssible but stagnation.
Switch back to after merge of FluidHeatFlowDesignFlaw?
No good idea, I don't get rid of thhe DesignFlaw (future backwards compatibility).
Merge the new FluidHeatFlowEnhancement?
Best option. Stable now after my latest commits.

AHaumer:
Sorry for the troubles but I'll end in a mess with FluidHeatFlow if we can't finalize this now. 
I use FluidHeatFlow a  lot in projects with students.

dietmarw:
The problem is that you seem to repeatedly submit finished versions only to fix them the next day again. So that's why I think it is a good idea to let this settle. I don't doubt its usefulness but am very hesitant to include new features that are little tested. By little I mean not even been available more than 24 hours in that final (if it is final) shape.
So Option 1 seems at this point in time (after 1 week after the first beta) the only sensible solution.

christiankral:
As the applied changes definitely improved the re-usability of code by applying more object orientation, I think a significant improvement was made. I would definitely not step back to a previous code basis!

dietmarw:
Just to emphasise my point that this is not ready yet. There are several (new) models that use obsolete components `FlowPort_a|b`. This does not make sense unless the model is marked obsolete itself. So there is still some clean up work to be done.

dietmarw:
Actually I take that back. It was simply the icon looked similar to the Obsolete frame. So that part is fine.

PR #2583, opened 2018-06-04
Reduce git checkout depth and avoid mail notification at Travis CI

beutlich:


.travis.yml
```diff
@@ -1,9 +1,18 @@
 language: generic
+
+git:
+  depth: 5
```
dietmarw: Why not simply 1?

beutlich: Just my default setting. If you have parallel jobs and push while not all jobs of previous build are started, it would fail for depth=1.

dietmarw: OK

PR #2584, opened 2018-06-04
Switch to secure website for haumer.at

beutlich:
There is no automatic redirection, thus make it explicit


Modelica/Electrical/MultiPhase.mo
```diff
@@ -91,15 +91,14 @@ This set of symmetrical components is repeated for each of the n<sub>Base</sub>
 <p>
 <strong>Dr. Christian Kral</strong><br>
 Electric Machines, Drives and Systems<br>
-1060 Vienna, Austria<br>
+1060 Vienna, Germany<br>
```
christiankral: Vienna is not in Germany

PR #2562, opened 2018-05-29
#2430 revised adaptor icons and nameing

StephanZiegler:
Following tobolar's suggestions in:

1. Icons:
- removed shading in icons for improved clearity in partial adaptor
- added colored rounded rectangle in domain specific adaptors
- unified sizing and location of real input icons 

2. Naming:
- renamed XToYAdaptor to GeneralXToYAdaptor

close #2430

Issue #2569, opened 2018-05-30
Default component name for adaptors


tobolar:
Flow to potential adaptors (and potential to flow) have different settings concerning default component name. In particular:
- `GeneralCurrentToVoltageAdaptor` and `GeneralVoltageToCurrentAdaptor` have no defaut names,
- `GeneralAngleToTorqueAdaptor` and `GeneralTorqueToAngleAdaptor` have default names omitting "general" prefix,
- `GeneralPositionToForceAdaptor` and `GeneralForceToPositionAdaptor` have no defaut names,
- `HeatflowToTemperatureAdaptor` and `TemperatureToHeatflowAdaptor` have default names "adaptor".

This should be harmonised. I prefer the version which omits "general" prefix in component name.

AHaumer:
see PR #2570

PR #2585, opened 2018-06-04
Fix rotational adaptor icons

tobolar:
refs #2562

### DIFF
--- a/Modelica/Mechanics/Rotational.mo
+++ b/Modelica/Mechanics/Rotational.mo
@@ -5562,20 +5562,21 @@ blocks of the block library Modelica.Blocks.Sources.
               radius=10,
               lineThickness=0.5),
             Text(
-              extent={{-20,90},{20,68}},
+              extent={{-20,92},{20,70}},
               fillPattern=FillPattern.Solid,
-              textString="phi"),Text(
+              textString="phi"),
+            Text(
               visible=use_w or use_a,
-              extent={{-20,40},{20,18}},
+              extent={{-20,62},{20,40}},
               fillPattern=FillPattern.Solid,
               textString="w"),
             Text(
               visible=use_a,
-              extent={{-20,-20},{20,-42}},
+              extent={{-20,32},{20,10}},
               fillPattern=FillPattern.Solid,
               textString="a"),
             Text(
-              extent={{-20,-70},{20,-92}},
+              extent={{-20,-68},{20,-90}},
               fillPattern=FillPattern.Solid,
               textString="tau"),
             Text(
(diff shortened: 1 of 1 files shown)


tobolar:
Actually, my updates concern PR #2562 which already have been merged. There is no connection to #2569 except that the same components are affected. But the topic is different.

StephanZiegler:
I added the rounded rectangular to the icon for consistency. Looks fine for me.

tobolar:
The original idea was to only revert changes on the two adaptors. But having rectangle's border of double width is fine for consistency reasons.

@AHaumer and @StephanZiegler Please review again.

tobolar:
I don't understand. On one side you have changed rectangle's border width in de2a8984d8a9a1da98e76ed7d870366c9ef5337f to be consistent with general adaptors but on the other side you want to change position of description strings in c302514b1a16a85adb327054108507ffec8574dd which in turn is absolutely inconsistent.

I have to stress again that the original idea of this PR was to only revert changes on the two adaptors. Not to experiment on new icons for them.

StephanZiegler:
So maybe i got the meaning of this ticket wrong in the first place. 
Under "Fix" I understand repair/improve or in in this case "make it matching". I'm sorry if this is considered nitpicking, but my opinion is, if we put effort into if, we should do it properly.

Regarding your statement:

> ...but on the other side you want to change position of description strings in c302514 which in turn is absolutely inconsistent. 

In My tool (Dymola 2018FD01) this looks like this:
before:
adaptor_before

and after:
adaptor_after

So for me the the input/output names are aligned after the changes. In the first version the names are just spread over the icon, which is in my opinon not consistent with the other adaptors.
Does it look different on your tool?
Why do you consider it inconsistent? Is there any modeling or styling rule applying that i am not aware of?
Please clarify.

General question: if the changes should just be reverted, is a review required in the first place?

dietmarw:
@StephanZiegler For icon design, the new MSL that we are working on has now a guideline. Check `Modelica.UsersGuide.Conventions.Icons` from master.

tobolar:
@StephanZiegler Sorry, I was confused since supposed your commit is to this PR - but it is a separate one, isn't it? (I wonder this is even possible due to double review process.)

Your changes on description strings are fine, of course.

Nontheless I would perform your changes on my fork again and commit it on my own. Then you can review again.

StephanZiegler:
@tobolar yes, sorry i pushed my changes wrongly. 
I wanted to wait for the result of this discussion before pushing and maybe causing even more confusion. 

Yes commit to your fork and i will review again.

tobolar:
@StephanZiegler  GitHub is still a mystic thing for me.

StephanZiegler:
Look fine on my end now.

AHaumer:
I have no strong preference on the icons but I'd prefer to be able to distinguish the icon of the "old" adaptors and the "new" generic adaptors (in Mechanics.Rotational). And I would prefer to have a consistent look (maybe with domain-specific colours) in Electrical.Analog, Mechanics.Rotational, Mechanics.Translational and Thermal.HeatTransfer.

tobolar:
@AHaumer IMHO this is the case now.

Issue #1865, opened 2017-01-15
Consistent use of terms multi-phase and single-phase


modelica-trac-importer:
**Reported by beutlich on 16 Dec 2015 23:07 UTC**
The terms multi phase / single phase and multiphase / singlephase are used within the Electrical library. It would be more standard-like if they would be used consistently. Also multi-phase / single-phase are possible. Some native speakers may advise here.

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/1865

modelica-trac-importer:
**Comment by ahaumer on 18 Dec 2015 12:13 UTC**
You're right. 
Since this is no big issue, I'm waiting for native speakers' advise (shifting the mielstone).

ceraolo:
I think that more than native speakers what counts is IEC's opinion.
On the International Electrotechnical vocabulary we find single-phase :
http://dom2.iec.ch/iev/iev.nsf/SearchView?SearchView&Query=field+SearchFields+contains+single-phase+and+field+Language=en&SearchOrder=4&SearchMax=0

beutlich:
So: single-phase and polyphase are to be used, right?

ceraolo:
I agree.
IEV is the highest source of information for electricity terms.
They use single-phase and polyphase.

beutlich:
I closed #2290 due to plenty of merge conflicts. Feel free to reopen in case you want to have consistent terms one day.

christiankral:
I discussed this issue with @ahaumer and double checked with my English text books on electric machines and systems. So on my opinion the term polyphase seems to be the right one to be used. After implementing the terms in the MSL correctly, we will have to consider renaming the sub library Electrical.MultiPhase to Electrical.PolyPhase in the next major release.  

@beutlich Sorry for corrupting your proposal. I was not not aware of that when I worked on #2276.

PR #2588, opened 2018-06-05
Improve wording of electrical and magnetic connector descriptions

christiankral:
The changes are based on the discussion of #2560. Due to the added and improved doc strings of the connectors, the examples of the MSL (and other user libraries) are much better to understand. The tooltips associated with the connectors now show a much more complete and accurate description of the connectors.

### DIFF
--- a/Modelica/Electrical/Analog/Interfaces.mo
+++ b/Modelica/Electrical/Analog/Interfaces.mo
@@ -4,7 +4,7 @@ package Interfaces
   extends Modelica.Icons.InterfacesPackage;
 
   connector Pin "Pin of an electrical component"
-    SI.ElectricPotential v "Potential at the pin" annotation (
+    SI.ElectricPotential v "Potential of the pin" annotation (
         unassignedMessage="An electrical potential cannot be uniquely calculated.
 The reason could be that
 - a ground object is missing (Modelica.Electrical.Analog.Basic.Ground)
@@ -44,7 +44,7 @@ The reason could be that
   end Pin;
 
   connector PositivePin "Positive pin of an electric component"
-    SI.ElectricPotential v "Potential at the pin" annotation (
+    SI.ElectricPotential v "Potential of the pin" annotation (
         unassignedMessage="An electrical potential cannot be uniquely calculated.
 The reason could be that
 - a ground object is missing (Modelica.Electrical.Analog.Basic.Ground)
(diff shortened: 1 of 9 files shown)


dietmarw:
@christiankral Please be careful in future and do **not** create branches at modelica/ModelicaStandardLibrary but your fork.

christiankral:
@dietmarw OK, the Modelica branch is a result of syncing the upstream repository -- sorry. When pushing the branch the git interface proposed to apply `git push --set-upstream origin wording-interfaces-3.2.3` which is apparently not what I should have done. How can I then only push it to my (synced) fork? Do you know what the command git is?

christiankral:
I hope the wording is fine now: also added "quasi-static" to the User's Guide spelling section.

dietmarw:
The only other alternative is that you create a new PR based on your branch and we close this one. But I would prefer you pushing just that last commit.

christiankral:
@dietmarw You're right; my fault. It is now pushed upstream to modelica/ModelicaStandardLibrary.

AHaumer:
Just to clarify some questions:
I'm not convinced that polyphase is the better term than multiphase. I guess both terms are used in English literature.
Regarding plug and pin: pin is a single phase electric connector, plug is a multi-/poly-phase connector containing m pins. 
Regarding the wording: I have no strong preference whether to use an article (the) or not. 
When we're talking about a potential (remember, a potential variable can ONLY be a potential quantity, not a potential difference!) I would use "at the connector". For a flow variable I would use "into the connector" (from outside).

dietmarw:
We already had a discussion on polyphase vs multiphase in #1865 were you @AHaumer also agreed. So we should stick to polyphase.

tobolar:
> Moreover, there is "electrical pin" but "electric plug".

I wanted to stress _electrical_ against _electric_ in that review, but found no possibility to highlight it by text format - obviously no such possibility in review window.
> Regarding the wording: I have no strong preference whether to use an article (the) or not.

I rather mean general usage of articles in description strings within whole MSL. I guess they are generally not used.

christiankral:
OK, changed "electric" to "electrical" in the connector descriptions. 

Please note, that the terms are not yet entirely changed throughout all libraries. I just wanted to improve the wording of the connectors, so far. I will continue on the unification later.

PR #2592, opened 2018-06-05
Update release notes for beta.2

beutlich:
* list of resolved GitHub issues (including pull requests) generated by `python Generate-ReleaseNotes.py`
* comparison table between v3.2.3 and v3.2.2 generated by `ModelManagement.Compare.compareModels("Modelica323", "Modelica 3.2.3", "Modelica322", "Modelica 3.2.2", "DifferencesTo322.html", true)`

### DIFF
--- a/Modelica/Resources/Documentation/Version-3.2.3/DifferencesTo322.html
+++ b/Modelica/Resources/Documentation/Version-3.2.3/DifferencesTo322.html
@@ -11136,6 +11136,36 @@ style="font-weight: bold;">Modelica 3.2.2</span><br>
 </table>
 
 <br><hr style="width: 100%; height: 6px;"><br>
+<big style="font-family: Arial;">model Fluid.Pipes.BaseClasses.FlowModels.PartialGenericPipeFlow</big><br><br>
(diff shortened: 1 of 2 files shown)


AHaumer:
Sorry nearly impossible to track changes that late after a long day. Tomorrow I'm travelling, you might add other reviewers.

christiankral:
Why are we using font "Arial" in the HTML code? I suppose, the HTML editor of Dymola is inserting this, but maybe I am wrong... However, I think we should rather avoid "Arial" -- which I am not using anyways, as Linux user -- and we should avoid any other font name and leave this up the HTML engine.

I will have a look at the content around 10-11 this morning.

beutlich:
> Why are we using font "Arial" in the HTML code? I suppose, the HTML editor of Dymola is inserting this

Indeed, it is generated by running ModelManagement.Compare.compareModels in Dymola on Windows. That is also how it is in MSL v3.2.2, see e.g., https://doc.modelica.org/Resources/Documentation/Version-3.2.2/Differences322To321Build4.html.

PR #2593, opened 2018-06-06
Reintroduce UTF-8 characters that were removed for v3.2.3 release

dietmarw:
For the v3.2.3 release some non-ASCII characters were removed (as part of #2587) in order to keep the MSL ASCII. The next MAJOR version should allow UTF8 encoding and hence also these characters again.


dietmarw:
I just noticed that there is a nice side effect, now that the ASCII check via CI is in place. This PR will automatically stay blocked until we switch the CI to allow UTF-8 :smile:

beutlich:
This should go along with updating moparser to v3.4, see https://trac.modelica.org/Modelica/changeset/9902

beutlich:
@sjoelund @christiankral Can you please review?

PR #2594, opened 2018-06-07
Mark Modelica.Fluid.Icons as obsolete

beutlich:
This is in addition to #2532 to also mark the Icon package obsolete.


beutlich:
If no domain-specific icons will be added, we have an empty Icons package then. You still can set it obsolete now and reintroduce later (non-obsolete) once there are domain-specific icons.

dietmarw:
No if now icons are added you can simply remove that Icon package together with the content because nobody would have used Icon package itself is what I am saying.

Issue #1990, opened 2017-01-15
Modelica.Utilities.Examples.ReadRealMatrixFromFile issue


modelica-trac-importer:
**Reported by peter&#46;lundberg on 28 Apr 2016 08:21 UTC**
The variable:

`parameter String file = ...`

is declared but never used.
The first two uses of (yet undeclared) "file1" should probably be replaced by "file".

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/1990

beutlich:
Resolved in master and maint/3.2.2.

PR #2598, opened 2018-06-11
Exclude regression issues from generated Release Notes

beutlich:


beutlich:
It is `V: 3.2.2-XXX`, not `V: v3.2.2-XXX`.

sjoelund:
But it seems that the currently closed issues include things like #1990 which is a bug intended to target the milestone 3.2.2 and is not a bug in 3.2.2, right?

beutlich:
#1990 is a bug in v3.2.2 and targeted for v3.2.3. All fine.

beutlich:
The idea is to exclude (once closed) https://github.com/modelica/ModelicaStandardLibrary/labels/V%3A%203.2.3-beta.2 from the Release Notes for v3.2.3.

PR #2599, opened 2018-06-11
Update moparser to v3.4

sjoelund:
This updates the moparser docker image used to moparser 3.4; it still uses the 3.2 grammar. The base image was updated to a slim Debian; 32-bit support is no longer necessary since moparser comes with 64-bit support.

Small note: moparser depends on GNU libc and libstdc++, so Alpine Linux for 2MB base images is not an option although I did try that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,12 +8,12 @@ services:
 
 before_install:
   - time docker pull openmodelica/openmodelica:v1.12.0-minimal
-  - time docker pull openmodelica/moparser:xenial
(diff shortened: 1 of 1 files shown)


dietmarw:
@sjoelund I normally use https://github.com/phusion/baseimage-docker as basis for my docker images. This might be a better alternative.

sjoelund:
Phusion might be good, but for just running a single 230 kB executable it seems overkill at 77MB (Debian 9.4-slim is 22MB).

PR #2605, opened 2018-06-11
Smooth line in icon of Modelica.Mechanics.Rotational.Sources.EddyCurrentTorque

beutlich:


beutlich:
> But I would prefer to wait with such cosmetic changes after v3.2.3 release.

Why? EddyCurrentTorque is new for v3.2.3 and we should force ourselves to make it right from the beginning.

tobolar:
It's just for my personal availability at the moment.

AHaumer:
I agree with tobolar. This release process is too much in a hurry.

PR #2618, opened 2018-06-12
Add files to ignore upon archive to .gitattributes.

dietmarw:


dietmarw:
The problem with ModelicaTest is that **if** we would ship that too then we also would need to take care of backward compatibility and conversion scripts for that (developer) library. So that is really out of a question.

If there really are very valuable examples one could argue to move them under MSL instead.

dietmarw:
Btw, wait with the merge since I think I have som nice additional renaming feature I can add.

dietmarw:
Actually renaming feature can not so easily (or not at all) implemented inside `.gitattributes` so I'll publish a release script instead for that and link it in the [Wiki](https://github.com/modelica/ModelicaStandardLibrary/wiki/Generating-a-new-MSL-release#automated-generation). So will merge this as is.

dietmarw:
And now that I merged it I notice that the `export-subst` definition is actually not useful since we exclude it anyway ... oh well.

PR #2621, opened 2018-06-12
Fix text positions in FluidHeatFlow icons

### LINKED ISSUE #2620: Broken icon for Modelica.Thermal.FluidHeatFlow.Sources.Ambient
The icon for Modelica.Thermal.FluidHeatFlow.Sources.Ambient changed from 3.2.2 to 3.2.3. In 3.2.2 the text was on top of the circle:
ambient3 2 2
In 3.2.3 the text got covered by the circle:
ambient3 2 3

The difference is that in 3.2.2 the circle was inherited from the partial class it extended from and the text added after. In 3.2.3 the class it extends from has changed and it does not contain the circle. The code to draw the circle was added after the code that specifies the text. Starting on line 2540:
https://github.com/modelica/ModelicaStandardLibrary/blob/e69a99ad2abd6ba6a5006d50e896dca97f2dee03/Modelica/Thermal/FluidHeatFlow.mo#L2540
```modelica
</html>"), Icon(coordinateSystem(preserveAspectRatio=true, extent={{-100,-100},{
                100,100}}), graphics={Text(
              extent={{20,80},{80,20}},
              textString="p"), Text(
              extent={{20,-20},{80,-80}},
              textString="T"),         Ellipse(
              extent={{-90,90},{90,-90}},
              lineColor={255,0,0},
              fillColor={0,0,255},
              fillPattern=FillPattern.Solid)}));
```
The suggestion is to change the order in which the circle and the text are written. The above block would be:
```modelica
</html>"), Icon(coordinateSystem(preserveAspectRatio=true, extent={{-100,-100},{
                100,100}}), graphics={Ellipse(
              extent={{-90,90},{90,-90}},
              lineColor={255,0,0},
              fillColor={0,0,255},
              fillPattern=FillPattern.Solid), Text(
              extent={{20,80},{80,20}},
              textString="p"), Text(
              extent={{20,-20},{80,-80}},
              textString="T")}));
```

beutlich:
close #2620

### DIFF
--- a/Modelica/Thermal/FluidHeatFlow.mo
+++ b/Modelica/Thermal/FluidHeatFlow.mo
@@ -2354,7 +2354,7 @@ Record containing (constant) medium properties.
 </html>"),
         Icon(coordinateSystem(preserveAspectRatio=true, extent={{-100,-100},{100,100}}),
           graphics={Text(
-              extent={{-22,-20},{20,-60}},
+              extent={{-30,-20},{30,-60}},
               lineColor={0,0,255},
               textString="p")}));
     end PressureSensor;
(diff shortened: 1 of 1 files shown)


beutlich:
> Similar corrections should be done e.g. in `Electrical.Analog.Sensors` as well.

The electrical sensors are fine from my since they only use one-letter text strings

> Moreover, what about the text color? I guess black is the preferred one.

OK, fc05d5c564ac75bb0a80d25664a700fde2734c9a changes it to black.

PR #2622, opened 2018-06-12
Add default component name to old rotational adaptors

beutlich:
Have identical default components names of old and new rotational adaptors to better support tools.


beutlich:
Reason: Tools create angleToTorqueAdaptor for the new adaptor, but angleToTorqueAdaptor1 for the old adaptor. This PR harmonizes the behaviour.

PR #2631, opened 2018-06-13
Fix recently added HTML errors and missing links 

dietmarw:


casella:
Not sure about adding ModelicaTest as a release asset. It is not meant as a reusable component library, but more for library and tool developers. I'm fine with the external link

dietmarw:
Yes exactly like @casella says. The ModelicaTest should stay dev only since it is not properly documented. Also tools would have to bundle the ModelicaTest in addition which might not happen and should be discussed anyway. So I like to keep it now as it is. Can you live with that @beutlich?

dietmarw:
Probably a good idea which also makes it then easier later on to identify. PR follows.

PR #2657, opened 2018-06-20
Upgrade icon of robot r3

tobolar:
refs comment by @eshmoylova on commit 9a6f157d078a4807fbc0d86315da7045cf995129


beutlich:
Should it read `Upgrade diagram of robot r3` since you touched not only the icon?

Modelica/Mechanics/MultiBody/Examples/Systems/RobotR3.mo
```diff
@@ -1663,35 +1676,35 @@ This model contains the mechanical components of the r3 robot
...
               textString="4",
-              lineColor={0,0,255}),
+              lineColor={0,0,0}),
             Text(
-              extent={{-198,150},{-138,110}},
+              extent={{-190,110},{-130,70}},
+              textString="5",
+              lineColor={0,0,0}),
+            Text(
+              extent={{-188,170},{-128,130}},
```
eshmoylova: The difference is small enough that it shouldn't matter. But just for consistency with others shouldn't it be {{-190,170},{-130,130}}?

tobolar:
@eshmoylova Fixed. Thanks!

PR #2678, opened 2018-06-28
Add missing `each`

casella:
Yet another missing 'each' in `MixtureGasNasa.` 

Causes OpenModelica to fail on [Modelica.Media.Examples.TestOnly.FlueGas](https://libraries.openmodelica.org/branches/newInst/Modelica_trunk/files/Modelica_trunk_Modelica.Media.Examples.TestOnly.FlueGas.err) an on similar components


casella:
One more missing `each` added. We should really finally get rid of these issues in MSL 3.2.3...

casella:
I catched two more missing 'each' in Modelica.Magnetics using OpenModelica

casella:
I just looked at the the [reports using the new front-end](https://libraries.openmodelica.org/branches/newInst/), which add the `-d=newInst` flag. The new front end is not tolerant of missing 'each'.

casella:
@beutlich, I just need your review before the 8 Aug deadline to merge this 😃

PR #2682, opened 2018-06-29
resolves #2675

### LINKED ISSUE #2675: Temperature jumping to 0K with V_flow = 0
Using Dymola 2019 FD01 Dev1 with the model [CoolingCircuit.zip](https://github.com/modelica/ModelicaStandardLibrary/files/2144475/CoolingCircuit.zip) the temperature exhibits discontinuities when the volume flow is set to 0 (temperature jumps to 0K). 

Notes:
- I used MSL Version: 3.2.3, 2018-08-01, build 1 (2018-08-01 12:00:00Z) - that comes with Dymola 2019 FD01 Dev1
- The problem was not existant with MSL 3.2.2 in Dymola 2019 FD01 Dev1 or Dymola 2019.
- The problem does exist in Dymola 2019 and 2018 FD01 with MSL 3.2.3
- The problem is resolved when the relative pressure sensor is removed from the simulation.

AHaumer:
Switched back from inheritance from TwoPort for sensors.
In the same run anhanced (I hope all) instations to full qualified names.

Close #2675

### DIFF
--- a/Modelica/Thermal/FluidHeatFlow.mo
+++ b/Modelica/Thermal/FluidHeatFlow.mo
@@ -2691,7 +2691,7 @@ leads to neglect of temperature transient cv*m*der(T).
         dp = 0;
         flange_a.tau = 0;
       else
-        dp = -dp1*(1-V_flow/V_flow1);
+        dp = -dp1*(1-noEvent(abs(V_flow/V_flow1)));
         flange_a.tau*w = -dp*V_flow;
       end if;
       // no energy exchange with medium
--- a/Modelica/Thermal/FluidHeatFlow.mo
+++ b/Modelica/Thermal/FluidHeatFlow.mo
@@ -182,17 +182,17 @@ This library provides simple components for 1-dimensional incompressible thermo-
         "Source over Coolant";
       output Modelica.SIunits.TemperatureDifference dTCoolant=pipe.dT
         "Coolant's temperature increase";
-      FluidHeatFlow.Sources.Ambient ambient1(constantAmbientTemperature=TAmb, medium=medium,
+      Modelica.Thermal.FluidHeatFlow.Sources.Ambient ambient1(constantAmbientTemperature=TAmb, medium=medium,
         constantAmbientPressure=0)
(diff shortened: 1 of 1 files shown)


Modelica/Thermal/FluidHeatFlow.mo
```diff
@@ -2692,7 +2691,7 @@ leads to neglect of temperature transient cv*m*der(T).
         dp = 0;
         flange_a.tau = 0;
       else
-        dp = -dp1*(1-noEvent(abs(V_flow/V_flow1)));
+        dp = -dp1*(1-V_flow/V_flow1);
         flange_a.tau*w = -dp*V_flow;
```
dietmarw: Why this change? Is this related to the regression?

AHaumer: No. It was a flaw since the beginning. Should I leave the flaw in?

dietmarw: It would have been good to not mix this fix with the fix that should purely rectify the regression problem we are facing. Otherwise it becomes more difficult to pinpoint sources of regressions. It's probably fine to leave it in now but for the future please isolate each dedicated fix in its own commit since it can then be removed or re-applied much easier.

beutlich: @AHaumer Please remove this additional fix from this PR and file a separate PR for it, to make it show up in the Release Notes.

Since `abs(v)` expands to `noEvent (if v >= 0 then v else –v)` there is no need to have the `noEvent` operator around the `abs` function.

AHaumer: see 5ce114d (my fork)

AHaumer:
It seems that I mixed something up (sorry I'm not really familiar with git):
Please merge 5ce114d to fix #2675, and 497f7b9 to fix #2687.
I tried to create a new branch (in my form) for #2687 but it went to the same branch ... sorry.

AHaumer:
ok I managed to undo the unwanted change (don't ask how), so 96199d1 should be ready to merge

AHaumer:
and I made a separate PR #2688

PR #2701, opened 2018-08-06
Fix equation of pressure increase versus volume flow in IdealPump

### LINKED ISSUE #2549: FluidHeatFlow.Sources.IdealPump: equations have a flaw
The dependency of pressure increase versus volumeFlow is (according to Euler) linear without the abs around volumeFlow.
To get volumeFlow negative, the pressure increase has to get higher than at zero flow.

### LINKED ISSUE #2687: IdealPump has an erroneous equation
FluidHeatFlow.Sources.IdealPump has an erroneous equation. In many cases, this won't be detected but it should be fixed.

Additionally, I'll revoke the PumpTurbine model (new in 3.2.3-beta.1, see 0ade4359bd76c44cfd6d20e3102f606fd6283708), since it is the same as IdealPump (with that fix). Enhancing the new PumpTurbine would cause backwards compatibility issues, it is better to bring it into the next release.

AHaumer:
fixes the erroneous equation in FluidHeatFlow.Sources.IdealPump

Close #2687
Close #2549

### DIFF
--- a/Modelica/Thermal/FluidHeatFlow.mo
+++ b/Modelica/Thermal/FluidHeatFlow.mo
@@ -2596,7 +2596,7 @@ leads to neglect of temperature transient cv*m*der(T).
         dp = 0;
         flange_a.tau = 0;
       else
-        dp = -dp1*(1-noEvent(abs(V_flow/V_flow1)));
+        dp = -dp1*(1 - V_flow/V_flow1);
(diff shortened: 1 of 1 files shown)


beutlich:
Removed the `V: 3.2.2` label since the issue is present since 60f640b170704a713dfb1d05137f3894e03def43 (MSL v2.2.1) where IdealPump was added.

PR #2702, opened 2018-08-13
Move ModelicaFileType to header ModelicaInternal.h

beutlich:
This enables the reuse the ModelicaFileType enum values for return value comparisons of ModelicaInternal_stat.

In addition to #2052.


beutlich:
Currently I have in my code `ModelicaInternal_stat(buf) == 2`, but want to use `ModelicaInternal_stat(buf) == FileType_RegularFile`.

> None of the functions return this enumeration; only an integer.

Yes, due to the Modlica C API which only knows about integer. But in fact, ModelicaInternal_stat returns ModelicaFileType .

beutlich:
> Perhaps there should even be two versions of the function exported so it is possible to explicitly call the function using the enumeration in internal code

This is already the case (by static function Internal_stat), but was not desired here.

PR #2704, opened 2018-08-22
ModelicaIO::readTxtTable Avoid message on non-matching table dimensions

beutlich:
If there is a line between tables that is all white space (blanks or tabs), no message about non-matching table dimensions shall be printed. The character index needs to be reset for every new line.

### DIFF
--- a/Modelica/Resources/C-Sources/ModelicaIO.c
+++ b/Modelica/Resources/C-Sources/ModelicaIO.c
@@ -917,6 +917,7 @@ static double* readTxtTable(_In_z_ const char* fileName, _In_z_ const char* tabl
                     unsigned long lineNoPartial = lineNo;
                     int tableReadPartial = 0;
                     while (readLine(&buf, &bufLen, fp) == 0) {
+                        k = 0;
                         lineNoPartial++;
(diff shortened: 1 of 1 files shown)


beutlich:
Some more explanation while there have not been any public reviews. PR fixes two problems that can be reproduced with the following table (note the 5 tabulators at the line between the two tables)

```
#1
double tab1(2,2)
0 1
1 2
					
double tab2(2,2)
0 1
1 2
```

1. Character index was incremented to 5 when parsing line no. 5 and was not reset to 0 when parsing the next line no. 6.
2. Line no. 6 starts with the word `double`, where the word was tokenized at the k-th position, thus at the character `e`. This token `e` was checked if it represents a floating-point number in function `isNumber`, which the test succeeded, where it should not.

beutlich:
The `isNumber` test now fails correctly for the following tokens: `e1.`, `1e1.`, `e1.1`, `1e1.1`, `e`, `1e`, `e+`, `e1`, `.e`, `e.`, `1e+`, `1.e+`
It still succeeds for the follwoing tokens: `1.e1`, `.1e1`, `1.`, `.1`

Issue #2696, opened 2018-08-01
Icon of models with states


tobolar:
There are defined models in MSL dealing with states:
- Magnetic.FundamentalWave.Interfaces.StateSelector
- Mechanics.Rotational.Components.RelativeStates
- Mechanics.Translational.Components.RelativeStates

They all use similar icon design but differ in the color of the background circle element.
I wonder whether this is done on purpose or by mistake. If the latter applies, one color should be defined for such cases.

![grafik](https://user-images.githubusercontent.com/9588978/43518698-216c8d14-958d-11e8-9966-fdcd7b3d87a0.png)

Btw. they also differ in the component name position and in the size of the circle.

tobolar:
I have changed (in #2710)
- Mechanics.Rotational.Components.RelativeStates
- Mechanics.Translational.Components.RelativeStates

to have similar look.

In contrast, no changes have been made to `Magnetic.FundamentalWave.Interfaces.StateSelector` since the model's purpose is indeed different.

Suggestion: close this issue when #2710 is merged.

beutlich:
Closinf as proposed by @tobolar .

PR #2710, opened 2018-09-03
RelativeStates icon

tobolar:
Icons of RelativeStates models from both Mechanics.Rotational and Mechanics.Translational have similar icons

refs #2696


beutlich:
OK for 3,2,3, even though you did not set assignees and milestone?

tobolar:
The changes are not critical I think. And the images used in the documentation are updated as well. So I believe there should be no problem.

PR #2711, opened 2018-09-04
Upgrade icon of rotational relative states

tobolar:


beutlich:
@AHaumer @MartinOtter Can you add your please review.

AHaumer:
Would be helpful if you upload images comparing before - after (similar as the issue with GE and LE).

tobolar:
> Would be helpful if you upload images comparing before - after (similar as the issue with GE and LE).

I will consider this for the next PRs.

PR #2713, opened 2018-09-24
Fix false extent of text "name" in icon

tobolar:


tobolar:
@beutlich You are right! I will fix this as well.

tobolar:
Please review again.

PR #2730, opened 2018-10-08
Set guess values of ModelicaTest.MultiBody.Joints.SphericalSpherical2

### LINKED ISSUE #2705: Initialization of ModelicaTest.MultiBody.Joints.SphericalSpherical2
I'd like to apologize in advance in case I have missed something obvious in this model, but I don't understand why the model is expected to initialize with a the positive solution for the _x_ coordinate of the `bodyBox`.  All I can see is that the other two coordinates are supposed to initialize to 0 due to the `freeMotionScalarInit`, but I don't see any initialization at all for the _x_ coordinate.  As I understand it, this leads to an initial guess of 0 for the quadratic equation for satisfying the length constraint of the `sphericalSpherical`.  Since 0 is equally close to the negative and positive solutions, I don't see why one would expect to get the positive solution.

tobolar:
Close  #2705

### DIFF
--- a/ModelicaTest/MultiBody.mo
+++ b/ModelicaTest/MultiBody.mo
@@ -6137,37 +6137,35 @@ and plot gasForce.press over gasForce.s_rel.
 
       Modelica.Blocks.Interfaces.RealOutput r1[3]
         "Relative position vector frame_b.r_0 - frame_a.r_0 resolved in frame defined by resolveInFrame"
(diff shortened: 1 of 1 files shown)


beutlich:
@tobolar Do yo agree to merge only 320f904 for v3.2.3 and cca9b3e04fb12f0e65ec8248bc021c21393fcef8 for the next major release?

tobolar:
Fine to me.

PR #2733, opened 2018-10-11
Add missing `each` in Magnetic.FundamentalWave

perost:
- `ir` is an array, and so `each` should be used in
  Modelica.Magnetic.FundamentalWave.Examples.BasicMachines.SMEE_DOL.


casella:
@beutlich  Can we sneak this in 3.2.3?

PR #2737, opened 2018-10-14
Add missing stop time

beutlich:


beutlich:
This PR is obsolete in case #2738 gets merged.

PR #2738, opened 2018-10-14
Move eddy current brake test models to ModelicaTest

beutlich:
Plain test models belong to ModelicaTest, especially, if there already is a TestBraking test model.


beutlich:
This PR supersedes #2737.

beutlich:
> 
> 
> This PR supersedes #2737.

Rebased and resolved conflicts, now that #2737 is merged.

beutlich:
TestEddyCurrentBrake is a simple component test model and not a proper example model. It even is labeled "Test" which disqualifies it even more. Have a look at Modelica.Media where we also moved all these simple test-only examples to ModelicaTest.Media.

tobolar:
@christiankral To have an example is always a good idea. Even better if it's simple but realistic. Do you think you could prepare such a model utilizing EddyCurrentBrake -  instead of the original TestEddyCurrentBrake?

beutlich:
:+1: for your active reviews: That is why we have reviews for.

I reverted the removal, renamed and updated the two models in Modelica.Mechanics, but also want to keep them in the TestBraking test models.

Issue #2735, opened 2018-10-12
Modelica.Electrical.Analog.Basic.M_Transformer missing dotted line


kristinmajetta:
In the Icon of Modelica.Electrical.Analog.Basic.M_Transformer the dotted line between the second and the third inductor symbol is missing. It was there in MSL.3.2.2 and should be added again.
It is not clear why it is gone

PR #2746, opened 2018-10-17
Add number of inductors to icon display of M_Transformer

beutlich:
grafik


kristinmajetta:
As I can see it there are no dots bewtween the second and the third inductor. That is ok for the case N=3 as shown here.
But how would the icon look like when there are e.g. 4 inductors?
Would there still be three in the picture with some dots between the second and the third inductor to indicate that the number of inductors N=4 in the model does not match with the number of inductors in the icon?

beutlich:
This PR is independent of #2735 as it just adds the number to the icon display and does not touch the number of inductor lines or dots.

grafik

PR #2736, opened 2018-10-12
Issue2735 add dotted line

kristinmajetta:
added dotted line in icon in Modelica.Electrical.Analog.Basic.M_Transformer according to #2735

close #2735

PR #2748, opened 2018-10-17
Refs #2735: Improve dots in icon of M_Transformer

### LINKED ISSUE #2735: Modelica.Electrical.Analog.Basic.M_Transformer missing dotted line

christiankral:
This is an alternative proposal to #2736, applying to solid circles instead of a dotted line:

image

close #2735

### DIFF
--- a/Modelica/Electrical/Analog/Basic.mo
+++ b/Modelica/Electrical/Analog/Basic.mo
@@ -583,7 +583,18 @@ package Basic "Basic electrical components"
           Line(
             points={{30,40},{31,46},{38,54},{52,54},{59,46},{60,40}},
             color={0,0,255},
-            smooth=Smooth.Bezier)}),Documentation(info="<html>
+            smooth=Smooth.Bezier),
+          Ellipse(
+            extent={{-2,0},{2,-4}},
+            lineColor={0,0,255},
+            fillColor={0,0,255},
+            fillPattern=FillPattern.Solid),
+          Ellipse(
+            extent={{-2,-14},{2,-18}},
+            lineColor={0,0,255},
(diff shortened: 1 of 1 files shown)


tobolar:
To my opinion, both two or three circles should be fine to understand the meaning. Within the MSL documentation, there are often used just two dots to indicate a range (i.e. min..max, see e.g. `Modelica.Fluid.Valves.BaseClasses.ValveCharacteristics.baseFun`).
The dotted line, in contrast, changes the number of dots depending on component size. Moreover IMHO it is too thin.

kristinmajetta:
I agree that the dotted line might not be the best possibility since it changes the number of dots according to the icon size. But I suggest to draw a third circle in the icon. Is it possible for you to do that @christiankral ?

christiankral:
My opinion here: the three dots do not make it better... The smaller the screen resolution is the worse it gets...

image 
image


Shall I commit the three dots?

tobolar:
A possibility is to redesign the icon so that there is more space left for the three dots. Something like:
grafik

kristinmajetta:
The redesigned icons of @tobolar look fine to me

christiankral:
Same here: I would go for the icon of @tobolar with more space if three dots are the way to go.

christiankral:
OK, just committed version with three dots

image 
image

beutlich:
> 
> 
> I'll care for the merge conflict.

Done.

grafik

PR #2758, opened 2018-10-28
Fix icons to utilize graphics instead of text

### LINKED ISSUE #2757: Icons for Blocks.Logical (LessThreshold, LessEqualThreshold, GreaterThreshold and GreaterEqualThreshold) misleading when flipped
The icon of LessThreshold is a sign "<". When the component is used as is, it is meaningful.
If the component flipped horizontally, the same icon "<" is displayed, which indicates greater instead of less!

Also GreaterThreshold is affected by this issue. The issue can be solved using a drawing for the less or greater symbol instead of a character. 

Less, Greater, LessEqual and GreaterEqual are unaffected, since their icons are already drawings.

beutlich:
Ensure consistent looks of all 8 affected icons

close #2757

### DIFF
--- a/Modelica/Blocks/Logical.mo
+++ b/Modelica/Blocks/Logical.mo
@@ -214,9 +214,10 @@ the output is <strong>false</strong>.
   equation
     y = u > threshold;
     annotation (Icon(coordinateSystem(preserveAspectRatio=true, extent={{-100,-100},
-              {100,100}}), graphics={Text(
-            extent={{-90,-40},{60,40}},
-            textString=">")}), Documentation(info="<html>
+              {100,100}}), graphics={
+          Line(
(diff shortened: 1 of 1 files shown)


beutlich:
Note, that the diagonal lines for the ge or le comparisons already are in MSL v3.2.1. I am not sure why @MartinOtter introduced such diagonal lines by 2d0cf3579a34b1a5b6fa289f48e51c5f9c028681. Maybe it looks like roman numbers otherwise.

tobolar:
> Maybe it looks like roman numbers if rotated.

I'm aware of it. The icons with horizontal line look better then the original ones when oriented horizontaly. But not necessarilly better when rotated.
Fine to me if you discard this suggestion.

MartinOtter:
I have no strong opinion, both are fine with me.

PR #2779, opened 2018-11-20
Fix input array dimensions of mixtureViscosityChung

beutlich:
@finback-at FYI

### DIFF
--- a/Modelica/Media/IdealGases/Common/package.mo
+++ b/Modelica/Media/IdealGases/Common/package.mo
@@ -993,13 +993,13 @@ end gasMixtureViscosity;
   extends Modelica.Icons.Function;
 
     input Temperature T "Temperature";
-    input Temperature[nX] Tc "Critical temperatures";
-    input MolarVolume[nX] Vcrit "Critical volumes (m3/mol)";
-    input Real[nX] w "Acentric factors";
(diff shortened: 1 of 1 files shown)


finback-at:
Excuse me but I can’t under stand it as bug.

If we accept these changes, 
this function does not have the information of nX, number of substances of a medium,
and just assumed it from the length of mole fraction vector.

Consistency and independency of this function will be more strong, 
because number of element n and the sizes of protected all vectors in this function,
 and sizes of all input vectors will be same.

I think about the case that size of y and nX are different.
It is convenient to calculate viscosity of a part of a medium, for example,  
gas phase of multi phase medium, reactant part or product part of chemical reaction..

beutlich:
Likely that I am wrong. Please give me minimal example how to call mixtureViscosityChung for methane and where nX comes from. This was my example, which does not work w/o the patch applied:

```mo
model Test "Call mixtureViscosityChung on methane"
  Real eta=Modelica.Media.IdealGases.Common.MixtureGasNasa.mixtureViscosityChung(
      100 + 100*time,
      {190.56},
      {9.86E-05},
      {0.011},
      {0},
      {0.01604246},
      {1},
      {0});
end Test;
```

PR #2812, opened 2019-01-20
Add final to modifiers in examples HeatingNPN_NORGate and HeatingPNP_NORGate

casella:
These two examples define `tauVal = 0` and `capVal = 0` with `annotation(Evaluate = true)`. Then, these parameters are passed to capacitor and transistor parameter values. However, if they are not passed as `final`, a tool does not necessarily evaluate the actual capacitor and transistor parameter values, possibly ending up in a singular model. This actually happens to OpenModelica, see [report](https://libraries.openmodelica.org/branches/newInst/Modelica_trunk/files/Modelica_trunk_Modelica.Electrical.Analog.Examples.HeatingNPN_OrGate.sim)

Adding `final` to the capacitor and transistor parameter modifiers ensures that those values are immutably assigned to parameters evaluated to zero, so they can also be evaluated to zero, thus avoiding the singularity thanks to some further symbolic processing.

### DIFF
--- a/Modelica/Electrical/Analog/Examples/HeatingNPN_OrGate.mo
+++ b/Modelica/Electrical/Analog/Examples/HeatingNPN_OrGate.mo
@@ -65,17 +65,17 @@ model HeatingNPN_OrGate "Heating NPN Or Gate"
   annotation (Placement(transformation(extent={{-10,16},{10,36}})));
   Modelica.Electrical.Analog.Basic.Ground Gnd4
   annotation (Placement(transformation(extent={{30,-52},{50,-32}})));
-  Modelica.Electrical.Analog.Basic.Capacitor C1(C=CapVal)
+  Modelica.Electrical.Analog.Basic.Capacitor C1(final C=CapVal)
   annotation (Placement(transformation(
         origin={-70,38},
         extent={{-10,-10},{10,10}},
         rotation=270)));
-  Modelica.Electrical.Analog.Basic.Capacitor C2(C=CapVal)
+  Modelica.Electrical.Analog.Basic.Capacitor C2(final C=CapVal)
   annotation (Placement(transformation(
         origin={60,42},
         extent={{-10,-10},{10,10}},
         rotation=270)));
-  Modelica.Electrical.Analog.Basic.Capacitor C3(C=CapVal)
+  Modelica.Electrical.Analog.Basic.Capacitor C3(final C=CapVal)
   annotation (Placement(transformation(
         origin={-16,-40},
         extent={{-10,-10},{10,10}},
@@ -91,11 +91,11 @@ model HeatingNPN_OrGate "Heating NPN Or Gate"
     Br=1,
     Is=1e-14,
     Vak=0,
-    Tauf=tauVal,
-    Taur=tauVal,
-    Ccs=CapVal,
-    Cje=CapVal,
-    Cjc=CapVal,
+    final Tauf=tauVal,
+    final Taur=tauVal,
+    final Ccs=CapVal,
+    final Cje=CapVal,
+    final Cjc=CapVal,
     Phie=1,
     Me=0.5,
     Phic=1,
@@ -111,11 +111,11 @@ model HeatingNPN_OrGate "Heating NPN Or Gate"
     Br=1,
     Is=1e-14,
     Vak=0,
-    Tauf=tauVal,
-    Taur=tauVal,
-    Ccs=CapVal,
-    Cje=CapVal,
-    Cjc=CapVal,
+    final Tauf=tauVal,
+    final Taur=tauVal,
+    final Ccs=CapVal,
+    final Cje=CapVal,
+    final Cjc=CapVal,
     Phie=1,
     Me=0.5,
     Phic=1,
--- a/Modelica/Electrical/Analog/Examples/HeatingPNP_NORGate.mo
+++ b/Modelica/Electrical/Analog/Examples/HeatingPNP_NORGate.mo
@@ -61,17 +61,17 @@ model HeatingPNP_NORGate "Heating PNP NOR Gate"
   annotation (Placement(transformation(extent={{-10,16},{10,36}})));
   Modelica.Electrical.Analog.Basic.Ground Gnd4
   annotation (Placement(transformation(extent={{30,-52},{50,-32}})));
-  Modelica.Electrical.Analog.Basic.Capacitor C1(C=CapVal)
+  Modelica.Electrical.Analog.Basic.Capacitor C1(final C=CapVal)
   annotation (Placement(transformation(
         origin={-70,38},
         extent={{-10,-10},{10,10}},
         rotation=270)));
-  Modelica.Electrical.Analog.Basic.Capacitor C2(C=CapVal)
+  Modelica.Electrical.Analog.Basic.Capacitor C2(final C=CapVal)
   annotation (Placement(transformation(
         origin={60,42},
         extent={{-10,-10},{10,10}},
         rotation=270)));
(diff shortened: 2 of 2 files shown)


HansOlsson:
> These two examples define tauVal = 0 and capVal = 0 with annotation(Evaluate = true). Then, these parameters are passed to capacitor and transistor parameter values. However, if they are not passed as final, a tool does not necessarily evaluate the actual capacitor and transistor parameter values, possibly ending up in a singular model. 

As I understand it (but I might be wrong), the intention is not to edit the parameters that are equal to tauVal and capVal, so from that perspective the change makes sense.

However, in general I see a risk that we modify models to add tool-specific hints in obscure ways, and I don't see why the tool would need this hint. 
If tau=tauVal (without final) it seems complicated to override it with a value after translation, and obviously if tauVal has Evaluate=true and is treated as a literal then everything equal to it could also be treated the same way.

casella:
@HansOlsson, the point here is that the model has a potential structural singularity depending on the value of those two parameters, whether they are zero or not, and the obvious intention of the model developers is that the zero case must be handled. So, it is not enough that these parameters are evaluated at the top level, but they also need to be evaluated _inside the component_, to handle the structural singularity when they are zero. I think we agree on this point.

The question is then precisely how do we make sure that _any tool_ would do that, and that the structural well-posedness of the simulation model is guaranteed with any tool, without depending on potentially obscure heuristics.

Consider this minimalistic test case, that reproduces the issue.
```
package P
  model M
    parameter Real tau;
    Real x;
  equation
    tau*der(x) + x = sin(time);
  end M; 
  model S
    parameter Real tau = 0 annotation(Evaluate = true);
    M m(tau = tau);
  end S;
  model S2
    parameter Real tau = 0 annotation(Evaluate = true);
    M m(final tau = tau);
  end S2;
end P;
```
In the case of `S`, since the binding of `tau` is not final, I think that it is absolutely reasonable that a tool allows to change the value at runtime, which would unfortunately prevent the structural simplification when its value is zero. Of course one may introduce a heuristic rule that if the bound expression contains parameters with `Evaluate = true`, then the parameter should also be evaluated. IMO this is questionable, the details of this rule may be a bit obscure, and the validity of a model in the MSL should not depend on such a heuristic rule, which is not stated in the language specification.

To the contrary, in the case of `S2` the `final` modifier makes it clear that the value of `S2.m.tau` is identically equal to `S2.tau`, so if the latter is evaluated, there is no reason why the former shouldn't be evaluated as well. Even though this is not explicitly stated in the language specification, this seems to me an obvious consequence of the language semantics, that a modeller could expect any tool to follow. I don't see anything obscure in that, and in fact the `final `modifier is not at all a hint, but is rather a precise statement of a structural property of the model made by the modeller.

@beutlich, I don't think this is a domain-specific issue for electrical models, the very same issue could arise with thermal or mechanical models, each time you have a generalized mass or inertia parameter that can have a zero value, that must in turn be handled by structural analysis.

I would like to hear the opinion of other tool developers other than Dymola and OpenModelica.

HansOlsson:
> In the case of S, since the binding of tau is not final, I think that it is absolutely reasonable that a tool allows to change the value at runtime, which would unfortunately prevent the structural simplification when its value is zero. 

Except, that I don't see how a tool can easily do that. The default for `m.tau` is `tau`; so it seems complicated to both show that default and allow setting `m.tau` to another value.

I could understand if a tool evaluates `tau` (to 0) and presents 0 as default for `m.tau` - but that gives the users the wrong impression of the model.

Obviously if users aren't supposed to change `m.tau` directly it should be `final `- not because it will helps some tools, but because that is the intention of the modeler. 

This is also one of those cases where the tool (if we don't provide the hints) sees that something will fail and anyway proceeds. We should aim to be smarter to attract new users.

henrikt-ma:
In SystemModeler, we do at the same time show the expression used by default, and allow the user to override with a numeric value after translation.  The only problem with this is that it sometimes makes it a bit too easy to override binding equations that was intended to be final, or at least only modified in the form of a Modelica modifier before translation.  For this reason, we hide these parameters behind a filter flag that is off by default.  However, if "final" was used consistently wherever it doesn't make sense to override a binding equation, I think that the need to filter these out would be reduced significantly.  Now, I suspect that "final" is not use as much as it should due to tools almost treating parameters with binding equation as if they were final anyway.


Henrik

casella:
We have solved this issue in OpenModelica by introducing a rule that parameters that have a binding to structural parameters (such as those with `Evaluate = true`) are also marked as structural, and thus get evaluated by the front-end. This actually solves the issue reported here without the need of introducing any final modifier.

I guess users actually expect the `Evaluate = true` property to be passed on when propagating parameters, and I understand that's what Dymola does. This is fine for me, but then I would suggest to add this rule to the specification, so that we get consistent behaviour across tools. Without this feature, models will fail because of lacking structural simplification in the case of zero parameter values.

@HansOlsson, should I open a ticket on this topic for Modelica 3.5?

henrikt-ma:
In my opinion, propagating `Evaluate = true` in this direction is just a mess.

The proper way of handling this is to declare components as `constant` instead of `parameter` wherever they are supposed to be evaluated during translation.  This will then propagate in the other direction, making it clear all the way up that some things will not be possible to change after translation.

HansOlsson:
> I guess users actually expect the Evaluate = true property to be passed on when propagating parameters, and I understand that's what Dymola does. 

The effect is similar, but we don't do exactly that. It's just that we cannot change a parameter bound to a (parameter-)expression, and thus the evaluation of one parameter replaces it with its value - triggering a similar result.

We deliberately "propagate" _in the other direction_, so that `parameter Real p=p2 annotation(Evaluate=true)` causes p2 to be evaluated (if possible). (Obviously "=p2" is normally a modifier.)

**Unless parameter-alias is enabled**, because then there is only one parameter and Evaluate=true is propagated to the common one.

> @HansOlsson, should I open a ticket on this topic for Modelica 3.5?

Sure.

henrikt-ma:
Since `Evaluate = true` has such drastic impact on the symbolic processing, it doesn't make sense to just take it as a hint that may or may not be obeyed.  There is simply too high risk that the library author used a tool that didn't ignore the hint, and that the translation time evaluation of the parameter has become an assumption that must hold in order for the model to work as expected.  Hence, a tool is forced to perform translation time evaluation, and we must make it very what this actually comes down to.

Besides, this also illustrates why `Evaluate = true` shouldn't be an annotation in the first place.  If it was a built-in property, it would also be possible to set it from the outside using a modifier.  I think that this could be the solution for at least some of the scenarios where one wants a parameter in the general case, but sometimes get the simplified equations for a particular value of the parameter.

beutlich:
What is the current state of the PR? Can we close it as wont-fix?

henrikt-ma:
Given that it doesn't seem clear how the `Evaluate = true` is expected to be propagated, why not approve this PR instead of closing as wont-fix?

HansOlsson:
> Given that it doesn't seem clear how the Evaluate = true is expected to be propagated, why not approve this PR instead of closing as wont-fix?

_As I understand it_ the propagation were supposed to be final in this model (i.e. you weren't supposed to override them later - but only change "capval"). 

So, assuming that it is right, we should approve the PR; regardless of the Evaluate-part.

However, we should probably change the title to say to just something like "Added final to modifiers". (We could say that it is to avoid modification of them - but that is what "final" means. Note that I wrote "add final to modifiers" not "final modifiers" since adding a "final modifier" could also mean adding a new modifier that is final.)

beutlich:
Right. Waiting for @kristinmajetta's, @christophclauss's, @christiankral's or @dietmarw's approval.

PR #2816, opened 2019-02-05
Add safeguard for log(T)

### LINKED ISSUE #2815: moist air at very low temperatures
The nonlinear solver sometimes uses very low or even negative temperatures in function calls to `h_pTX`, resulting in errors, because of the `log(T)` in the `h_Tlow` function. 
I suggest to catch negative temperatures and return h=-inf for T<=0, will send a PR.

The following model was used to investigate and it revealed another problem:
```model MoistAirTLowTest
  replaceable package Medium = Modelica.Media.Air.MoistAir;
  Modelica.Blocks.Sources.Ramp Tramp(
    height=-320,
    duration=0.8,
    offset=323.15,
    startTime=0.1)
    annotation (Placement(transformation(extent={{-80,38},{-60,60}})));

  Medium.AbsolutePressure pamb= 101325;
  Medium.Temperature T;
  Medium.SpecificEnthalpy h;

equation 
  T = Tramp.y;
  h = Medium.h_pTX(pamb,T,Medium.reference_X);
  annotation (uses(Modelica(version="3.2.2")));
end MoistAirTLowTest;
```
After simulation, plot h with T as independent variable. The bump in the curve might push the solver even further in the wrong direction, I believe. 

moistairtlowtest

thorade:
close #2815

### DIFF
--- a/Modelica/Media/IdealGases/Common/Functions.mo
+++ b/Modelica/Media/IdealGases/Common/Functions.mo
@@ -119,7 +119,7 @@ package Functions
 
   algorithm
     h := data.R*((-data.alow[1] + T*(data.
-      blow[1] + data.alow[2]*(if T>0 then Math.log(T) else -Modelica.Constants.inf) + T*(1.*data.alow[3] + T*(0.5*data.
+      blow[1] + data.alow[2]*(if T>0 then Math.log(T) else Modelica.Constants.inf) + T*(1.*data.alow[3] + T*(0.5*data.
       alow[4] + T*(1/3*data.alow[5] + T*(0.25*data.alow[6] + 0.2*data.alow[7]*T))))))
       /T) + (if
       exclEnthForm then -data.Hf else 0.0) + (if (refChoice
--- a/Modelica/Media/IdealGases/Common/Functions.mo
+++ b/Modelica/Media/IdealGases/Common/Functions.mo
@@ -119,7 +119,7 @@ package Functions
 
   algorithm
     h := data.R*((-data.alow[1] + T*(data.
-      blow[1] + data.alow[2]*Math.log(T) + T*(1.*data.alow[3] + T*(0.5*data.
+      blow[1] + data.alow[2]*(if T>0 then Math.log(T) else -Modelica.Constants.inf) + T*(1.*data.alow[3] + T*(0.5*data.
       alow[4] + T*(1/3*data.alow[5] + T*(0.25*data.alow[6] + 0.2*data.alow[7]*T))))))
       /T) + (if
       exclEnthForm then -data.Hf else 0.0) + (if (refChoice


Modelica/Media/IdealGases/Common/Functions.mo
```diff
@@ -119,7 +119,7 @@ package Functions
 
   algorithm
     h := data.R*((-data.alow[1] + T*(data.
-      blow[1] + data.alow[2]*Math.log(T) + T*(1.*data.alow[3] + T*(0.5*data.
+      blow[1] + data.alow[2]*(if T>0 then Math.log(T) else -Modelica.Constants.inf) + T*(1.*data.alow[3] + T*(0.5*data.
```
beutlich: Should be noEvent(T>0), right?

HansOlsson: It is a function, right? So no need for noEvent.

thorade: It is inside function h_Tlow

thorade: Pushing another commit, because I did not save. The term is multiplied by T, so the opposite sign is correct.

thorade:
The case T=0 also needs to be covered, but I think it is
https://math.stackexchange.com/questions/470952/limit-of-x-log-x-as-x-tends-to-0

beutlich:
Modelica.SIunits.ThermodynamicTemperature has min = 0 set. I wonder why it is all needed then and if it is a tool issue.

thorade:
Similar safeguards are used in other places as well, e.g. here:
https://github.com/modelica/ModelicaStandardLibrary/blob/master/Modelica/Media/IdealGases/Common/package.mo#L244
All input to that sqrt should be positive.

thorade:
If connectors use p and h, then h_T and the inverse T_h will be called very often (with sometimes extreme inputs). 
A similar safeguard could be introduced for other functions, but only if it is a problem for anybody.

beutlich:
@thorade I cannot confirm that the patch fixes the MoistAirTLowTest model given by you in #2815

thorade:
To see the difference, the ramp height has to be larger than the offset, so you actually reach negative values. Then, comment the call to sublimationPressure.

h_Tlow

beutlich:
I am confused. I thought the issue is about the monotonicity break around 7...8K in MoistAirTLowTest. Is this really the case as I can still reproduce it?

thorade:
This is only addressing negative input to log(T). 
I am not fully conviced this fix is necessary, also there might be more elgant ways to fix it

beutlich:
> 
> 
> This is only addressing negative input to log(T).
> I am not fully conviced this fix is necessary, also there might be more elgant ways to fix it

Shall we close it as wontfix then?

thorade:
> Shall we close it as wontfix then?

OK for me, I can re-open if I see the need.

PR #2824, opened 2019-02-06
Increment package version number to 4.0.0-dev

beutlich:


beutlich:
> 
> 
> ModelicaObsolete3 does that still make sense?

Probably not. Time to get rid of it, but different PR.

PR #2826, opened 2019-02-08
Remove outdated documentation on bivariate extrapolation

beutlich:


dietmarw:
14 days have passed, so approving instead of LO

PR #2823, opened 2019-02-06
Fix icon frame color for Complex blocks

beutlich:
close #2477

PR #2827, opened 2019-02-10
Improve block icons and diagrams of Math.{Add,Add3,Division} and ComplexMath.{Add,Add3,Division}

beutlich:


beutlich:
@tobolar Thanks for the approval and the commit.

The main motivation was #2823 when I noticed that the ComplexMath.{Add,Add3} icon and diagram graphics is messed up.

PR #2828, opened 2019-02-12
Utilize enumeration for clean-up alternatives in ModelicaStandardTables.c

beutlich:


beutlich:
Yes, I know, that you gain no type-safety here in C. Just wanted to give things a name and avoid those magic numbers in combination with a comment.

PR #2837, opened 2019-02-23
ModelicaReference: add AutoCompletion subpackage

atrosinenko:
To start with, add an experiment annotation as an example, so one can debug this feature and extend it separately.

This PR is related to OpenModelica/OMEdit#219 and is based on already existing text in this repo, as well as a Modelica Standard. Unfortunately, this will probably be partial copy-paste of `ModelicaReference.Annotations` instead of plain its usage, since descriptions there are more standard-related and human-readable than machine-readable formal record-like form.

Another idea, is to gather there documentation on the keywords, too, to not hardcode it in the IDE code.

### DIFF
--- a/ModelicaReference/package.mo
+++ b/ModelicaReference/package.mo
@@ -871,9 +871,12 @@ Evaluate is for example used for axis of rotation parameters in the Modelica.Mec
 </html>"));
   end Evaluate;
 
-  class experiment "experiment"
+  record experiment "experiment"
     extends ModelicaReference.Icons.Information;
-
+      Real StartTime(unit = "s") "Default start time of simulation";
+      Real StopTime(unit = "s") "Default stop time of simulation";
+      Real Interval(unit = "s", min = 0) "Resolution for the result grid";
+      Real Tolerance(unit = "1", min = 0) "Default relative integration tolerance";
     annotation (Documentation(info="<html>
 <p>
 Define default experiment parameters


atrosinenko:
default

ModelicaReference/package.mo
```diff
@@ -7013,6 +7013,22 @@ package Icons "Library of icons"
 end Icons;
 
 
+package AutoCompletion "Auto completion information for Modelica IDEs"
+  package Annotations "Auto completion information on annotations"
+      record experiment "Define default experiment parameters"
+        Real StartTime(unit = "s") "Default start time of simulation";
```
dietmarw: Instead of hardcoding the unit here it should be using the proper type Modelica.SIunits.Time instead of Real. Same for the rest.

ModelicaReference/package.mo
```diff
@@ -7013,6 +7013,22 @@ package Icons "Library of icons"
...
+        Real StopTime(unit = "s") "Default stop time of simulation";
+        Real Interval(unit = "s", min = 0) "Resolution for the result grid";
+        Real Tolerance(min = 0) "Default relative integration tolerance";
+      end experiment;
+    annotation(
+        Documentation(info = "<html><head></head><body>In this package annotations are gathered in their record-like form together with meta information such as descriptions, units, min, max, etc.</body></html>"));
```
dietmarw: Remove superfluous <head> and <body> tags.

ModelicaReference/package.mo
sjoelund: It's a little bit weird actually. Because ModelicaReference does not depend on the standard library. So I think it's correct to define it as Real with a unit set manually...

dietmarw: Good point. Didn't think about that.

atrosinenko:
Fixed tags, restored `unit`s.

HansOlsson:
I'm not sure if this fits into ModelicaReference at the moment (there is some old MCP(s) about standardizing annotations as Modelica code - which has been inactive for quite some time).

The leads to the second part: if it weren't in ModelicaReference - but somewhere else - the unit-dependency would also be a non-issue.

dietmarw:
I think the ModelicaReference is a good place for all existing, standardised, annotations. But I would not want them duplicated. So if the existing documentation could be combined with the type then I think it would be a nice enhancement. 

I can not see a problem in changing the existing
```mo
class experiment "experiment"
  extends ModelicaReference.Icons.Information;
  annotation ();
end experiment;
```
to 
```mo
record experiment "experiment"
  extends ModelicaReference.Icons.Information;
  Real StartTime(unit = "s") "Default start time of simulation";
  annotation ();
end experiment;
```

atrosinenko:
Agree, duplicating is bad, I thought it is inherently user-readable docs, but I have not noticed it can be naturally combined.

```
record experiment "experiment"
                   ^ -- should this always be equal to the record name or can be changed?
```

atrosinenko:
Hmm, I realized why description == name -- it is viewed better in doc browser then. Is it necessary to have some short summaries at all (in some "strict" machine readable form, that is, not "the first paragraph if it is HTML, otherwise the first string, or maybe something else...") or always show the full HTML description (it would probably be useful for the user).

dietmarw:
Could you rephrase that? I'm not sure I understand what you mean.

HansOlsson:
> I think the ModelicaReference is a good place for all existing, standardised, annotations. But I would not want them duplicated.

I agree that duplication within ModelicaReference would be bad. It is possible that ModelicaReference is a good place for the annotations in this form, but I'm not certain. My issue is that we will try to do two things with one class and that it may create conflicting requirements; not only for this class - but also e.g., for marking which classes are relevant in this way. Maybe that is not a problem in practice - I don't know.

atrosinenko:
@dietmarw I tried first to ask whether I need `record recordName "recordName"` instead of traditional `record recordName "Record description for user"`, but then I opened the second variant in a documentation browser of Open Modelica and seen 1) duplication of the summary (copied from the first paragraph) 2) lack of easily noticeable annotation name -- so I need to use this `record name "name"` form.

default

atrosinenko:
@HansOlsson my original thoughts were:
1. I had to implement annotation autocompletion
2. I implemented it in the editor, so I need some place to fetch the annotation descriptions from
    * **hardcode it in C++:** reimplementing description machinery for hierarchical structures, types of fields, their units, etc. In addition, it would be maintained by the **editor** developers while the **compiler** developers have more info
    * **implement it as a Modelica package:** this would reuse parsers and storage format, allows more easy editing, even HTML snippet generator from regular autocompletion. Even more, AFAIK it is the record form that considered kind of reference form for annotation description
3. Then @casella suggested to move it to standard library for general use

Some random thoughts:
* Other autocompletion can be fetched likewise: keywords, etc. (so the original structure: ModelicaReference / AutoCompletion / **Annotations** / _experiment_)
* It would be great to have vendor annotations here as well -- it may be handy for understanding / porting purposes

atrosinenko:
Another thought: maybe user-accessible Modelica package would encourage more people to contribute to this machine-readable documentation (as well as will be easily browsable source of "textbook-form" annotation descriptions) but another separate root package may clutter package tree in editor.

dietmarw:
In order to move forward. Any problem to add the annotations as [I suggested](https://github.com/modelica/ModelicaStandardLibrary/pull/2837#issuecomment-466987145)?
As for vendor annotations,  those should be kept inside the tool. We (MAP-LIB) can not take care of them. So only standardised annotation will be added to ModelicaReference.

atrosinenko:
Refactored as suggested. I just waited a bit to collect different opinions. For now, refactored only single annotation, so I can support this and see how it would behave.

thorade:
There is one technology that I would like to promote whenever I hear auto-completion or syntax-highlight and it is the Language Server Protocol (LSP)!

LSP was inventend and open-sourced by Microsoft for Visual Studio Code for offering advanced IDE features (as the mentioned ones, but also go to definition, type checking, linting, code folding, find usage, prettyprint etc). LSP is a protocol for communication between an editor/IDE and a language server. On the editor side it is supported by Visual Studio Code, emacs, Sublime, Atom, Eclipse and many others. On the language side, there are servers for most of the popular languages (C++, Python, Go, Rust...).

But, there is no language server yet for Modelica! Most of the features required like parsing a library are implemented in each tool, Dymola, OpenModelica, JModelica..., so in my naive view, implementing a Modelica language server would be a somewhat low hanging fruit!?

Read more about LSP here:
https://microsoft.github.io/language-server-protocol/
https://langserver.org/

See some nice animated gifs of LSP in action here:
https://github.com/palantir/python-language-server

atrosinenko:
> But, there is no language server yet for Modelica!

I even tried to implement it in Scala with OMC as a backend (at the first glance, it seemed the simplest solution to me), but stumbled upon LSP support in different editors (or probably my implementation was broken, but diagnostic in tools was quite poor -- most probably I was doing something wrong...). If this does matter, it was about half-a-year ago.

dietmarw:
Based on the feedback until now it really turns into a tool specific problem. And unless a complete solution based on Modelica syntax is presented and can be evaluated as replacing (parts) of ModelicaReference I can only agree with @HansOlsson that ModelicaReference is not the right place (yet). 

@atrosinenko I suggest you continue with `AutoCompletion` library inside the OpenModelica/OMEdit repository. Then we can still see of it makes sense to move this into here. As mentioned, there have been previous attempts made to "classify" the Annotations but this seems to be *non-trivial*.

casella:
> Based on the feedback until now it really turns into a tool specific problem. And unless a complete solution based on Modelica syntax is presented and can be evaluated as replacing (parts) of ModelicaReference I can only agree with @HansOlsson that ModelicaReference is not the right place (yet).
> 
> @atrosinenko I suggest you continue with `AutoCompletion` library inside the OpenModelica/OMEdit repository. Then we can still see of it makes sense to move this into here.

Sounds good to me.

PR #2871, opened 2019-03-29
Move Polynomials from Media to Math

### LINKED ISSUE #978: Polynomial functions
**Reported by kdavies on 31 Jan 2013 01:53 UTC**
This ticket is to add a Polynomials package to Modelica.Math.  Unfortunately, after writing the package for another library, I realized that Modelica.Media.Incompressible.TableBased.Polynomials_Temp seems to have the same functions and probably more.  Maybe the two packages can be combined.

One advantage of the new Polynomial package is that it generates polynomial expressions in a nested form (((... a2)\*x + a1)\*x) which should be more computationally efficient.  Also, the tests (added to ModelicaTest.Math) may be more thorough; they are based on the der() operator as recommended in the Dymola 7.4 User's Manual Volume 2.

Regardless, it may be good to move Polynomials_Temp to the Math package so that users can find it more easily.

Attachment: [0001-polynomials.patch](https://github.com/modelica/ModelicaStandardLibrary/files/3020337/0001-polynomials.zip)

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/978

beutlich:
When Polynomials were added to Modelica.Media.Incompressible.TableBased they were considered temporary. This is a cleanup to move the existing Polynomials package to Modelica.Math (as proposed by #978).

In #978, there has been some intent to introduce the Modelica_LinearSystems2.Math.Polynomial package instead, but given the current limited resources and requirements, it is enough to do this cleanup.

Close #978.

### DIFF
--- a/Modelica/Math/Polynomials.mo
+++ b/Modelica/Math/Polynomials.mo
@@ -1,0 +1,261 @@
+within Modelica.Math;
+    package Polynomials
+      "Library of functions operating on polynomials (including polynomial fitting)"
+      extends Modelica.Icons.FunctionsPackage;
+
+      function evaluate "Evaluate polynomial at a given abscissa value"
+        extends Modelica.Icons.Function;
+        input Real p[:]
+          "Polynomial coefficients (p[1] is coefficient of highest power)";
+        input Real u "Abscissa value";
+        output Real y "Value of polynomial at u";
+      algorithm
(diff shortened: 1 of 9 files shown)


beutlich:
Merge conflict in ConvertModelica_from_3.2.3_to_4.0.0.mos is resolved.

PR #2879, opened 2019-04-04
Avoid Boolean comparison with Real in DynamicSelect

beutlich:
Comparisons of type Boolean with Real is illegal, but so far it was not detected if applied in DynamicSelect.


beutlich:
@HansOlsson You might want to fix this issue also in Modelica_StateGraph2.

PR #2882, opened 2019-04-06
Rename HeatingNPN_OrGate to HeatingNPN_NORGate

### LINKED ISSUE #2521: Wrong name of example Modelica.Electrical.Analog.Examples.HeatingNPN_OrGate
The name of the example HeatingNPN-OrGate is wrong since the circuit in that example is a NOR circuit.
Unfortunatelly changing the name would break backwards compatibility. Therefore we need to wait for the next major release to change the name.

beutlich:
Close #2521.

### DIFF
--- a/Modelica/Electrical/Analog/Examples/HeatingNPN_NORGate.mo
+++ b/Modelica/Electrical/Analog/Examples/HeatingNPN_NORGate.mo
@@ -1,5 +1,5 @@
 within Modelica.Electrical.Analog.Examples;
-model HeatingNPN_OrGate "Heating NPN Or Gate"
+model HeatingNPN_NORGate "Heating NPN NOR Gate"
   extends Modelica.Icons.Example;
   parameter SI.Capacitance CapVal=0 "Value for capacitances" annotation(Evaluate=true);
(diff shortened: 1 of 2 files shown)


christiankral:
I am actually wondering why we need both the models `Modelica.Electrical.Analog.Semiconductors.HeatingNPN` and `Modelica.Electrical.Analog.Semiconductors.NPN`, as they both have a heat port. @christophclauss I suppose the underlying physical model is the same. 

Having components with `Heating` in their name are usually a leftover from the time when components were designed without heat port. Afterwards, the `Heating` components were add to the MSL. Later, step by step, the heat ports were added in a backwards compatible way. Therefore, we may consider removing all these `Heating` components from the MSL.

dietmarw:
I agree. All those Heating* models that have been now included in the * themselves should be removed (and usages converted).

beutlich:
> Therefore, we may consider removing all these Heating components from the MSL.

Can you please create a separate issue for that. This one is just on the example renaming.

christiankral:
#2899 is intended to remove the name `Heating` from obsolete models and examples

beutlich:
> It looks like the conversion script entry is missing

Since it is not required for example models.

PR #2886, opened 2019-04-08
Fix formula of LinearDependency

### LINKED ISSUE #1724: Modelica.Blocks.Math.LinearDependency has strange formulation
**Reported by Joe Riel on 1 Jun 2015 14:58 UTC**
The relatively new block LinearDependency seems strange. It's defining equation is `y = y0*(1 + k1*u1 + k2*u2);`. y0 is a parameter; its description says it is an initial value, which makes no sense.  It is a scale factor (with a useless default of 0).  However, if its value is nonzero, then one cannot produce an output with a non-zero offset.  I'm thinking the block would be slightly more useful if the defining equation were `y = y0 + k1*u1 + k2*u2`.

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/1724

beutlich:
Close #1724.

### DIFF
--- a/Modelica/Blocks/Math.mo
+++ b/Modelica/Blocks/Math.mo
@@ -3613,7 +3613,7 @@ provide both values as output.
     parameter Real k1=0 "u1 dependency";
     parameter Real k2=0 "u2 dependency";
   equation
-    y = y0*(1 + k1*u1 + k2*u2);
+    y = y0 + k1*u1 + k2*u2;
     annotation (Icon(coordinateSystem(preserveAspectRatio=false, extent={{-100,
               -100},{100,100}}), graphics={
           Line(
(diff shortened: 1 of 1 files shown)


beutlich:
I consider this as a bugfix, that will be well documented in the release notes. LinearDependency is not used in the MSL and probably also not elsewhere due to its non-intuitive behaviour.

MartinOtter:
> I consider this as a bugfix, that will be well documented in the release notes. LinearDependency is not used in the MSL and probably also not elsewhere due to its non-intuitive behaviour.

Note, this is certainly no bug fix, because the model works exactly as stated in the documentation. 

As library officer of the Modelica.Blocks library, I forbid you to make this change.

If you want to improve the situation, make it in a way that the user can cope with it. For example:

- Move this model in the ObsoleteModelica libray and provide a conversion script. If your claim is correct, that no one is using this model, no one will notice it. 
- If necessary, provide a model with a different name and your desired behavior.

beutlich:

beutlich:
Superseded by #3095 where I added conversion to keep the backward compatibility.

PR #2890, opened 2019-04-09
Clarify model description of temperature conversion models

beutlich:
They are models (but could be blocks).


beutlich:
@christiankral OK for you?

beutlich:
Reminder @christiankral

PR #2896, opened 2019-04-10
Improve initialization dialog of Modelica.Mechanics.Translational.Components.Mass

beutlich:
Have identical initialization dialog as Modelica.Mechanics.Rotational.Components.Inertia.

I needed to break the inheritance from PartialRigid, since the showStartAttribute of s should only be avilable for Mass and not for all other components extending from PartialRigid.

See #1855. @max-privato @ceraolo


MartinOtter:
Same reason as for Rotational. Only make changes, if there is an improved concept for initialization

beutlich:

beutlich:
Whereas Dymola always displays the start parameters for ODE states and derivatives in the Initialization group

grafik

SimulationX does not (since both the showStartAttribute and the group name are not specified in the dialog annotation).

grafik

I really wonder why unspecified dialog behaviour is encouraged and why Mass must not be consistent with Inertia. 😕

PR #2897, opened 2019-04-10
Remove showStartAttribute annotation from Inertia.a

beutlich:
This is the minimal requested change by of #1855.

@max-privato


MartinOtter:
Additional note: In case there is a better concept for initialization in MSL and the Modelica tools, the initialization of the Mechanics libraries can be adapted, but not beforehand.
I do no longer want to spend my spare time on this issue. So respect my decision (or come up with a better, general concept for initialization)

beutlich:

PR #2900, opened 2019-04-15
Remove prefix "wheel" in parameters' name 

### LINKED ISSUE #2653: Delete superfluous prefix "wheel" from parameters' name
Remove prefix "wheel" from parameters' name of `Mechanics.MultiBody.Parts.RollingWheel` since actually all parameters concern the wheel. In particular:
- wheelRadius -> radius
- wheelMass -> mass
- wheel_I_axis -> I_axis
- wheel_I_long -> I_long
- wheelColor -> color

tobolar:
closes #2653

### DIFF
--- a/Modelica/Mechanics/MultiBody/Examples/Elementary/RollingWheel.mo
+++ b/Modelica/Mechanics/MultiBody/Examples/Elementary/RollingWheel.mo
@@ -4,10 +4,10 @@ model RollingWheel
    extends Modelica.Icons.Example;
 
   Modelica.Mechanics.MultiBody.Parts.RollingWheel wheel1(
-    wheelRadius=0.3,
-    wheelMass=2,
-    wheel_I_axis=0.06,
-    wheel_I_long=0.12,
+    radius=0.3,
+    m=2,
+    I_axis=0.06,
+    I_long=0.12,
     hollowFraction=0.6,
     x(start=0.2),
     y(start=0.2),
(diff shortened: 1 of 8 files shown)


tobolar:
> The required conversion test models in ModelicaTestConversion4.mo are still missing. Do you want to add them?

Done in 5bd35a9435518b562157af444a502208d31494a6
Thanks.

tobolar:
> Why are the parameter names of RollingWheel and RollingWheelSet are changed such inconsistently?

Well this is to have robust parameter names once the RollingWheelSet would be extended in the future - which IMO is not impossible. Currently, there are just two wheels included but there could be even more parts, e.g. a connecting shaft. Then, it will be necessary to clearly distinguish between the concerning parameters, i.e. m_wheel vs. m_shaft, etc.
So I decided to consider this option for the current parameters already.

beutlich:
Hm, I am still confused. This PR was on removing the "wheel" as prefix. This was correctly performed on Joints.RollingWheel, Joints.RollingWheelSet and Parts.RollingWheel. However, the name changes on Parts.RollingWheelSet seem arbitrary and inconsistent to me. If it only is for some future extension (which is not scheduled right now) we could either leave the names as is are change consistently already now.

tobolar:
IMO this is not inconsistent. Calling a variable something like _wheelMass_ is simply a mess and I follow the recommendation in `Modelica.UsersGuide.Conventions.ModelicaCode.Naming#Variable names`, thus preferring _m_ over _mass_. The specification that the mass concerns a wheel is then indicated by a subscript __wheel_. But the variables' names are consistent in both cases of `RollingWheel`and `RollingWheelSet`.

To me the only question is whether to call variable for wheel radius in RollingWheel `radius`, `r` or `R`. I choosed `radius` but to be really consistent, `R` would be more suitable.

PR #2914, opened 2019-05-04
Rename of QuasiStationary to QuasiStatic

### LINKED ISSUE #1189: Rename of Electrical.QuasiStationary to Electrical.QuasiStatic
**Reported by dietmarw on 29 Jun 2013 18:20 UTC**
The library `Electrical.QuasiStationary` should better be called `Electrical.QuasiStatic` 
since the term is a "false friend" from the German language where one could say "quasistatonär" and "quasistatisch" for the same phenomenon. Both translate to "quasistatic" in the English language.  
See also:
 * http://en.wiktionary.org/wiki/quasistatic
 * http://en.wikipedia.org/wiki/Quasistatic_approximation

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/1189

beutlich:
Fix #1189.


beutlich:
> 
> 
> Quite some changes... Looks good.

Indeed. I did it step by step and added conversion test models for each class and fixed all remaining spelling issues last.

beutlich:
> 
> 
> There are still occurrences of "stationary" present that should be fixed.

Should be fixed by 7c2feb92d92f425ac96730774f2e39d09ca57484.

PR #2915, opened 2019-05-05
Fix compilation of ModelicaStandardTables.c if both TABLE_SHARE and NO_FILE_SYSTEM are defined

beutlich:
TABLE_SHARE only is relevant if NO_FILE_SYSTEM is not defined.


beutlich:
> 
> 
> Seems OK, though I guess we could use some CI job to build the C libraries with different CPPFLAGS set to see it compiles fine and link to some dummy executable or unit test for the routines...

Here it is: https://travis-ci.org/tbeu/ModelicaStandardTables/jobs/528925986

PR #2918, opened 2019-05-06
Fix compilation if both TABLE_SHARE and NO_FILE_SYSTEM are defined

beutlich:
TABLE_SHARE only is relevant if NO_FILE_SYSTEM is not defined.

Backport of #2915 for maint/3.2.3.


beutlich:
> Looks ok, and good to have on maintenance.

Yes, anything fixed for MSL v4.0.0 that is labled as `bug` will be (manually) backported.

beutlich:
Reminder @sjoelund

PR #2919, opened 2019-05-06
Fix algorithm of IdealGases.Common.Functions.s0_Tlow_der

### LINKED ISSUE #1909: Functions missing derivative functions, that are already present
**Reported by veje&#x2e;johannes on 16 Feb 2016 10:04 UTC**
In Media.IdealGases.Common.Functions there are three functions where the corresponding derivative function is available but not implemented. 

These three functions are; h_T, h_Tlow, and s0_Tlow with the corresponding functions: h_T_der, h_Tlow_der, and s0_Tlow_der.

```
function h_T 
  ...
  annotation (Inline=false,smoothOrder=2);
end h_T;
```

For a fourth function, cp_Tlow, the derivative function is implemented.

```
function cp_Tlow 
  ...
  annotation (Inline=false, derivative(zeroDerivative=data) = cp_Tlow_der);
end cp_Tlow;
```

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/1909

### LINKED ISSUE #2870: Algorithm of Modelica.Media.IdealGases.Common.Functions.s0_Tlow_der
Current algorithm of s0_Tlow_der is 
```
  s := data.R*(data.blow[2] - 0.5*data.alow[
    1]/(T*T) - data.alow[2]/T + data.alow[3]*Math.log(T) + T*(
    data.alow[4] + T*(0.5*data.alow[5] + T*(1/3*data.alow[6] + 0.25*data.alow[
    7]*T))));
```
It is just as same as s0_Tlow.

From the relation 
   (ds/dT)_p = Cp/T,
I think it should be changed to 
```
s := T_der * Modelica.Media.IdealGases.Common.Functions.cp_Tlow(data,T)/T;
```

beutlich:
@finback-at @thorade FYI

Fix #2870, fix #1909.

### DIFF
--- a/Modelica/Media/IdealGases/Common/Functions.mo
+++ b/Modelica/Media/IdealGases/Common/Functions.mo
@@ -183,13 +183,10 @@ package Functions
     extends Modelica.Icons.Function;
     input IdealGases.Common.DataRecord data "Ideal gas data";
     input SI.Temperature T "Temperature";
-    input Real T_der "Temperature derivative";
-    output SI.SpecificEntropy s "Specific entropy at temperature T";
+    input Real T_der(unit="K/s") "Temperature derivative";
+    output Real s_der(unit="J/(kg.K.s)") "Derivative of specific entropy at temperature T";
   algorithm
-    s := data.R*(data.blow[2] - 0.5*data.alow[
-      1]/(T*T) - data.alow[2]/T + data.alow[3]*Math.log(T) + T*(
-      data.alow[4] + T*(0.5*data.alow[5] + T*(1/3*data.alow[6] + 0.25*data.alow[
-      7]*T))));
+    s_der := data.R*T_der*(data.alow[1]/(T*T*T) + data.alow[2]/(T*T) + data.alow[3]/T + data.alow[4] + T*(data.alow[5] + T*(data.alow[6] + T*data.alow[7])));
     annotation (Inline=true);
   end s0_Tlow_der;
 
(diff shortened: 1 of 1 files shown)


finback-at:
Sorry, I dropped T_der in my comment. The code value is same as,
```
T_der * cp_T_low(data,T)/T,
````
and It's OK!

tobolar:
> ...I dropped T_der in my comment...

In that case you can simply edit your comment again: 
ClipboardX

thorade:
My understanding of #1909 was that the smoothOrder annotation is preferred!?
So the better solution would be to use smoothOrder here, and also for cp_T.
The derivative functions still cannot be deleted, because they are called explicitly from MoistAir.

beutlich:
Well, stating the derivative was easier for me (as I do not know which smooth order to set).

casella:
> My understanding of #1909 was that the smoothOrder annotation is preferred!?
> So the better solution would be to use smoothOrder here, and also for cp_T.

I agree with @thorade. For functions which are not too involved, I think it is preferrable to leave the tools compute the derivatives via AD. This is certainly the case for these functions, which are based on generalized polynomials, logarithms, and other expressions for which derivative are computed in the most straightforward way by AD.

This makes the codebase smaller and more reliable, and avoids potential errors. AFAIK also OpenModelica supports this option.

casella:
> Well, stating the derivative was easier for me (as I do not know which smooth order to set).

The original functions are analytic, so they can be differentiated as many times as one wants. In practice, thermodynamic systems are at most index 1, so I guess you can write smoothOrder = 2 and that will be more than enough for all practical applications.

beutlich:
> 
> 
> I would suggest to add the `smoothOrder=2` annotation as explained in my comment.

OK. Resolved.

beutlich:
@hubertus65 @wischhusen @casella Please add your review.

beutlich:
@thorade OK for you to merge?

thorade:
Yes, good to merge, IIRC this was the version that passed the test.

PR #2912, opened 2019-05-03
Rename tau_constant to tau_nominal

tobolar:
close #2863

PR #2924, opened 2019-05-10
Fix ModelicaTest.Rotational.TestBraking.

perost:
- Update ModelicaTest with the change in #2912.


casella:
@beutlich, it would be nice to also patch this onto 3.2.3 maintenance

beutlich:
> @beutlich, it would be nice to also patch this onto 3.2.3 maintenance

This is a fix for a master-only change and must not be backported.

PR #2925, opened 2019-05-10
Improve the numerical conditioning of AST_BatchPlant example case

casella:
The AST_BatchPlant example case in Modelica.Fluid is very numerically ill-conditioned, which makes it critical for the MSL, that must run with all Modelica tools producing the same results. This PR addresses two specific issues, trying to improve the situation.

The first is that the two groups of valves V1-V2-V3 and V4-V5-V6 are closed during some time intervals (including initialization). This means that the ideal (zero-volume) tee junctions volume2 and volume8 find themselves enclosed by closed valves only, and thus have an undetermined pressure. Hence, the Jacobian of the corresponding nonlinear system is singular. Although some tools may somehow manage this situation, producing an arbitrary value of the pressure, I don't think this is a good idea for a MSL example, that should run and produce the same results in all tools. 

The second is that the two pipes pipePump1B1 and pumpPipe2B2 show extremely fast changes of pressure when the valves enclosing them are opened, producing fantastically high values of the pressure derivative for extremely short periods, and sometimes causing solver failure because of excessive stiffness of the system of equations. By the way, this is not realistic, as the real plant most likely uses rubber pipes, whose elasticity dominates the very low compressibility of water by several orders of magnitude. Unfortunately, the pipe model does not (yet) have a provision to include this effect, that would make the model a lot more realistic and numerically well-behaved.

I addressed the first issue by adding a small leakage coefficient (1e-5) to valves V3 and V6, so that the pressure of the two ideal junctions is well-defined and equal to the atmosferic ones when the valves are closed. Note that there is actually no leak through these valves during the simulation, because when V3 and V6 are closed, so are the other ones, completely blocking any flow.

I addressed the second issue by increasing the pressure losses of the valves surrounding the pipes when they are opened, still keeping them low compared to the static head that the pumps need to counteract. This is supposed to make the system less stiff. I also increased changed the value of dp_nom for the regularization of the pressure losses inside the pipes, for the same reason. Finally, I reduced the reporting interval to 1 s, which is also beneficial to help the solver deal with the state events that cause such abrupt pressure changes.

The important results of the simulations (the flows and the levels) are virtually unchanged.

I successfully tested the updated model with both Dymola and OpenModelica.

### DIFF
--- a/Modelica/Fluid/Examples/AST_BatchPlant.mo
+++ b/Modelica/Fluid/Examples/AST_BatchPlant.mo
@@ -95,6 +95,7 @@ package AST_BatchPlant
     Modelica.Fluid.Valves.ValveDiscrete V3(
       redeclare package Medium = BatchMedium,
       m_flow_nominal = 1,
+      opening_min = 1e-5,
       dp_nominal = 100)
       annotation (Placement(transformation(extent={{-150,210},{-130,230}})));
     Fittings.TeeJunctionIdeal volume2(
@@ -106,6 +107,7 @@ package AST_BatchPlant
     Modelica.Fluid.Valves.ValveDiscrete V6(
       redeclare package Medium = BatchMedium,
       m_flow_nominal = 1,
+      opening_min = 1e-5,
       dp_nominal = 100)
       annotation (Placement(transformation(extent={{130,210},{110,230}})));
     Fittings.TeeJunctionIdeal volume8(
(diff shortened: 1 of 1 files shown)


casella:
If the PR is accepted, I'd suggest to also patch it onto 3.2.3 maintenance

beutlich:
> 
> 
> If the PR is accepted, I'd suggest to also patch it onto 3.2.3 maintenance

Since it only is an example model and not a severe library issue there is no urgent need to backport this change.

casella:
> 
> 
> > If the PR is accepted, I'd suggest to also patch it onto 3.2.3 maintenance
> 
> Since it only is an example model and not a severe library issue there is no urgent need to backport this change.

In fact, we are trying to achieve 100% successful verification of MSL 3.2.3 with OpenModelica. Without this change, we get a simulation failure, which is not really the tool's fault.

As I understand MSL 3.2.3 will be here to stay for a while, I would suggest to port this fix there as well.

casella:
It's in [this paper](https://www.modelica.org/events/modelica2006/Proceedings/sessions/Session6b3.pdf). Shall I open a separate PR to add it to the documentation?

beutlich:
> I would suggest to port this fix there as well.

OK then from my side.

> Shall I open a separate PR to add it to the documentation?

Yes, please.

casella:
> > Shall I open a separate PR to add it to the documentation?
> 
> Yes, please.

Done in PR #2939. Please not that the description of the system is located in the package documentation, not in the specific model documentation. I guess that is fine as well.

casella:
@wischhusen, @GallLeo, @hubertus65, @rfranke, I'm looking for two positive reviews to this PR. Thanks for your help!

PR #2929, opened 2019-05-13
Convert BasicTransformer

beutlich:


beutlich:
> However, I do not understand why the code moved to Interfaces is not shown here, even though I can see it branch convert-BasicTransformer...

What do you mean? Modelica.Electrical.Machines.Interfaces.PartialBasicTransformer already is available in MSl v3.2.3.

PR #2930, opened 2019-05-13
Convert FirstOrderTemperaturCoefficient and SecondOrderTemperaturCoefficient

beutlich:
Originated by #2284.

### DIFF
--- a/Modelica/Resources/Scripts/Conversion/ConvertModelica_from_3.2.3_to_4.0.0.mos
+++ b/Modelica/Resources/Scripts/Conversion/ConvertModelica_from_3.2.3_to_4.0.0.mos
@@ -31,6 +31,10 @@ convertClass("Modelica.Electrical.Machines.Icons.QuasiStationaryTransformer",
               "Modelica.Electrical.Machines.Icons.QuasiStaticTransformer");
 convertClass("Modelica.Electrical.Machines.BasicMachines.QuasiStationaryDCMachines",
               "Modelica.Electrical.Machines.BasicMachines.QuasiStaticDCMachines");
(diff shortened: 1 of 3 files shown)


christiankral:
So why are two units put into `Modelica.SIunits.Conversions.NonSIunits` as they actually are both SI units? 

Both `Ohm/degC` and `Ohm/degC2` actually are the same units as `Ohm/K` and `Ohm/K2` as the denominator **always** refers to a **temperature difference** and a temperature difference is the same in `degC` and `K`. 

Note, that the already existing unit

```
type LinearTemperatureCoefficient = Real(final quantity = "LinearTemperatureCoefficient", final unit="1/K");
```

also refers to a **temperature difference**. 


Therefore, it may make sense to change the units from `Ohm/degC` and `Ohm/degC2` to `Ohm/K` and `Ohm/K2` and put them directly to `SIunits`.

beutlich:
> So why are two units put into `Modelica.SIunits.Conversions.NonSIunits` as they actually are both SI units?

Because of the discussion in #2288 and especially your request to have the types in SIunits, see https://github.com/modelica/ModelicaStandardLibrary/pull/2288#issuecomment-320187959.

AHaumer:
I think we dont need Ohm/K and Ohm/K2, we only need 1/K and 1/K2 which are the same as 1/degC and 1/degC2.

beutlich:
OK, closing then.

@christiankral Please provide a PR to properly convert the obsolete types (or units) FirstOrderTemperaturCoefficient and SecondOrderTemperaturCoefficient as requested by #2944.

PR #2933, opened 2019-05-14
Add Jenkins test for binaries

sjoelund:
Added a Jenkins pipeline
- Test the things Travis CI does on Jenkins
- Added gcc makefile target for ModelicaExternalC
- Fixed parallel make for gcc makefile target
- Use a parameter to optionally run more compiler targets
- For autotools, compile MINGW 32/64-bit, OSX 64-bit, Linux
  i686/amd64/armhf using both static and dynamic linking.
- For Linux i686/amd64 some simple executables are built, linking
  some selected functions and tests that they run as expected.
- For gcc makefile, compile and test static linking.

Should be generated under: https://test.openmodelica.org/jenkins/job/ModelicaAssociation/job/ModelicaStandardLibrary/

Hopefully this would prevent the C sources from breaking since I added at least some simple tests. It's around 5 minutes for the default short version and I guess 10 minutes for the version testing more compilers and flags.


sjoelund:
I am also considering if we should remove the Library binaries and instead let the CI generate a zip-file with  compiled binaries and the Libraries without README.md, etc (only the libraries ready to be installed).

beutlich:
Why Jenkins and not native Travis CI as https://travis-ci.org/tbeu/ModelicaStandardTables does for the MSL table sources?

sjoelund:
Mostly because I know Jenkins and it's able to upload files to Modelica.org in a safe and easy way. It's also used for the Modelica specification.

sjoelund:
Sure, I could move them. But I am against removing the C-Sources from future MSL releases due to a need for source-code FMU support sometime in the future: https://github.com/modelica/ModelicaSpecification/issues/2145

beutlich:
Sorry, if confsuing. Please only move your new test files and keep the existing C-Sources as is.

sjoelund:
I have updated the location of the tests. Is there anything we need to have in CI?

beutlich:
> 
> 
> I have updated the location of the tests. Is there anything we need to have in CI?

Hm, I meant Modelica**Test**\Resources\C-Sources (as test for Modelica\Resources\C-Sources).

sjoelund:
Ah, that I'm not so sure of if it's desired (it's a different package, so why would they be there?). They could be moved to .CI/Test I suppose...

beutlich:
Similar as ModelicaTest, I do not want to [distribute](https://github.com/modelica/ModelicaStandardLibrary/releases/download/v3.2.3%2Bbuild.2/ModelicaStandardLibrary_v3.2.3.zip) the new test files along with the MSL. Usually, we have the strict separation of MSL and ModelicaTest. Having them in .CI\Test might be possible, but I am not in favour of it, since the tests can be run independent of any CI.

sjoelund:
I think the problem with distributing those simple tests with ModelicaTest is that ModelicaTest might be in a different location than the rest of the MSL (which has the binaries that are tested); in particular Modelica might be in a directory called "Modelica 3.2.3" or similar. So it would need some paths or env.variables for the test script. I think .CI/Test would be slightly easier for this reason.

beutlich:
This still would work out if MSL and ModelicaTest are located in same parent directory and relative directories are used for the test. Anyway, I am fine with putting them in .CI/Test.

Modelica/Resources/C-Sources/ModelicaStandardTablesUsertab.c
```diff
@@ -48,7 +48,7 @@
 #include "ModelicaUtilities.h"
 
 #if defined(DUMMY_FUNCTION_USERTAB)
-#if (defined(__clang__) || defined(__GNUC__)) && !(defined(__apple_build_version__) || defined(__MINGW32__) || defined(__MINGW64__) || defined(__CYGWIN__))
+#if (defined(__clang__) || defined(__GNUC__)) && !(defined(__apple_build_version__) || defined(__APPLE__) || defined(__MINGW32__) || defined(__MINGW64__) || defined(__CYGWIN__))
```
beutlich: This fix is independet of the CI enhancements and should go to a separate PR.

dietmarw:
I agree with @beutlich that we should keep the MSL as lean as possible and trying to keep all purely test-related out of the MSL dir.

Modelica/Resources/C-Sources/ModelicaStandardTablesUsertab.c
sjoelund: I have moved it to #2950 . I expect this PR to be unstable until #2950 is merged.

sjoelund:
Well, we need 2 people to accept reviews regardless of how many people know the topic...

The CI web UI looks something like this:
https://test.openmodelica.org/jenkins/blue/organizations/jenkins/ModelicaAssociation%2FModelicaStandardLibrary/detail/PR-2933/6

Basically what is new is that we compile and link some of the C-Sources now, so we can be a bit more sure that things don't accidentally break (there is even limited checking of compilation on Windows as MINGW is used for cross-compilation). If you think this is a good idea you may as well accept the changes if you think waiting ~4 minutes before you can merge a commit.

This also performs the same checks as the Travis job today (parsing of the library using OpenModelica and moparser), in addition to checking that files are UTF-8 encoded.

beutlich:
> My other idea is to create ModelicaTest models including the five newly created C source files (using the Include annotation) and checking the similar results, but from a Modelica environment. Therefore, it would be good to move .CI/_.c to ModelicaTest/Resources/C-Sources/_.c.

@sjoelund What do you think?

sjoelund:
The problem with that is that you then wouldn't be able to test the PR without having a Modelica environment to test that with. They are mainly there to test that the libraries are possible to link and don't have missing dependencies.
For ModelicaTest it would be simpler to just test the external "C" functions as usual. Maybe also adding function prototypes for testing the older libraries (there are different entrypoints for the tables with the old input arguments and new ones, right?)

dietmarw:
Do I understand it right that the original issue of having the test sources placed under `.CI/...` has now been resolved? @beutlich ?

beutlich:
My idea would be to use the new source files in ModelicaTest/Resources/C-Sources/ both for theh CI test and for (not yet available) ModelicaTest models. That should be possible, right?

sjoelund:
The C-files would then need to exist in Modelica (not ModelicaTest). Because they directly refer to C-sources in Modelica. And when shipped and packaged, Modelica and ModelicaTest may exist in different directories or have the version number as part of it (as the language specification describes the way MODELICAPATH works) so you can't use `../../Modelica/Resources`. Using IncludeDirectory would not help other than to find the test in the Modelica/Resources.

beutlich:
OK, got it. We can't use relative includes then. 👎

dietmarw:
So what does that mean for this PR?

dietmarw:
> So what does that mean for this PR?
@beutlich, @sjoelund ping

sjoelund:
Did someone remove the webhook for this? Does not seem to be anything remaining and the last job did not run automatically

beutlich:
> 
> 
> Did someone remove the webhook for this? Does not seem to be anything remaining and the last job did not run automatically

It run before I force-pushed. I believe it did nor run, because we did not set the check as "Required".

sjoelund:
I checked the settings and there is no webhook anymore, so I guess the job never runs anymore (including parsing)

Issue #2936, opened 2019-05-15
Illegal usage of conditional component in Modelica.Magnetic.FundamentalWave.Examples.BasicMachines.SMEE_DOL


perost:
The `Modelica.Magnetic.FundamentalWave.Examples.BasicMachines.SMEE_DOL` model contains the declaration:
```modelica
Modelica.SIunits.Current irRMS = sqrt(smee.ir[1]^2+smee.ir[2]^2)/sqrt(2);
```
`ir` inside `smee` is a conditional component though, so using it in this context is not allowed since conditional components may only be modified and/or used in connections.

christiankral:
In this example it makes only sense to have the damper cage present. Otherwise the machine cannot start. So the parameter `useDamperCage` will be fixed so that the cage cannot be disabled.

sjoelund:
I think the following might work (didn't check it):
```mo
  Modelica.Blocks.Interfaces.RealOutput ir[2];
  Modelica.SIunits.Current irRMS = sqrt(ir[1]^2+ir[2]^2)/sqrt(2);
equation
  connect(ir, smee.ir);
```

christiankral:
This solution is not working. I tested it in Dymola and OpenModelica.

My conclusions: In case of having the cage not present, the connection 
```modelica
connect(ir, smee.ir); 
```
is removed and there are then no equations for the determineation of `ir[2]`...

sjoelund:
Oh yeah, I guess both need to have the same if condition...

AHaumer:
In Modelica.Electrical.Machines.BasicMachines.SynchronousInductionMachines.* and Modelica.Magnetic.FundamentalWave.BasicMachines.SynchronousInductionMachines.* we have the same problem: The DamperCage is continitional. Nevertheless we want to have an information about the DamperCage currents (if the DamperCage is present), and we need them for initialization (if the DamperCage is present). The solution with connect is fine and should work. So we know the components of the DamperCage current.
The problem is to provide the RMS of the DamperCage current only if the the DamperCage is present. 
I doubt that the conditional definition (and calculation) of iRMS works (since conditional variables are used).
@christiankral : We could define iRMS in the conditional DamperCage), so we could connect a conditional iRMS to to damperCage.iRMS (preferrably in the machine model).

sjoelund:
```mo
  Modelica.Blocks.Interfaces.RealOutput ir[2] if useDamperCage;
[...]

PR #2937, opened 2019-05-16
Make damper cage always present in SMEE_DOL example

christiankral:
See #2936

### DIFF
--- a/Modelica/Magnetic/FundamentalWave.mo
+++ b/Modelica/Magnetic/FundamentalWave.mo
@@ -4072,7 +4072,7 @@ to speed to achieve constant current and torque.</p>
           p=2,
           Jr=0.29,
           Js=0.29,
-          useDamperCage=true,
+          final useDamperCage=true,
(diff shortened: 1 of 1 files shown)


beutlich:
@HansOlsson I requested your review here since I am not sure if the PR is enough to fix the issue.

beutlich:
> The change does not suffice.

Thanks. That is what I suspected.

christiankral:
What could the solution then be? Remove the variable `irRMS` referring to the conditional component?

christiankral:
OK, I did not yet see https://github.com/modelica/ModelicaStandardLibrary/issues/2936#issuecomment-494300565
I will investigate this proposal

christiankral:
https://github.com/modelica/ModelicaStandardLibrary/issues/2936#issuecomment-494300565 seems not be the solution, see https://github.com/modelica/ModelicaStandardLibrary/issues/2936#issuecomment-496002400

AHaumer:
resolved by #2956

Issue #2953, opened 2019-06-02
PowerConverters DC/DC


AHaumer:
I'd recommend to enhance the DC/DC-examples of Modelica.Electrical.PowerConverters:
- add inductor and smoothing capacitor to StepDown
- delete example StepDown_RL
- change default parameterization of StepDown and StepUp
This makes the examples more realistic and comparable:
- StepDown results in 60 V -> 12 V / 1.2 A
- StepUp results in 60 V -> 75 V / 1.2 A

AHaumer:
see PR #2955

beutlich:
After closing #2955 w/o merge, is this now covered by #2970?

beutlich:
I believe it is resolved by #2970. Please reopen otherwise.

PR #2955, opened 2019-06-02
Enhanced DC/DC examples of PowerConverters

AHaumer:
Enhanced DC/DC examples of PowerConverters (issue #2953)


AHaumer:
Will be replaced by a new PR

PR #2958, opened 2019-06-04
issue #2936 same changes for maint/3.2.3 as PR #2956 for master

AHaumer:
As far as I see we should apply the same solution to maint/3.2.3 - right?


beutlich:
@sjoelund OK for you? Ist it really backward-compatible?

Issue #1493, opened 2017-01-14
Mechanics.Rotational.Interfaces.PartialFriction.w_small


modelica-trac-importer:
**Reported by ahaumer on 15 May 2014 15:27 UTC**
Default value for w_small=1e10 doesn't seem to be really small ...

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/1493

modelica-trac-importer:
**Comment by beutlich on 15 May 2014 19:14 UTC**
I guess this is a duplicate of #701.

modelica-trac-importer:
**Comment by otter on 20 Jun 2014 14:35 UTC**
Replying to [comment:1 beutlich]:
> I guess this is a duplicate of #701.

Yes. Here is a copy of the essential description of #701:

 w_small was introduced due to a bug report, where at zero velocity a reinit was triggered and then the friction handling logic is no longer correct. This was a hack to at least give the possibility to fix this manually for a specific model. On the other hand, introducing w_small in general gave other problems for connected friction elements. Therefore, it was decided to introduce w_small but with a large value. If the special case with reinit occurs, then the user has to set w_small to a small value.

 This is of course not nice, but there is currently no better idea how to handle it (this would require to remove the hack "reinit" and introduce proper handling of Dirac impulses).

 The description text of variable w_small seems to be also correct ("set to low value only if such impulses can occur").

modelica-trac-importer:
**Modified by beutlich on 6 Jan 2015 21:35 UTC**

AHaumer:
@tobolar 
There's a rather old ticket regarding the issue with w_small.
I seems to be a wrong parameter description and incorrect or missing documentation. 
Would you please talk to Martin Otter and take care of that?

tobolar:
@AHaumer I can do this. But is there any new/particular update regarding https://github.com/modelica/ModelicaStandardLibrary/issues/1493#issuecomment-272661458?

MartinOtter:
> 
> 
> @tobolar
> There's a rather old ticket regarding the issue with w_small.
> I seems to be a wrong parameter description and incorrect or missing documentation.
> Would you please talk to Martin Otter and take care of that?

Nothing has changed. Please, read my previous comment. Please, do not make a change because this might break existing models.

AHaumer:
Then PLEASE adapt the documentation - w_small=1E10 is total misleading!

MartinOtter:
[...]

PR #2963, opened 2019-06-07
Implement simple vehicle model

### LINKED ISSUE #2940: Translational driving resistance

AHaumer:
implement simple vehicle model (resolve #2940 / alternative PR to #2954)


christiankral (commented):
Looks OK

christiankral (approved):

AHaumer:
@tobolar we kindly ask you to review, seems like a mature proposal.

beutlich:
@AHaumer Please not that the rebased on force-pushed the branch. Please reset on your side to get these changes.

beutlich (dismissed):
The Vehicle examples mentions vehicle1D1/vehicle1D2 which are not available in the model.

beutlich:
I believe there will be some general discussion if such a Vehicle model belongs to Modelica.Mechanics.Translational.Components at all. Might be also possible to keep the Vehicle example model and move the Vehicle component model to Modelica.Mechanics.Translational.Examples.Utilities to basically gain a new example model.

I have no strong preference here and would leave the decision to the libary officers @christiankral, @tobolar and @MartinOtter.

christiankral:
> I believe there will be some general discussion if such a Vehicle model belongs to Modelica.Mechanics.Translational.Components at all. Might be also possible to keep the Vehicle example model and move the Vehicle component model to Modelica.Mechanics.Translational.Examples.Utilities to basically gain a new example model.
> 
> I have no strong preference here and would leave the decision to the libary officers @christiankral, @tobolar and @MartinOtter.

I am in favor of having the vehicle model under `Components` instead of `Utilities`. `Components` were the very first place where I would look at, if I were looking for the vehicle model.

AHaumer:
@beutlich since I'm not familiar with the enhanced git processes:
How do I reset on my side to get these changes?

beutlich:
In the git bash call:

```sh
git fetch --all
git reset --hard 2940A
```

tobolar (dismissed):
To me, the `Vehicle` component would be fine now.

But there is another issue. Even if I understand the reason to add signal input to `SignForce`, there are two reasons to reject this. First, the documentation of that component is obsolete since it still says it is a "constant force" component. This could be simply fixed.

Second, if this input signal would be accepted, `ConstantForce` and (at least) `ForceStep` shall be changed accordingly in order to have consistent definition/usage of those three components. Then the component `Force` could be removed since it would be of the same functionality like `ConstantForce`.

So we probably need another solution.

Additionally, the usage of rotationalMultiSensor is questionable as it is not necessary for the model functionality. This sensor can be simply add outside of the model, if needed by user.

And I would prefer to use wheel radius as parameter instead of wheel diameter. The radius is anyway used in docu in the equation of equivalent additional mass as well as in idealRollingWheel instance.

beutlich:
@AHaumer @christiankral Could you please care for the requested changes by the reviewers such that we can merge it finally?

AHaumer:
Followed @tobolar 's review regarding the multiSensor and wheel diameter resp. radius.
I've to check another solution for rolling resistance.

AHaumer:
@tobolar 
I understand that you prefer to have the sources consistent (also comparing Translational and Rotational). Anyhow, we need the rolling resistance dependent on rolling resistance coefficient and inclination and additionally there's the need for regularization around zero speed to exclude static friction (for simplicity). Would you accept a separate RollingResistance model (replacing the SignForce from which the input would be removed)? Placed in Sources or in Components?

tobolar:
> we need the rolling resistance dependent on rolling resistance coefficient and inclination

I fully accept this, no question about it.

Having RollingResistance is a preferable solution for me. As this is a force element, it could be placed in Sources. But on the other side it not really fit the concept of that package. So either we can
- let it in Components, or
- define it nested in the `Vehicle` model.

@beutlich Any preference from the MSL design view?

Some comments on `RollingResistance`:
- the inputs `inclination` and `cr` may not necessarilly by conditional; but if this is desired for some reason then `usecrInput` and `useInclinationInput` of the instance `rollForce` in the `Vehicle` shall be propageted from the corresponding parameters (instead of setting them both to true only - then you need not to set `crConstant` and `inclinationConstant` here)
- description of the parameter `fNormal`: this force is generally called "wheel load" or "vertical force"
- `inclinationConstant` has a default value, but `crConstant` just a start value; I guess both can have default; see also `Modelica.UsersGuide.ParameterDefaults`
- even `v0`could heve a default value, cf. `w_small` in `Modelica.Mechanics.Rotational.Components.Clutch`

AHaumer:
@tobolar 
You're right, Sources is not the right place for RollingResistance, it fits in Components. Please don't hide it in the Vehicle model.
I'll follow your hints regarding the inputs and the default values.

AHaumer:
@tobolar - would you please have a look at OneWayClutch (not Clutch):
w_small=1e10 doesn't seem like a "Relative angular velocity near to zero ..." to me.

tobolar:
> Would you please have a look at OneWayClutch (not Clutch).
w_small=1e10 doesn't seem like a "Relative angular velocity near to zero ..." to me.
Should this be a new ticket or am I wrong?

If any doubt, simply open a new ticket.

HansOlsson:
> If any doubt, simply open a new ticket.

There is already one: #1493

tobolar:
@AHaumer 
> - even v0 could have a default value, cf. w_small in Modelica.Mechanics.Rotational.Components.Clutch

My idea was not to set v0 to the value of w_small but to show some reference model where a regularization parameter (here w_small) has a default value as well. So your original value of v0 = 0.1 m/s is fine. Just not being a start value but default instead.

> For cr I have no reasonable default, therefore only a start value (strongly depends on material pairing).

[Rill](https://www.crcpress.com/Road-Vehicle-Dynamics-Fundamentals-and-Modeling/Rill/p/book/9781439838983) (formerly OTH :-) ) says: "Under normal conditions ... the rolling resistance for typical passenger car tires is in the range of 0.01 to 0.02."
So we can take an average: crConstant = 0.015 or the value you suggest as a start: 0.01.

> - description of the parameter fNormal: this force is generally called "wheel load" or "vertical force"

I have also found "vertical wheel load". IMO we can use this one.

tobolar:
@AHaumer Since you use fDrag, fRoll and fGrav in the documentation I suggest to rename the corresponding instances (dragForce, rollForce and gravForce, resp.) accordingly.

Modelica/Mechanics/Translational.mo
```diff
@@ -1712,6 +1712,125 @@ An eddy current brake reduces the speed of a moving mass. Kinetic energy is conv
 </html>"));
     end EddyCurrentBrake;
 
+    model Vehicle "One-dimensional vehicle with driving resistances"
+      extends Modelica.Icons.Example;
+      import Modelica.Constants.g_n;
+      parameter Modelica.SIunits.Mass m=100 "Mass of vehicle";
+      parameter Modelica.SIunits.Length R=0.25 "Radius of wheel";
+      parameter Modelica.SIunits.Area A=1 "Cross section of vehicle";
+      parameter Real cw=0.5 "Drag resistance coefficient";
```
christiankral: The variable name `cw` is a German thing. In English `cd` may be used, see [Wikipedia](https://en.wikipedia.org/wiki/Drag_coefficient). Possibly a native English speaking person can comment on this. 

@mtiller or @harmanpa Could you possibly comment on this?

Modelica/Mechanics/Translational.mo
```diff
@@ -3280,6 +3399,384 @@ following references, especially (Armstrong and Canudas de Wit 1996):
 </html>"));
     end IdealRollingWheel;
 
+    model RollingResistance "Resistance of a rolling wheel"
+      extends Modelica.Mechanics.Translational.Interfaces.PartialForce;
+      import Modelica.Constants.pi;
+      parameter SI.Force fNormal(start=0) "Wheel load due to gravity";
```
christiankral: The name `fNormal` may not be a good choice as the normal force is calculated by the expression

`fNormal*cos(atan(inclination_internal))`

in line 3440. As I understand the model `fNormal` is the weight (mass * acceleration of gravity) = `m * g` only, see line 3577.

Modelica/Mechanics/Translational.mo
harmanpa: The commonly understand variable name would be Cd

AHaumer:
I followed @tobolar 's review.
Regarding @christiankral 's review and @harmanpa 's comment, I changed the parameter names from cw to Cd and cr to Cr (to be consistent), as well as fNormal to fWeight "Wheel load due to gravity" in model RollingResistance.
See 6ee8640ff..c95c538c3

AHaumer:
@tobolar regarding the question about w_small (OneWayClutch):
@HansOlsson is right, there's already a ticket #1493 
Don't know why this ticket was closed - I reopened it, asking you kindly to take care of that.

christiankral (approved):
Looks OK from my side.

Modelica/Mechanics/Translational.mo
tobolar: I have also found c_D in https://www.crcpress.com/Road-and-Off-Road-Vehicle-System-Dynamics-Handbook/Mastinu-Ploechl/p/book/9781138075290 (but originating from a German-speaking author anyway ;-) )

tobolar:
When renaming parameter cw to Cd, also the input cr has been changed to Cr. This leads to a) breaking the recommendation 3 "Instance names" in `Modelica.UsersGuide.Conventions.ModelicaCode.Naming` (well, this is just a recommendation and thus not that bad) and b) to some discrepancies in the documentation (esp. of RollingResistance):
>The rolling resistance coeffcient _Cr_ is either constant (given by the parameter `CrConstant`) or prescribed by the input `Cr`. 

I suggest the following naming in the model:

1. _Cr_ to indicate in general the rolling resistance coeffcient.
2. `cr` for the real input. (so change it back to lower case)
3. Parameter `CrConstant` is fine.

AHaumer:
@tobolar done by c9d6441a1..e09ff284f

tobolar:
@AHaumer In the documentation of Vehicle there is:
> The rotational inertia J can be expressed as an equivalent additional mass mEq = J/R2, where R is the wheel radius, but only the vehicle mass influences rolling resistance and inclination resistance.

The relationshp between mEq and J for idealized wheel contact is clear but as the model doesn't deal with the acceleration only one can't simply transform J to mEq and add it to the vehicle's mass. This is even stated in the subclause. So what is the real benefit of this information for the user?

Modelica/Mechanics/Translational.mo
```diff
@@ -3677,8 +3677,9 @@ is not taken into account.
       annotation (Icon(coordinateSystem(preserveAspectRatio=false), graphics={
         Line(points={{-80,-70},{80,-70}}, color={0,0,0}),
         Line(points={{-80,0},{85.607,-1.19754}},
-                                          color={0,0,0}, origin={-3,-49}, rotation=15),
-        Ellipse(extent={{-130,-120},{-30,-20}}, lineColor={0,0,0}, startAngle=0, endAngle=15),
+          color={0,0,0}, origin={-3,-49}, rotation=15),
+        Ellipse(extent={{-130,-120},{-30,-20}}, lineColor={0,0,0}, startAngle=0, endAngle=15,
+          closure=EllipseClosure.None),
```
beutlich: @tobolar EllipseClosure is legal Modelica 3.4. Why was it removed here?

tobolar: Looking at the icon, the arc/ellipse is located between the two lines of the "road". It represents the inclination angle. As the two existing lines close the angle already, the additional closure of the ellipse would imply the lines being doubled. Just look at the following icon's cutoff (Dymola). 
grafik

beutlich: Ah, sorry. I was wrong here.

beutlich:
@tobolar Is this ready to merge then? If yes, please approve!

tobolar:
I wait for reply of @AHaumer to https://github.com/modelica/ModelicaStandardLibrary/pull/2963#issuecomment-565379766.

AHaumer:
@tobolar 
What's the problem of pointing out something that is clear to you but maybe not to all users. 
Maybe you should add a remark that mEq is only valid for acceleration, not for rolling resistance.
Or if you prefer remove the remark about mEq totally, both solutions are ok for me.

tobolar:
@AHaumer 
> What's the problem of pointing out something that is clear to you but maybe not to all users.

This is fine in general.
If we retain the information on mEq, we shall underline this is *not* applicable in this particular model. I will modify the documentation accordingly.

beutlich:
>
> **Anton Haumer** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

@AHaumer Your mail address anton.haumer@oth-regensburg.de is not associated with your GitHub account. If you do not want to add this address, you should avoid using that mail address for MSL contributions in future. I still can merge this PR (even w/o this CLA issue corrected).

tobolar:
@AHaumer 
> fine for me

I suppose this concerns both my comment https://github.com/modelica/ModelicaStandardLibrary/pull/2963#issuecomment-572496344 and that of @beutlich https://github.com/modelica/ModelicaStandardLibrary/pull/2963#issuecomment-573417236.

tobolar (approved):

PR #2986, opened 2019-06-21
Update icon and function comments of Modelica.Math.Matrices.LAPACK

beutlich:


beutlich:
> 
> 
> Looks good.
> I was a bit surprised by the change from "computes" to "Compute" (etc), but after a bit of thought it makes sense - as it is now a sentence without a singular subject.

Yes, see also 1937e390e7ebdad1577ef4aab1a2ff286bd73e69.

Issue #2957, opened 2019-06-03
Implement idealized battery


AHaumer:
I would like to implement an idealized battery with no-load voltage dependent on state of charge and inner resistance. Such a simple model is used over and over again in (simple) models of electric vehicles. Furthermore, it could be a trigger to start developemnt of more advanced battery models with effects like:
- temperature dependency
- self-discharge
- RC-elements to model dynamic response on current peaks
- constructing a battery stack of series and parallel connected cells with different behaviour
- ageing
The ElectricalEnergyStorage Library was intended to address these goals, but it has been overdone - the library is too complicated and it is not in a good state.
As a starting point, I would place an "IdealizedBattery" in Modelica.Electrical.Analog.Sources.
I have prepared a PR but I would like to discuss where to place the component.
@christiankral , @christophclauss , @kristinmajetta what is your opinion?

AHaumer:
For many tasks such a simple model would be sufficient, especially for teaching. 
I've attaches small lib (still based on 3.2.3) to give an idea.
[Battery.zip](https://github.com/modelica/ModelicaStandardLibrary/files/3249094/Battery.zip)

AHaumer:
To start the discussion early ;-) with a positive argument:
Yes even such small enhancements and new models attract new users, helping them to solve their problems with low effort using Modelica (and nothing like M**b/S**k, ...). It also supports teaching and students working on their thesis. Therefore I don't judge such implementations as "contaminating the MSL with applications specific models".

christiankral:
I am very much in favor of having simple battery models in the Modelica Standard Library (MSL). The proposed model is based on a table based OCV versus SOC curve. This makes a lot of sense for many applications. However, over and under charging the battery (SOC > 1 and SOC < 0) shall somehow be handled with asserts. 

Additionally an even much simpler battery model might be included in the MSL, based on a purely linear OCV versus SOC curve. In this case a linear capacitor model can be used. Then the upper and lower voltage limit and the nominal charge of the battery determine the capacitance. 

For now I can think of three places where such battery models can be placed:
1. `Modelica.Electrical.Analog.Sources` because this where I would look for an electric energy source
2. `Modelica.Electrical.Analog.Batteries` in case we want to distinguish ideal electrical sources from rechargeable batteries and we plan to include up to approx. 10 different battery models 
3. `Modelica.Electrical.Batteries` in case we have plans to include more than approx. 1-10 models in the near future.

[...]

PR #2992, opened 2019-06-22
Fix spelling and enhance battery example of #2957

beutlich:
I merged #2959 too early, while some small fixes are required.

Would be also very nice if the SOC is visualized in the icon of BatteryOCV_SOCtable using DynamicSelect.


beutlich:
Modelica.Electrical.Analog.Batteries.Partials should probably renamed to BaseClasses (following the discussions in #609).

beutlich:
> 
> 
> See [beutlich#1](https://github.com/beutlich/ModelicaStandardLibrary/pull/1)

Applied your patch.

PR #2997, opened 2019-06-22
Move FluidHeatFlow models of Interfaces.Partials to Interfaces

### LINKED ISSUE #2479: Partial interfaces in particular package in Modelica.Thermal.FluidHeatFlow.Interfaces
There are collected partial interface classes in a particular subpackage `Modelica.Thermal.FluidHeatFlow.Interfaces.Partials`. This is not common in the MSL and the classes should be moved one level higher, i.e. into `Modelica.Thermal.FluidHeatFlow.Interfaces`.

christiankral:
close #2479

### DIFF
--- a/Modelica/Resources/Scripts/Conversion/ConvertModelica_from_3.2.3_to_4.0.0.mos
+++ b/Modelica/Resources/Scripts/Conversion/ConvertModelica_from_3.2.3_to_4.0.0.mos
@@ -61,12 +61,6 @@ convertClass("Modelica.Electrical.Machines.BasicMachines.QuasiStationaryDCMachin
               "Modelica.Electrical.Machines.BasicMachines.QuasiStaticDCMachines");
 convertClass("Modelica.Electrical.Machines.BasicMachines.Components.BasicTransformer",
               "Modelica.Electrical.Machines.Interfaces.PartialBasicTransformer");
-convertClass("Modelica.Thermal.FluidHeatFlow.Components.IsolatedPipe",
-              "Modelica.Thermal.FluidHeatFlow.Components.Pipe");
-convertClass("Modelica.Thermal.FluidHeatFlow.Components.HeatedPipe",
-              "Modelica.Thermal.FluidHeatFlow.Components.Pipe");
-convertClass("Modelica.Thermal.FluidHeatFlow.Interfaces.Partials.Ambient",
-              "Modelica.Thermal.FluidHeatFlow.Interfaces.Partials.SinglePortLeft");
 convertClass("Modelica.Blocks.Sources.Clock",
               "Modelica.Blocks.Sources.ContinuousClock")
 convertClass("Modelica.Electrical.Digital.Sources.Clock",
@@ -235,6 +229,12 @@ convertClass("Modelica.Magnetic.FluxTubes.Interfaces.PartialGenericHysteresis",
               "Modelica.Magnetic.FluxTubes.BaseClasses.GenericHysteresis")
 convertClass("Modelica.Magnetic.FluxTubes.Interfaces.PartialGenericHysteresisTellinen",
               "Modelica.Magnetic.FluxTubes.BaseClasses.GenericHysteresisTellinen")
+convertClass("Modelica.Thermal.FluidHeatFlow.Components.IsolatedPipe",
+              "Modelica.Thermal.FluidHeatFlow.Components.Pipe");
+convertClass("Modelica.Thermal.FluidHeatFlow.Components.HeatedPipe",
+              "Modelica.Thermal.FluidHeatFlow.Components.Pipe");
+convertClass("Modelica.Thermal.FluidHeatFlow.Interfaces.Partials.Ambient",
+              "Modelica.Thermal.FluidHeatFlow.BaseClasses.SinglePortLeft");
 convertClass("Modelica.Thermal.FluidHeatFlow.Interfaces.Partials.TwoPort",
               "Modelica.Thermal.FluidHeatFlow.BaseClasses.TwoPort");
 convertClass("Modelica.Thermal.FluidHeatFlow.Interfaces.Partials.SinglePortLeft",
@@ -243,8 +243,6 @@ convertClass("Modelica.Thermal.FluidHeatFlow.Interfaces.Partials.SinglePortBotto
               "Modelica.Thermal.FluidHeatFlow.BaseClasses.SinglePortBottom");
 convertClass("Modelica.Thermal.FluidHeatFlow.Interfaces.Partials.SimpleFriction",
               "Modelica.Thermal.FluidHeatFlow.BaseClasses.SimpleFriction");
-convertClass("Modelica.Thermal.FluidHeatFlow.Interfaces.Partials.Ambient",
-              "Modelica.Thermal.FluidHeatFlow.Interfaces.Ambient");
 convertClass("Modelica.Thermal.FluidHeatFlow.Interfaces.Partials.AbsoluteSensor",
               "Modelica.Thermal.FluidHeatFlow.Interfaces.AbsoluteSensor");
 convertClass("Modelica.Thermal.FluidHeatFlow.Interfaces.Partials.RelativeSensor",
--- a/Modelica/Resources/Scripts/Conversion/ConvertModelica_from_3.2.3_to_4.0.0.mos
+++ b/Modelica/Resources/Scripts/Conversion/ConvertModelica_from_3.2.3_to_4.0.0.mos
@@ -219,7 +219,6 @@ convertClass("Modelica.Magnetic.QuasiStatic.FluxTubes.Interfaces.PartialFixedSha
               "Modelica.Magnetic.QuasiStatic.FluxTubes.BaseClasses.FixedShape");
 convertClass("Modelica.Magnetic.QuasiStatic.FluxTubes.Interfaces.PartialLeakage",
               "Modelica.Magnetic.QuasiStatic.FluxTubes.BaseClasses.Leakage");
-// Issue #496
 convertClass("Modelica.Magnetic.FluxTubes.Interfaces.PartialTwoPortsElementary",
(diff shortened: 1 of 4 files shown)


christiankral:
If we want to have a base class package, we probably should think about that in all packages, don't you agree? Unification of structure and naming.

* Pure interface definitions -> Interfaces
* Base classes such as TwoPorts, BaseMachine, SISO -> BaseClasses (or simply Base?)

It would then make sense to have separate ticket on this issue.

HansOlsson:
> If we want to have a base class package, we probably should think about that in all packages, don't you agree?

I'm not sure if that is a good idea. 

The problem is that "pure interface" is actually somewhat unclear - consider Modelica.Electrical.Analog.Interfaces.{TwoPin,OnePort,ConditionalHeatport} - one could say that they are "pure interface", but still all of them contain some equations and variables and are thus not really "pure".

Unless we can find a really good definition I believe it is better to have a few somewhat ambigous case in Interfaces, instead of creating a separate "BaseClasses"-package for them - whereas other subpackages may have a "BaseClasses" package.

AHaumer:
I agree with Hans.
On one hand there are pure interfaces, i.e. connectors, and partials with only a few equations (e.g. balance equations), and on the otehr hand there are partials with a rather large set of equations. For the latter ones I prefer a "BaseClass" package.

christiankral:
@ahaumer: So in case of `FluidHeatFlow` there should only go `SimpleFriction` to an additional package `BaseClasses`

image

beutlich:
Conflicts are resolved now.

AHaumer:
> 
> 
> @AHaumer: So in case of `FluidHeatFlow` there should only go `SimpleFriction` to an additional package `BaseClasses`
> 
> image

I'd vote for moving AbsoluteSensor, RelativeSensor and FlowSensor to Interfaces, amd the rest (SimpleFriction, TwoPort, SinglePoetLeft and SinglePortBottom) to BaseClasses.

christiankral:
@beutlich I need your support again. I didn't predict these conflicts and don't know what I did wrong when I force-pushed my commits.

beutlich:
> 
> 
> @beutlich I need your support again. I didn't predict these conflicts and don't know what I did wrong when I force-pushed my commits.

Resolved.

ConvertModelica_from_3.2.3_to_4.0.0.mos needs to be updated, too.

christiankral:
@ahaumer Can you please check if 7b3b832 is OK using

```modelica
convertClass("Modelica.Thermal.FluidHeatFlow.Interfaces.Partials.Ambient",
              "Modelica.Thermal.FluidHeatFlow.BaseClasses.SinglePortLeft");
``` 

or shall the partial Ambient model go to `ObsoleteModelica4.mo`?

Comparing `Modelica.Thermal.FluidHeatFlow.Sources.Ambient`:

* MSL 3.2.2: `extends Interfaces.Partials.Ambient;`
* MSL 3.2.3: `extends Modelica.Thermal.FluidHeatFlow.Interfaces.Partials.SinglePortLeft(final Exchange=true, final T0=293.15, final T0fixed=false);`
* MSL 4.0.0: `extends Modelica.Thermal.FluidHeatFlow.BaseClasses.SinglePortLeft(final Exchange=true, final T0=293.15, final T0fixed=false);`

beutlich:
@christiankral After I resolved the merge conflicts, can you please check if the sensor texts are all as expected (by 0638879d1f722f03f6680c257d7162d95769c1b7)?

beutlich:
> 
> 
> @christiankral After I resolved the merge conflicts, can you please check if the sensor texts are all as expected (by [0638879](https://github.com/modelica/ModelicaStandardLibrary/commit/0638879d1f722f03f6680c257d7162d95769c1b7))?

It's broken again. Will fix it now.

christiankral:
Everything looks good now.

tobolar:
> On one hand there are pure interfaces, i.e. connectors, and partials with only a few equations (e.g. balance equations), and on the otehr hand there are partials with a rather large set of equations. For the latter ones I prefer a "BaseClass" package.

I'm afraid the definition is still ambiguous. How to interpret "a rather large set of equations"?
There are various indicators which can be used to sort those partial classes to either "interfaces" or "base classes". I can imagine it could be the definition an icon which indicates where to find a partial class. Then, `SinglePortLeft` should be located in "Interfaces" whereas 'AbsoluteSensor' in "base classes". (No idea where to place `SimpleFriction` then.)

I have no preference how to solve this problem, but we shall be aware of defining a clear rule for this. Otherwise users (and we as well) will be confused. At the moment, I would prefer to reject the new `Modelica.Thermal.FluidHeatFlow.BaseClasses` and would move all classes simply into `Modelica.Thermal.FluidHeatFlow.Interfaces`.

AHaumer:
@tobolar
I see quite a big difference between {TwoPort, SinglePortxxx} and xxxSensor:
xxxSensor contains only basic assumptions for the Sensor, .e.g. no pressure drop -> Interfaces
TwoPort contains a large portion of physical modeling concepts with complex equations -> BaseClasses
We should rely on the knowledge of the LibraryOfficers instead of cerating rules and rules and rules ...

tobolar:
> contains only basic assumptions for the Sensor, .e.g. no pressure drop -> Interfaces

Following this idea, e.g. `Modelica.Mechanics.MultiBody.Interfaces.PartialTwoFrames` or `Modelica.Mechanics.Rotational.Interfaces.PartialTwoFlanges` should stay located as they are, i.e. in Interfaces. In contrast, a "two port" component (which - following the naming logic - appears to be of the similar functionality) shall be located in "base classes" This would be confusing.

But maybe the problem is rather in the naming of the component, because `Modelica.Thermal.FluidHeatFlow.BaseClasses.TwoPort` can rather be compred to `Modelica.Mechanics.Rotational.Interfaces.PartialCompliantWithRelativeStates`. So possibly the solution would be to rename `TwoPort` and, optionally, a new real `TwoPort` class can be included in "Interfaces".

(Btw. are really all the equations from current `TwoPort` needed e.g. for `Modelica.Thermal.FluidHeatFlow.Components.Pipe`?)

beutlich:
> (Btw. are really all the equations from current `TwoPort` needed e.g. for `Modelica.Thermal.FluidHeatFlow.Components.Pipe`?)

Is this a separate issue @AHaumer ?

beutlich:
Here's some visual support for the proposed restructuring:

Before:
before

After:
after

beutlich:
> 
> 
> > (Btw. are really all the equations from current `TwoPort` needed e.g. for `Modelica.Thermal.FluidHeatFlow.Components.Pipe`?)
> 
> Is this a separate issue @AHaumer ?

At least it is not a regression of MSL4 development ➡️ off-topic.

beutlich:
@tobolar Since you were involved in the discussion I'll wait for your aproval, too, before going to merge.

AHaumer:
@beutlich yes all euqations from PartialTwoPort are needed.

AHaumer:
The library officers agree on the solution. 
If we don't get a third review from @tobolar :
 @beutlich how can we reolve the conflicts and merge?

beutlich:
> 
> 
> The library officers agree on the solution.
> If we don't get a third review from @tobolar :
> @beutlich how can we resolve the conflicts and merge?

Conflicts are resolved. Going to merge soon.

tobolar:
> We should rely on the knowledge of the LibraryOfficers instead of cerating rules and rules and rules ...

This is surely a case for package-specific classes but I mean classes called `TwoPin`, `TwoPort` or `TwoFlanges` shall definitely be placed consistently.
If this could not be followed, then the classes shall be renamed in order to avoid ambiguity.

If we cannot decide it now, I would follow the suggestion of @HansOlsson (provided I understand it well):
> Unless we can find a really good definition I believe it is better to have a few somewhat ambigous case in Interfaces, instead of creating a separate "BaseClasses"-package for them...

beutlich:
https://github.com/modelica/ModelicaStandardLibrary/pull/3053 added some definition about which classes should be Interfaces and which BaseClasses. Still leaves some room of interpretation, but better what we had before.

tobolar:
> ...better what we had before.

Really? "containing connectors and partial models" sounds clear to me. "simple partial interface models" and "models with physical equations other than trivial" don't.

beutlich:
I chatted with @tobolar and I understand his preference for strict guide lines and consistent structuring. His reason is that similar types (TwoFlanges in MultiBody, TwoPin in E.Analog etc.) should have identical semantics and should always be structured in package Interfaces, such that MSL users can rely on it. This is given by this PR in both improving the naming and not losing consistency. Hence, it was okay to merge it finally (though it would have been better to wait for a clear and accepted resolution across the four active main developers).

christiankral:
> If we cannot decide it now, I would follow the suggestion of @HansOlsson (provided I understand it well):
> 
> > Unless we can find a really good definition I believe it is better to have a few somewhat ambigous case in Interfaces, instead of creating a separate "BaseClasses"-package for them...

In case of `FluidHeatFlow` it is clear that `SimpleFriction` definitely is a base class and shall thus go to `BaseClasses`. The models 

* `Modelica.Thermal.FluidHeatFlow.Interfaces.Partials.TwoPort`
* `Modelica.Thermal.FluidHeatFlow.Interfaces.Partials.SinglePortLeft`*
* `Modelica.Thermal.FluidHeatFlow.Interfaces.Partials.SinglePortBottom`

may be ambiguous to some users. So `BaseClasses` was not created for these three models. Instead `BasesClasses` was used as it was created in favor of `SimpleFriction`.

christiankral:
Shall we create a separate ticket for the discussion on what a base class may be?

Implemented in #3053 (full wording) so far:

* **Interfaces** containing connectors and simple partial interface models. 
* **BaseClasses** containing partial models with physical equations other than trivial and balance equations.

beutlich:
Might be good to have it as future issue but currently I am afraid of introducing another major restructuring site. For MSL v4.0.0 it sound to ambitious to me.

Issue #2972, opened 2019-06-14
Cubic interpolation Moody diagram


dofware:
Dear all,
we have an issue about the interpolation cubic spline expression at the following modelica standard fluid library line:

https://github.com/modelica/ModelicaStandardLibrary/blob/e2dafbe029f7171130f46800077251834a40e716/Modelica/Fluid/Pipes.mo#L2930

We are not able to obtain the same "Re" expression as the one implemented. In particular the "c3" coefficient in the expression multipies the "dx^2" term, while to us the final expression should be different.

Could anybody help us about this issue?

Thanks a lot

beutlich:
@casella @thorade Can you please have a look at it.

beutlich:
@wischhusen I noticed that the aboved mentioned function Modelica.Fluid.Pipes.BaseClasses.WallFriction.PartialWallFriction.massFlowRate_dp.interpolateInRegion2 is a duplicate of Modelica.Fluid.Dissipation.Utilities.Functions.General.CubicInterpolation_Re (previously named CubicInterpolation_DP) and Modelica.Fluid.Pipes.BaseClasses.WallFriction.PartialWallFriction.pressureLoss_m_flow.interpolateInRegion2 of Modelica.Fluid.Dissipation.Utilities.Functions.General.CubicInterpolation_lambda (previously named CubicInterpolation_MFLOW), respectively. Is there any chance we can avoid the (undocumented) protected functions interpolateInRegion2 in favour of the well-documented functions in Fluid.Dissipation.Utilities.Functions.General?

wischhusen:
Yes, it should be possible by calling the function from Fluid.Dissipation.Functions.General. But I have to admit, that this particular function is undocumented, too. It seems that we picked it from the Modelica_Fluid package of Summer 2006. I am not sure, who implemented the original code but it must have been from the people that created the predecessor of the Modelica.Fluid library. Eventually, Martin Otter recalls who did the work?

beutlich:
[...]

PR #3001, opened 2019-06-23
Avoid code duplication in Modelica.Fluid.Pipes.BaseClasses.WallFriction.Detailed

beutlich:
See #2972.


beutlich:
@wischhusen @thorade What do you think?

thorade:
For my understanding: Would that other function be identical or would it already have fixes for issue #2972?

beutlich:
This PR actually is independent of #2972.

PR #3004, opened 2019-06-24
Fix uninitialized memory and realpath behaviour in ModelicaInternal_fullPathName

### LINKED ISSUE #3003: ModelicaInternal_fullPathName truncates trailing '/' on Linux
The function ModelicaInternal_fullPathName truncates trailing '/' on Linux, but not on Windows.
The truncating behavior in the Linux realpath needs to be compensated for, by checking for existing trailing '/' in the input argument and then add this later to the resulting fullName if needed.

beutlich:
close #3003

### DIFF
--- a/Modelica/Resources/C-Sources/ModelicaInternal.c
+++ b/Modelica/Resources/C-Sources/ModelicaInternal.c
@@ -598,8 +598,8 @@ _Ret_z_ const char* ModelicaInternal_fullPathName(_In_z_ const char* name) {
 #if (_BSD_SOURCE || _XOPEN_SOURCE >= 500 || _XOPEN_SOURCE && _XOPEN_SOURCE_EXTENDED || _POSIX_VERSION >= 200112L)
     {
         /* In case of realpath: Retain trailing slash to match _fullpath behaviour */
-        size_t len = strlen(name) - 1;
-        if ('/' == name[len] || '\\' == name[len]) {
+        size_t len = strlen(name);
+        if (len > 0 && ('/' == name[len - 1] || '\\' == name[len - 1])) {
             strcat(fullName, "/");
         }
     }
--- a/Modelica/Resources/C-Sources/ModelicaInternal.c
+++ b/Modelica/Resources/C-Sources/ModelicaInternal.c
@@ -30,6 +30,10 @@
 */
 
 /* Release Notes:
+      Jun. 24, 2019: by Thomas Beutlich
+                     Fixed uninitialized memory and realpath behaviour in
+                     ModelicaInternal_fullPathName (ticket #3003)
+
       Jun. 28, 2018: by Hans Olsson, Dassault Systemes
                      Proper error message when out of string memory
                      in ModelicaInternal_readLine (ticket #2676)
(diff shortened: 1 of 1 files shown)


beutlich:
> 
> 
> I don't like the "len = strlen(name)-1;" due to the potential that it is an empty string.
> Instead of investigating that in detail I would prefer:
> 
> ```
> ... len =strlen(name);
> if (len>0 && ('/' ==name[len-1] || '\\'==name[len-1])) {
> ```

Right. Will update the PR.

beutlich:
For the sake of completeness: [Here](https://gist.github.com/beutlich/88e7662045e4197d8e9add8c5756b44f) is my test case.

Modelica/Resources/C-Sources/ModelicaInternal.c
```diff
@@ -588,9 +592,18 @@ _Ret_z_ const char* ModelicaInternal_fullPathName(_In_z_ const char* name) {
...
     }
-    fullName = ModelicaAllocateString(strlen(tempName));
+    fullName = ModelicaAllocateString(strlen(tempName) + 1);
     strcpy(fullName, tempName);
     ModelicaConvertToUnixDirectorySeparator(fullName);
+#if (_BSD_SOURCE || _XOPEN_SOURCE >= 500 || _XOPEN_SOURCE && _XOPEN_SOURCE_EXTENDED || _POSIX_VERSION >= 200112L)
+    {
+        /* In case of realpath: Retain trailing slash to match _fullpath behaviour */
+        size_t len = strlen(name);
+        if (len > 0 && ('/' == name[len - 1] || '\\' == name[len - 1])) {
```
sjoelund: Looking for '\\' here will give the wrong behaviour. You are changing the path from a file to a directory (remember: '\\' on POSIX systems is a file name, not a path separator).
The only problem is if you want MINGW to be handled differently. In that case those defines might require `&& !defined(MINGW)` and using `_fullpath` instead (if that's what you want to mimic, why not call it directly?).

beutlich: I tested Cygwin only I need to admit. Can you update my example above with the test case you have in mind, please! Thanks.

beutlich: Hm, I am rather confused with your comment. Whereas Cygwin uses the realpath implementation, MinGW uses the _fullpath implementation (and is not affected by this PR). Tested with

```
tbeu@LAPTOP-RUH9UABA /cygdrive/C/temp
$ gcc --version
gcc (GCC) 7.4.0
Copyright (C) 2017 Free Software Foundation, Inc.
Dies ist freie Software; die Kopierbedingungen stehen in den Quellen. Es
gibt KEINE Garantie; auch nicht für MARKTGÄNGIGKEIT oder FÜR SPEZIELLE ZWECKE.
```

and

```
C:\temp>gcc --version
gcc (tdm64-1) 5.1.0
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

sjoelund: Oh, true... You don't include unistd.h for MINGW... And for Cygwin you don't use \ as path separator. So on that line the check for '\\' at the end of the string should simply be removed as we're dealing with Unix systems.

beutlich: Right. Will update PR now,

PR #3006, opened 2019-06-25
Update Matio library v1.5.17

### LINKED ISSUE #2905: ModelicaMatIO.c conversion warnings
When compiling a model that makes use of ModelicaMatIO.c I get the following warnings:
C:\Program Files\Dymola 2020 Beta 3\Source\ModelicaMatIO.c(6615): warning C4244: '=': conversion from 'size_t' to 'double', possible loss of data
C:\Program Files\Dymola 2020 Beta 3\Source\ModelicaMatIO.c(6624): warning C4244: '=': conversion from 'size_t' to 'double', possible loss of data

These warnings are due to the lines 6615 and 6624 in ModelicaMatIO.c:
tmp = matvar->dims[0];
and
tmp = matvar->dims[1];

I suggest that the following be used 
tmp = (double)matvar->dims[0];
and
tmp = (double)matvar->dims[1];

### LINKED ISSUE #3050: Error reading mat file in Modelica.Blocks.Tables.CombiTable1D
The attached mat file "my_test.mat" was created using the attached Python script "test_scipy.py" using Anaconda python and scipy version 1.2.1.  When I try to simulate the attached model "test3.mo" I get the following error during initialization:

    InflateFieldNames: inflate returned stream end

I've tested that I get the same error in Dymola 2018 FD01 (Modelica version 3.2.2, 2016-04-03, build 3 (2016-04-03 08:44:41Z)), 2020 (Modelica versions 3.2.3, 2019-01-23, build 2 (2019-03-20 12:00:00Z) and 4.0.0-dev, 2019-01-23, build 1 (2019-01-23 12:00:00Z), with appropriate class name changes) and JModelica.

I can load my_test.mat without error into both MATLAB (version 2015b) and GNU Octave (version 5.1.0).  If I re-save the mat file (using the following command) from either MATLAB or Octave, the model test3.mo then simulates without error.

    load my_test.mat
    save -v6 my_test.mat struct1

I've also tried disabling compression when saving the file with "do_compression=False".  In that case the simulation still fails but the error message is different:

    fields[0] not MAT_T_MATRIX, fpos = 784

Can anyone comment if this is an error with the mat file, as created by scipy, or a limitation of the ModelicaMatIO code, (or user error)?

[190626.zip](https://github.com/modelica/ModelicaStandardLibrary/files/3393950/190626.zip)

beutlich:
Superseding #2877. Closes #2905, closes #3050.

### DIFF
--- a/Modelica/Resources/C-Sources/ModelicaMatIO.c
+++ b/Modelica/Resources/C-Sources/ModelicaMatIO.c
@@ -442,15 +442,12 @@ static size_t InflateSkip(mat_t *mat, z_streamp z, int nbytes);
 static size_t InflateSkipData(mat_t *mat,z_streamp z,enum matio_types data_type,int len);
 static size_t InflateVarTag(mat_t *mat, matvar_t *matvar, void *buf);
 static size_t InflateArrayFlags(mat_t *mat, matvar_t *matvar, void *buf);
-static size_t InflateDimensions(mat_t *mat, matvar_t *matvar, void *buf);
-static size_t InflateVarNameTag(mat_t *mat, matvar_t *matvar, void *buf);
+static size_t InflateRankDims(mat_t *mat, matvar_t *matvar, void *buf, size_t nbytes, mat_uint32_t** dims);
 static size_t InflateVarName(mat_t *mat,matvar_t *matvar,void *buf,int N);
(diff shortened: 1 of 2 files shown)


beutlich:
> I haven't examine it in detail, but assume it is just the new version.

Exactly.

Issue #813, opened 2017-01-14
Remove classes that were obsoleted (e.g., because of name fixes) and provide coversion scripts


modelica-trac-importer:
**Reported by dietmarw on 11 Sep 2012 16:42 UTC**
This is a collection ticket for all those classes that need to be removed because they were simply introduced for backward compatibility and which normally extend from `Modelica.Icons.ObsoleteModel`. Typical examples are classes that were misspelled or simply renamed.

Classes to be removed and converted:
 
- [x] Modelica.Mechanics.MultiBody.Sensors.{TansformAbsoluteVector, TansformRelativeVector} by #2987
- [x] Modelica.Math.{baseIcon1, baseIcon2} by #3008
- [x] Modelica.Math.{tempInterpol1, tempInterpol2} by #2989
- [x] Modelica.Blocks.Interfaces.Adaptors by #3012
- [x] Modelica.StateGraph.Temporary by #2880
- [x] Modelica.Electrical.Digital.Converters.{LogicToXO1, LogicToXO1Z} by #2889
- [x] Modelica.Electrical.Machines.BasicMachines.Components.BasicTransformer by #2929
- [x] Modelica.Fluid.Icons by #3009
- [x] Modelica.Thermal.FluidHeatFlow by #3013
- [x] Modelica.Icons.{Library, Library2, GearIcon, MotorIcon, Info} by #3011
- [x] Modelica.SIunits.Conversions.NonSIunits.{FirstOrderTemperaturCoefficient, SecondOrderTemperaturCoefficient} by #2981
- [x] Modelica.SIunits.RadiantExitance by #2991
- [x] Modelica.SIunits.FluxoidQuantum by #2893
- [x] Modelica.SIunits.{Loundness, LoundnessLevel} by #3025
----

Migrated-From: https://trac.modelica.org/Modelica/ticket/813

modelica-trac-importer:
**Comment by dietmarw on 1 Aug 2015 21:36 UTC**
Milestone renamed

modelica-trac-importer:
**Comment by dietmarw on 1 Aug 2015 21:39 UTC**
Milestone renamed

beutlich:
See the [wiki](https://github.com/modelica/ModelicaStandardLibrary/wiki/Changes-targeting-MSL-v4.0.0-requiring-conversion) on how to remove/convert obsolete classes for MSL v4.0.0.

PR #3011, opened 2019-06-28
Remove obsolete icons in Modelica.Icons and also Modelica.Mechanics.MultiBody.Icons.MotorIcon

beutlich:
See #813.


beutlich:
> So probably `Modelica.Mechanics.MultiBody.Icons.MotorIcon` could be deleted as well and changed accordingly in the conversion script.

Done.

PR #3016, opened 2019-07-01
Rename Modelica.Icons.{RotationalSensor, TranslationalSensor}

### LINKED ISSUE #340: Rename and remove icons from Modelica.Icons
**Reported by christiankral on 26 Mar 2010 06:27 UTC**
At the 66th design meeting in Hamburg Harburg it was decided to incorporate fully backwards compatible icons. With the next conversion script in MSL 4.0 the following changes shall be applied:

The following icons of `Modelica.Icons` shall be renamed
 * `Icons.RotationalSensor` -> `Icons.RoundSensor`
 * `Icons.TranslationalSensor` -> `Icons.RectangularSensor`

The following icons shall be removed (left side) and instead other icons (right side) shall be used
 * `Icons.Info` -> `Icons.Information`
 * `Icons.Library` -> `Icons.Package`
 * `Icons.Library2` -> `Icons.Package`

The following icons shall be indicated as deprecated
 * `Icons.MotorIcon`
 * `Icons.GearIcon`

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/340

beutlich:
Closes #340.

### DIFF
--- a/Modelica/Icons.mo
+++ b/Modelica/Icons.mo
@@ -439,7 +439,7 @@ directly utilized by a user.
 </html>"));
   end MaterialProperty;
 
-  partial class RotationalSensor "Icon representing a round measurement device"
+  partial class RoundSensor "Icon representing a round measurement device"
 
     annotation (
       Icon(coordinateSystem(preserveAspectRatio=false, extent={{-100,-100},{100,100}}), graphics={
@@ -472,9 +472,9 @@ directly utilized by a user.
 This icon is designed for a <strong>rotational sensor</strong> model.
 </p>
 </html>"));
-  end RotationalSensor;
+  end RoundSensor;
 
-  partial class TranslationalSensor
+  partial class RectangularSensor
     "Icon representing a linear measurement device"
 
     annotation (
@@ -499,7 +499,7 @@ This icon is designed for a <strong>rotational sensor</strong> model.
 <p>
 This icon is designed for a <strong>translational sensor</strong> model.
 </p></html>"));
-  end TranslationalSensor;
+  end RectangularSensor;
 
   partial function Function "Icon for functions"
 
--- a/ModelicaTestConversion4.mo
+++ b/ModelicaTestConversion4.mo
@@ -1493,7 +1493,7 @@ Conversion test for <a href=\"https://github.com/modelica/ModelicaStandardLibrar
 
   package Icons
     extends Modelica.Icons.ExamplesPackage;
-    model Issue813 "Conversion test for #813"
+    model Issue340 "Conversion test for #340"
       extends Modelica.Icons.Example;
       package P1
         extends Modelica.Icons.Library;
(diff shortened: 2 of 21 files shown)


Modelica/Icons.mo
```diff
@@ -472,9 +472,9 @@ directly utilized by a user.
 This icon is designed for a <strong>rotational sensor</strong> model.
 </p>
 </html>"));
-  end RotationalSensor;
+  end RoundSensor;
 
-  partial class TranslationalSensor
+  partial class RectangularSensor
     "Icon representing a linear measurement device"
```
tobolar: The element's desription (and documentation) shall be modified accordingly. Whereby the `RotationalSensor` is "Icon representing a __round__ measurement device", the `RectangularSensor` is "... a __linear__ measurement device".

To be really honest, is is not clear at all when to use which icon for a sensor component.

ModelicaTestConversion4.mo
```diff
@@ -931,7 +931,7 @@ Conversion test for <a href=\"https://github.com/modelica/ModelicaStandardLibrar
 
   package Icons
     extends Modelica.Icons.ExamplesPackage;
-    model Issue813 "Conversion test for #813"
+    model Issue340 "Conversion test for #340"
```
tobolar: Does not this concerns both 813 and 340 issues?

christiankral:
Main first | Main secon
------------- | ----------------
`sensorVoltage` | `voltageSensor`
`voltageSignal` | `signalVoltage`
`` | ``
`` | ``
`` | ``
`` | ``
v

christiankral:
> I stumble a bit about the naming of the icons. As I stated in [#2301 (comment)](https://github.com/modelica/ModelicaStandardLibrary/issues/2301#issuecomment-361894874), I rather prefer to follow the Indexing rules decribed e.g. in https://www.asindexing.org/wp-content/uploads/2014/04/ASI-BestPracticesIndexing2015-linked-updated-copyright-v3-02-26-16.pdf. Following that rules, a subheading (round, rectangular) shall be predeceased by a main heading (sensor). The suggested names follow indeed reverse order.

Is there a membership required to access the PDF file provided by the link above? I get the following error message when trying to access the PDF file: `Error 404 - Page not found!`

christiankral:
The new notation would require to rename **most** of the actual class names. So someone familiar with the structure of the MSL, will not recognize it any more after the re-design. So I am not convinced that this is the right way to go. However, even though I could live with most of the new names, the class name `MachineInduction` sounds definitely a lot to crazy to me. In case of the electric machines a hierarchical structure could solve the naming problem, somehow: 

* `Machines.Induction.SquirrelCage`
* `Machines.Induction.SlipRing`
* `Machines.Synchronous.PermanentMagnet`
* `Machines.Synchronous.ElectricalExcited`
* `Machines.Synchronous.ReluctanceRotor`

In the very same way, for example, sources could be structured as:

* `Analog.Sources.Voltage.Signal` 
* `Analog.Sources.Voltage.Constant` 
* `Analog.Sources.Current.Signal` 
* `Analog.Sources.Current.Constant` 
etc. 

I think it were difficult to decide when a new package shall be introduced. Are two machine models enough? So the structuring would also change the existing hierarchy of the MSL a lot. To me it seems to be a an overkill, too.

tobolar:
> Is there a membership required...

No idea why the link is not working. Please find the pdf via https://www.asindexing.org/best-indexing-practices.

tobolar:
@christiankral 
> In case of the electric machines a hierarchical structure could solve the naming problem,...

This would definitely be the consequence. But I agree it would require too much effort. So let us just make a notice of this.

AHaumer:
Please stop this discussion ... I'm strictly agianst such crazy things like
Machines.Induction.SquirrelCage or Analog.Sources.Voltage.Signal

christiankral:
@AHaumer This PR does not introduce any crazy namings. This PR sticks to the existing naming conventions. So you can just go ahead regularly review this PR.

PR #3017, opened 2019-07-01
Remove all obsolete classes from Modelica.Electrical.Spice3.Internal

### LINKED ISSUE #608: Remove unneeded function bjtRenameParametersType
**Modified by dietmarw on 13 Sep 2011 11:16 UTC**
In 33337aea75ee77ea9ada61a0dfaf4ef5db32bffc the function `bjtRenameParametersType` was removed. But since this removal  is a non-backward compatible change (user models might already use that function) the removal was reverted in 00580861925fd409b177a5cf5c3e564d0a76dd82. 

The function should be removed for the next non-backward compatible version of the MSL which also needs to supply a conversion script.

<sub>Keep this ticket as a reminder of that task.</sub>

----

**Reported by dietmarw on 13 Sep 2011 11:16 UTC**
In 33337aea75ee77ea9ada61a0dfaf4ef5db32bffc the function `bjtRenameParametersType` was removed. But since this removal  is a non-backward compatible change (user models might allready use that function) the removal was reverted in 00580861925fd409b177a5cf5c3e564d0a76dd82. 

The function should be removed for the next non-backward compatible version of the MSL which also needs to supply a conversion script.

<sub>Keep this ticket as a reminder of that task.</sub>

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/608

### LINKED ISSUE #784: Remove Bjt.Bjt3Variables (was: Documentation of Bjt package unclear)
**Reported by dietmarw on 25 Jul 2012 12:54 UTC**
The info layer of `Modelica.Electrical.Spice3.Internal.Bjt` says:
 This package Bjt3 contains functions and records with data of the Bjt3 bipolar transistor models.

Should this not be "Bjt" instead of "Bjt3"? If yes it should be fixed in 3.2.1 maintenance and in trunk.

Also within `Bjt` package is `Bjt.Bjt3Variables`. This looks like a mistake right?

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/784

beutlich:
There is neither conversion provided nor possible. Instead, a conversion message is raised, if internal classes of Spice3 are to be converted.

There is one backward-compatibility breaking change in Modelica.Electrical.Spice3.Internal.MOS2, since the final parameter vp was declared as return value of the (obsolete and now removed) function Modelica.Electrical.Spice3.Internal.Mos2.mos2ModelLineParamsInitEquations, but mos2ModelLineParamsInitEquationsRevised has a different return type. In the current PR I simply decided to remove that parameter since it was not used (think so) and set final anyway, thus could not be modified in derived classes.

Closes #608, closes  #784.

### DIFF
--- a/Modelica/Electrical/Spice3.mo
+++ b/Modelica/Electrical/Spice3.mo
@@ -4696,30 +4696,20 @@ on the model behaviour.
 
   //-------------------------------------------------------------------------------------------------------------------------------------
 
-    final parameter Spice3.Internal.Mos2.Mos2ModelLineVariables
-                                                 vp=
-           Spice3.Internal.Mos2.mos2ModelLineParamsInitEquations(
-             p,
-             C,
-             m_type) "Model line variables";
     final parameter Spice3.Internal.Mos2.Mos2Calc
-                                   c1=Spice3.Internal.Mos.mos2CalcInitEquations(
+                                   c1=Spice3.Internal.Mos.mos2CalcInitEquationsRevised(
              p,
-             C,
-             vp,
              m) "Precalculated parameters";
     final parameter Spice3.Internal.Mos2.Mos2Calc
-                                   c2=Spice3.Internal.Mos.mos2CalcCalcTempDependencies(
+                                   c2=Spice3.Internal.Mos.mos2CalcCalcTempDependenciesRevised(
              p,
-             C,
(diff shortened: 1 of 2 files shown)


beutlich:
@christophclauss @JoeRiel Please review.

PR #3018, opened 2019-07-01
Fix text in icons of sensors

christiankral:
This is a first attempt to get #2628 resolved. I did not yet touch `Modelica.Mechanics.MultiBody` and `Modelica.Fluid` as the library officers may do a much better job than I can make. I propose to go ahead with this PR and try to close it and then let the library officers of MultiBody and Fluid create additional PRs.

However, I tried to get all the sensors to look the same as much as possible. My PR is a massive change of the visual appearance. So suppose to checkout the branch https://github.com/christiankral/ModelicaStandardLibrary/tree/text_sensor and go through each of the packages indicated under Labels.


christiankral:
@beutlich For a reason I do not understand there occurred some conflicts. Could you please have a look at these conflicts as I do not actually understand what their cause is...

christiankral:
I will add some notes in the User's Guide of the MSL on the proper text icon design.

beutlich:
> 
> 
> @beutlich For a reason I do not understand there occurred some conflicts. Could you please have a look at these conflicts as I do not actually understand what their cause is...

Should be good again now.

christiankral:
> In Modelica.Magnetic.QuasiStatic.FluxTubes.Sensors.Transient.FundamentalWavePermabilitySensor, if both outputs have the same unit I'd prefer to have the unit in the circle as it was before.

image

The left output is `mur` with SI unit 1, and the right output is `mu` with SI unit H/m. I thought that unity quantity does not need an indication of `1`

christiankral:
> You use `VA`, but also `N.m`. For consistency, should it be `V.A` (or `Nm`)?

I think it is good to have a consistent way of indicating units. I will therefore change `VA` to `V.A`

maltelenz:
The images at the bottom of the documentation of `modelica://Modelica.UsersGuide.Conventions.Icons` do not show up. Maybe they have not been added yet?

The following have no spacing around the text, which makes them look cramped:
```
Modelica.Mechanics.Translational.Sensors.RelSpeedSensor
Modelica.Mechanics.Translational.Sensors.RelAccSensor
```
Apart from that, this looks good to me.

christiankral:
In Dymola some SI units are not even rendered due to the small text size. I will thus change the design proposal and increase the text size for SI units of more than three characters.

christiankral:
Give me a couple more minutes -- we can always roll back if required.

maltelenz:
> In Dymola some SI units are not even rendered due to the small text size. I will thus change the design proposal and increase the text size for SI units of more than three characters.

Can you give an example? I'm curious how SystemModeler renders them.

christiankral:
I did two measures:
1. I changed the text box format for longer texts of the RotationalSensor to improve readibility
2. I moved the pointer of the TranslationalSensor to the left to have more space available:

image

christiankral:
> > In Dymola some SI units are not even rendered due to the small text size. I will thus change the design proposal and increase the text size for SI units of more than three characters.
> 
> Can you give an example? I'm curious how SystemModeler renders them.

Have a look at c786c9f

christiankral:
Dymola renders the latest proposal of b8dfa02 as:

image

maltelenz:
SystemModeler doesn't elide either of f22333f and b8dfa02. I think we might only elide tags (%name, %par, ...), and assume the library developer knows what they are doing for fixed length strings.

Although I prefer the smaller size, I'm fine with the new one as well.

beutlich:
@christiankral Is this OK to get merged?

christiankral:
@beutlich Yes, it can get merged

PR #3024, opened 2019-07-05
Remove Modelica.Electrical.Analog.Basic.OpAmpDetailed.Pi

beutlich:


AHaumer:
Would like to see a comment from @christophclauss and / or @kristinmajetta

Issue #2519, opened 2018-04-26
Including figures in HTML documentation


christiankral:
The actual guidelines for including figures are summarized in ```Modelica.UsersGuide.Conventions.Documentation.Format.Figures```.  

```html
<table border="0" cellspacing="0" cellpadding="2">
  <caption align="bottom">Caption starts with a capital letter</caption>
  <tr>
    <td>
      <img src="modelica://Modelica/Resources/Images/Blocks/PID_controller.png"
           alt="PID_controller.png">
    </td>
  </tr>
</table> 
```

It is advised to use a table in order to include a figure. The table captions are included by 

```html
<caption align="bottom">Caption starts with a capital letter</caption>
```
 
The reason for this is to align the caption below the figure such that the caption does not become any wider than the figure width.  The attribute align="bottom" does work, e.g., in OpenModelica, but it doesn't in Dymola -- where the caption is place above the figure. I am not quite happy with this design of the figure layout since it is quite a complex construction. 

I guess that a lot of documentation of the MSL is, however, implemented in a different way. Even the referenced example ```Modelica.Blocks.Examples.PID_Controller``` is designed differently:

```html
<p>
In the following figure, results of a typical simulation are shown:
</p>

<img src="modelica://Modelica/Resources/Images/Blocks/PID_controller.png"
     alt="PID_controller.png"><br>

<img src="modelica://Modelica/Resources/Images/Blocks/PID_controller2.png"
     alt="PID_controller2.png">
```
 
In HTML5 the tag [figcaption](https://www.w3schools.com/tags/tag_figcaption.asp) is supported. So going for the figure format

```html
<p>
 <figure>
  <img src="modelica://Modelica/Resources/Images/Blocks/PID_controller.png" 
       alt="PID_controller.png">
  <figcaption>Caption starts with a capital letter.</figcaption>
</figure> 
</p>
```

were on my opinion the way better choice. I so have no idea what HTML version is supported in Modelica. I suspect the Modelica specification does not say anything about it.  I actually tested this implementation with OpenModelica -- where it works -- and Dymola -- where the tag ```<figcaption>``` seems not to be supported.

Any suggestions or comments here?

christiankral:
This issue was actually triggered by #2517 where ```valign="bottom"``` was mentioned...

beutlich:
The HTML5 tag figcaption is supported by SimulationX 3.9 which uses IE11 for HTML rendering.

christiankral:
[...]

PR #3033, opened 2019-07-11
Change <table> to <figure> in User's Guide

christiankral:
Refs #2519

### DIFF
--- a/Modelica/package.mo
+++ b/Modelica/package.mo
@@ -694,8 +694,8 @@ Figures should in particular be included to examples to discuss the problems and
 <ol>
 <li> Figures have to be placed <strong>outside</strong> of paragraphs to be HTML compliant.</li>
 <li> Figures need to have <strong>at least</strong> a <code>src</code> and an <code>alt</code> attribute defined to be HTML compliant.</li>
-<li> Technical figures should be placed within a table environment. Each technical figure should then also have a caption. The figure caption starts with a capital letter.</li>
-<li> Illustration can be embedded without table environment.</li>
+<li> Technical figures should be placed within a figure environment. Each technical figure should then also have a caption. The figure caption starts with a capital letter.</li>
+<li> Illustration can be embedded without figure environment.</li>
 </ol>
 
 <h4>Location of files</h4>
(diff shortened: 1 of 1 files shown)


tobolar:
@HansOlsson 
> Dymola's internal rendering does not seem to fully support the new variant...

Your are right!

grafik

christiankral:
> This highlights one part that was only part of the old examples - but not stated: the placement of the caption.
> 
> Old caption in tables used align="bottom" to indicate that they should be placed beneath the image. That is also part of the examples. However, it is not documented and for figures that instead depend on the order between img and figcaption. I believe that should be made clear if that is the intent; especially since the caption was normally placed before the table-data (and thus has to be moved).

This is fixed in f6f93c11fffa2ac5e51a1a4e9c1820cb1b882fcd.
 
> I also noticed that Dymola's internal rendering does not seem to fully support the new variant; which causes a bit of problem for us.

As the release of the MSL 4.0.0 is due in March 2020, I guess there is sufficient time to fix this display issue in Dymola. @HansOlsson is this a proper approach?

beutlich:
> As the release of the MSL 4.0.0 is due in March 2020

Here's my proposal of the release schedule:

- January 2020 MSL v4.0.0 Alpha
- February 2020 MSL v4.0.0 Beta
- Early March 2020 MSL v4.0.0 RC.1
- 2020-03-25 MSL v4.0.0 Release

Modelica/package.mo
```diff
@@ -734,22 +745,20 @@ The <code>PNG</code> files should be placed in a folder which exactly represents
...
-           alt=&quot;PID_controller.png&quot;&gt;
-    &lt;/td&gt;
-  &lt;/tr&gt;
-&lt;/table&gt;
+&lt;figure&gt;
+  &lt;img src=&quot;modelica://Modelica/Resources/Images/Blocks/PID_controller.png&quot;
+       alt=&quot;PID_controller.png&quot;&gt;
+  &lt;figcaption&gt;&lt;strong&gt;Fig. 2:&lt;/strong&gt; Caption starts with a capital letter&lt;/figcaption&gt;
+&lt;/figure&gt;</pre>
 </pre>
```
dietmarw: Double `</pre>` ?

christiankral: Fixed in 57744d1

christiankral:
@HansOlsson Do you think we shall proceed with this PR?

beutlich:
I'd prefer to wait with merging this PR after https://github.com/modelica/ModelicaSpecification/issues/2395 is resolved. Thus, it only can be merged after MSL v4.0.0 (which will be based on Modelica v3.4)

Similarily, we have no SVGs in the MSL, but need to wait for https://github.com/modelica/ModelicaSpecification/issues/174.

christiankral:
OK. I understand.

beutlich (dismissed):
Suspending this PR just to avoid accidental merging.

PR #2970, opened 2019-06-13
Enhancements of PowerConverters (incl. fix for Signal2mPulse and new Voltage2DutyCycle)

AHaumer:
Replaces PR #2955, summarizing the following enhancements:
- Implements bugfix for Signal2mPulse (PR #2965 for master and #2967 for 3.2.3)
- Enhanced Voltage2DutyCycle
- VoltageToDutyCycle kept for compatibility, marked as obsolete and moved to ObsoleteModelica4
- Adapts Machines.Examples.ControlledDCDrives.Utilities.SwitchingDcDc (new Voltage2DutyCycle)
- Enhanced Examples.DCDC.{ChopperStepUp, ChopperStepDown}:
  - add inductor and smoothing capacitor
  - change default parameterization (more realistic and comparable)
  - StepDown results in 60 V -> 12 V / 1.2 A, StepUp results in 60 V -> 75 V / 1.2 A
- Implement ACAC.Triac
- Implement ACAC.Control.{VoltageToAngle, SoftStartControl}
- Implement Examples.ACAC.SoftStarter

PR #3042, opened 2019-07-13
Bugfix Machines.Examples.ControlledDCDrives.Utilities.SwitchingDcDc

AHaumer:
Changes in other sublibrary caused a bug


beutlich:
Is this really a bug or just a regression when Voltage2DutyCycle was introduced by #2970?

AHaumer:
Regression regarding Voltage2DutyCycle- thanks for merging quickly!
I checked all libraries where bugs could have occured - seems that I found the only 2.

Issue #2465, opened 2018-02-12
Modelica.Mechanics.MultiBody.Joints.Prismatic - s_offset - constant vs parameter


natehorn:
Is there any reason that s_offset is a constant rather than a parameter?  Thanks!

beutlich:
This lines
https://github.com/modelica/Modelica/blob/1de3f959ddaff0c106a9e0fcb56af61ff54b25ac/Modelica/Mechanics/MultiBody/Joints.mo#L22-L24

FrenkelJ:
By the way the Evaluate Annotation is only specified for a component declared with the prefix parameter. See Spec. 18.3 Evaluate.

beutlich:
It looks like a bug introduced by c025e058b032a9542ceba9ca0335a6a59ed6da00. Note that `s_offset` is a proper parameter (without Evaluate annotation) in the other aggregate joints with prismatic degree of freedom.

MartinOtter:
No, this is not a bug, but a feature. The reason is that I wanted to get rid of offset, to make joints simpler. For backwards compatibility reasons, "offset" was not removed but kept. Via "constant" it was, however, removed from the parameter menu. Note, a constant "phi_offset" is also present in the Revolute joint (for the same reasons).

beutlich:
- Still it would be wise to remove the Evaluate annotation for MSL 3.2.3.
- Please file a new ticket with milestone MSL_next-MINOR-version to remove the superfluous `s_offset` constant then.

tobolar:
Another option is to let `s_offset` be a parameter with `annotation (Dialog(enable=false))`. This enables a user to still see there is such a parameter with some value which influences the model behavior.

tobolar:
@beutlich 
> ... new ticket with milestone MSL_next-MINOR-version ...

Your rather mean MSL_next-MAYOR-version?

beutlich:
> Your rather mean MSL_next-MAYOR-version?

Yes, edited MSL_next-MAJOR-version above. Thanks.

PR #3048, opened 2019-07-15
Remove constants s_offset/phi_offset from 1-DoF joints

### LINKED ISSUE #2468: Modelica.Mechanics.MultiBody.Joints: remove superfluous parameters/constants
In `Modelica.Mechanics.MultiBody.Joints.Prismatic` and `Modelica.Mechanics.MultiBody.Joints.Revolute` there are superfluous parameters/constants `s_offset` and `phi_offset`, respectively.

See also #2465 for more details.

beutlich:
Close #2468.

### DIFF
--- a/ModelicaTest/MultiBody.mo
+++ b/ModelicaTest/MultiBody.mo
@@ -330,7 +330,7 @@ alt=\"model Examples.Elementary.DoublePendulum\">
             origin={50,-50},
             extent={{-10,-10},{10,10}},
             rotation=90)));
-      Modelica.Mechanics.MultiBody.Parts.FixedTranslation b3(r={1.2,0,0},
+      Modelica.Mechanics.MultiBody.Parts.FixedTranslation b3(r={1,0,0},
           animation=false) annotation (Placement(transformation(extent={{-32,-80},
                 {-12,-60}})));
       Modelica.Mechanics.MultiBody.Joints.Spherical spherical(animation=false)
@@ -453,7 +453,7 @@ for rodLength:
       Modelica.Mechanics.MultiBody.Joints.SphericalSpherical sphericalSpherical(
           computeRodLength=true, m=1) annotation (Placement(transformation(
               extent={{0,20},{-20,40}})));
-      Modelica.Mechanics.MultiBody.Parts.FixedTranslation b3(r={1.2,0,0},
+      Modelica.Mechanics.MultiBody.Parts.FixedTranslation b3(r={1,0,0},
           animation=false) annotation (Placement(transformation(extent={{-32,-80},
                 {-12,-60}})));
     equation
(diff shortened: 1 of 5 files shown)


ModelicaTest/MultiBody.mo
```diff
@@ -441,9 +439,7 @@ for rodLength:
         n={1,0,0},
         a(fixed=false),
         s(fixed=true),
-        v(fixed=false),
-        s_offset=-0.2) annotation (Placement(transformation(extent={{12,-80},{
-                32,-60}})));
+        v(fixed=false)) annotation (Placement(transformation(extent={{12,-80},{32,-60}})));
```
tobolar: Deleting `s_offset` here results in an offset in `j2.s` by -0.2 compared to reference results, which is fine. The consequence is that the reference results have to be generated again for further tests. This holds similarly for other test models.
Clipboard22

beutlich:
Yes, I also wondered why we have modified constant s_offset there, if it is legal and intentional. For me it looked like a mistake of c025e058b032a9542ceba9ca0335a6a59ed6da00, but @MartinOtter claimed in https://github.com/modelica/ModelicaStandardLibrary/issues/2465#issuecomment-365172291 that is by design. After all, I am relieved that only ModelicaTest is affected, but it may affect user models.

@tobolar Should we add FixedTranslation components to the modified test models in ModelicaTest to get reproducable results?

tobolar:
> but it may affect user models.

The referenced commit is dated on 10 Dec 2007. Which MSL Version was it? A possibility is to put a corresponding `Prismatic` joint with constant `s_offset` into "ObsoleteModelica3".

> Should we add FixedTranslation components to the modified test models in ModelicaTest to get reproducable results?

Fortunately, we just need to change position vectors of existing fixed translations. In the case of e.g. `ModelicaTest.MultiBody.FourbarVariants.SphericalAndUniversal` it is sufficient to adapt `b3(r={1.0,0,0}`.

beutlich:
> The referenced commit is dated on 10 Dec 2007. Which MSL Version was it?

It was a rather large commit for MSL 3.0.1, see

grafik

> A possibility is to put a corresponding `Prismatic` joint with constant `s_offset` into "ObsoleteModelica3".

We could add both joints to ObsoleteModelica4, sure. But we would not have an automatic conversion for them, which basically means they will remain as zombie models forever. What do you propose? I have no strong opinion. In case of moving them to ObsoleteModelica4 we also could replace the constant by parameter.

> > Should we add FixedTranslation components to the modified test models in ModelicaTest to get reproducable results?
> 
> Fortunately, we just need to change position vectors of existing fixed translations. In the case of e.g. `ModelicaTest.MultiBody.FourbarVariants.SphericalAndUniversal` it is sufficient to adapt `b3(r={1.0,0,0}`.

OK, let me see if I get it right.

ModelicaTest/MultiBody.mo
beutlich: OK, joints are moved to yield reproducible results.

PR #3049, opened 2019-07-15
resolve #2116 Sources with variable frequency

### LINKED ISSUE #2116: Sine sources with varying frequency
**Reported by clauss on 15 Nov 2016 13:23 UTC**
A. Haumer suggests to add sine sources with varying frequency into Modelica.Blocks.Sources as well as into Modelica.Electrical.Analog.Sources (both voltage and current cource). The frequency of the sources is prescribed by a real input connector.
The models are a useful improvement of the MSL, we are willing to add them to the trunk version soon.
Attached there are the models to be inspected as well as a small example circuit.
Please inspect and test the models, each hint for improving is welcome. 

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/2116

AHaumer:
This is a resolution to fix #2116 
Before doing that, I removed the "encapsulated" from 3 examples (don't know why that was used).
To implement the voltage and current source with the same icon, I moved the icons form the partial models to a new Icon package. 
Implementation is robust. Parameter startTime omitted, since this affects the initial phase shift (which can be initialized anyway) and if the sine should start later, you could keep the amlitude to zero until the sine should start.
To test the sources, there are 2 new examples.


AHaumer:
@beutlich As far as I see, all sources inherit from partial VoltageSource or CurrentSource, and these inherit from the same icons. So the icons are defined only once.
The partial VoltageSource or CurrentSource cannot bes used for the new varaible frequency sources, since the replaceable signalSource is of no use for variable frequency.

AHaumer:
Just wait a little bit with merging, 
I'll provide the signal source block, too.

AHaumer:
Just wait an hour or two, I'll provide two nice examples.

AHaumer:
@beutlich It's more a source. You may define frequency and amplitude constant, or prescribe them via inputs.

AHaumer:
@beutlich There's also only an ExpSine, not ExpCosine.
With ExpSine you achie an ExpCosine by phase = pi/2.
In SineVariableFrequencyAndAmplitude you achieve a cosine by choosing the initial value of the angle (state).
BTW, may I propose a sinc-source (sin(x))/x ? It's important when dealing with signals ...

AHaumer:
I've also improved the icon.

AHaumer:
I've added cosine with variable frequency and variable amplitude

PR #3051, opened 2019-07-16
Refs #496: Fix inconsistent partial model naming of FluxTubes

### LINKED ISSUE #496: Inconsistent partial model naming in Modelica.Magnetic

christiankral:
close #496

### DIFF
--- a/ModelicaTestConversion4.mo
+++ b/ModelicaTestConversion4.mo
@@ -902,7 +902,7 @@ Conversion test for <a href=\"https://github.com/modelica/ModelicaStandardLibrar
 <p>
 Conversion test for <a href=\"https://github.com/modelica/ModelicaStandardLibrary/issues/496\">#496</a>.
 </p>
-</html>"),           experiment(Tolerance=1e-06, StopTime = 1));
+</html>"), experiment(Tolerance=1e-06, StopTime = 1));
           end PartialTwoPortsElementary;
 
           partial model PartialTwoPorts "Partial component with magnetic potential difference between two magnetic ports p and n and magnetic flux Phi from p to n"
@@ -911,7 +911,7 @@ Conversion test for <a href=\"https://github.com/modelica/ModelicaStandardLibrar
 <p>
 Conversion test for <a href=\"https://github.com/modelica/ModelicaStandardLibrary/issues/496\">#496</a>.
 </p>
-</html>"),           experiment(Tolerance=1e-06, StopTime = 1));
+</html>"), experiment(Tolerance=1e-06, StopTime = 1));
           end PartialTwoPorts;
 
           partial model PartialFixedShape "Base class for flux tubes with fixed shape during simulation"
@@ -920,7 +920,7 @@ Conversion test for <a href=\"https://github.com/modelica/ModelicaStandardLibrar
 <p>
 Conversion test for <a href=\"https://github.com/modelica/ModelicaStandardLibrary/issues/496\">#496</a>.
 </p>
-</html>"),           experiment(Tolerance=1e-06, StopTime = 1));
+</html>"), experiment(Tolerance=1e-06, StopTime = 1));
           end PartialFixedShape;
 
           partial model PartialLeakage "Base class for leakage flux tubes with position-independent permeance and hence no force generation; mu_r=1"
@@ -929,7 +929,7 @@ Conversion test for <a href=\"https://github.com/modelica/ModelicaStandardLibrar
 <p>
 Conversion test for <a href=\"https://github.com/modelica/ModelicaStandardLibrary/issues/496\">#496</a>.
 </p>
-</html>"),           experiment(Tolerance=1e-06, StopTime = 1));
+</html>"), experiment(Tolerance=1e-06, StopTime = 1));
           end PartialLeakage;
           annotation (Documentation(info="<html>
 <p>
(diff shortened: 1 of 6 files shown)


christiankral:
> Might be good to have a common sorting everywhere. Currently it is mixed troughout the sub-packages. Proposal: Sensors, Sources, Interfaces, BaseClasses.

Makes sense to me. I will update the PR.

ModelicaTestConversion4.mo
```diff
@@ -725,35 +725,118 @@ Conversion test for <a href=\"https://github.com/modelica/ModelicaStandardLibrar
...
         extends Modelica.Icons.ExamplesPackage;
-        partial model PartialTwoPort "Two magnetic ports for graphical modeling"
+        partial model PartialTwoPort
           extends Modelica.Magnetic.FundamentalWave.Interfaces.PartialTwoPort;
           annotation (Documentation(info="<html>
 <p>
 Conversion test for <a href=\"https://github.com/modelica/ModelicaStandardLibrary/issues/496\">#496</a>.
 </p>
-</html>"));
+</html>"),         experiment(Tolerance=1e-06, StopTime = 1));
```
beutlich: No need for this experiment annotation here and further below (if this is a partial model only).

PR #3061, opened 2019-07-18
Split Thermal.{FluidHeatFlow, HeatTransfer} into single files

AHaumer:
Since for Thermal there are no open PRs at this time, I'd want to grab the opportunity and split the sublibraries into single files


AHaumer:
@beutlich what's the reason?

beutlich:
* Probably a personal aspect: I do all my editing (basically the many renaming bulk changes) in Notepad++. The less files I see the more convenient it gets w.r.t. to opened files in Notepad++.

* I assigned @dietmarw who will be happy to merge it though. 😛

beutlich:
Same holds for Search+Replace. If I want to check if the "Ah" issue is resolved in the other PR, I previously only needed to search in Batteries.mo. Now I need to check 36 files (excluding the package.order files) to see which file it actually was in. Thus in Notepad++ I can no longer do a single file search but need to do a directory search (where search hits are given as list). For me it simply is less convenient.

DrMartinKuhn:
So it is more a matter of tools. Dymola itself got quite efficient with renaming and search for usage by other components. Nevertheless, I would prefer files since changes for one model do not affect a whole sub-library. @AHaumer : I think this is a very general problem and should be agreed among the library officers.

christiankral:
@DrMartinKuhn The library officers of both `Thermal` libraries are @AHaumer and @christiankral and we are both very much in favor of splitting the sub-libraries into single files. However, we need two positive approvals in order to be able to merge to PR into the master branch.

beutlich:
> However, we need two positive approvals in order to be able to merge to PR into the master branch.

To be exact: At least two positive approvals and no negative review.

AHaumer:
Hm I've no idea how to resolve the conflicts. Seems to be the old "all-in-one" files that should be deleted, merging the directorires (with single files). Who could help me, please?

beutlich:
> Hm I've no idea how to resolve the conflicts. Seems to be the old "all-in-one" files that should be deleted, merging the directorires (with single files). Who could help me, please?

Done.

AHaumer:
@beurlich thanks a lot - just to learn: how did you do that?

PR #3066, opened 2019-07-20
Cures #3064 SaturatingInductor for 4.0.0-dev

### LINKED ISSUE #3064: Singular inconsistent scalar system for SaturatingInductor

AHaumer:
Resolves #3064 SaturatingInductor, added a test example to ModelicaTest


AHaumer:
@beutlich : Sorry for trying to avoid cherry-picks...
Fixed eps and atan.
Though Simulation-X is smart enough to master a critical situation, the implemenation should be clean.

beutlich:
@AHaumer Please only focus on this PR and consider #3067 later (after merge).

AHaumer:
@DrMartinKuhn we want to keep hysteresis out here, just focus on saturation.
Ok I'll try to imrpove documentation. 
You (or @christiankral ) could also create a PR to my PR.

christiankral:
I will update the documentation of #3067 once this PR is approved.

AHaumer:
@christiankral very nice images, many thanks! Just a minor issue: 
Fig. 2 Flux linkage the y-axis should be be labeled "Psi" instead of "Lact".
The figure caption is displayed underaneth the figure but shifted right. I assume this is a tool issue with html5?

christiankral:
> @christiankral very nice images, many thanks! Just a minor issue:
> Fig. 2 Flux linkage the y-axis should be be labeled "Psi" instead of "Lact".

Good point. This was a copy + paste error. Fixed in 2a9c9c3.
> The figure caption is displayed underaneth the figure but shifted right. I assume this is a tool issue with html5?

Yes indeed, #3030 indicated this as a tool issue. The figures render fine in OpenModelica.

christiankral:
**Note.** The script for creating the images is stored at https://gitlab.com/christiankral/ModelicaImages/tree/master/ModelicaImages/Electrical/Analog/Basic/SaturatingInductor

DrMartinKuhn:
Nice work. Just one thing: I see Fig 1 in the parameters list again. This seems to be a bug?

beutlich:
Fig 1. and Fig 2. should be placed vertically in the documentation. Should the figure captions be also left-aligned?

AHaumer:
@christiankral Thanks a lot!
@DrMartinKuhn To show a figure in the parameter dialogue is intended to support the user, not a bug.
@beutlich In Dymola2020 I see from left to right 1. figure 1, 2. caption 1, 3. figure 2, 4. caption 2. 
I believe that's a toll issue when rendering html5.

AHaumer:
@christiankral Nice utilities, splendid idea.
What tool do you use to plot the csv? Matlab / Octave or a spreadsheet like LibreOfficeCalc?

christiankral:
The figures are plot in Julia using the script [SaturatingInductor.jl](https://gitlab.com/christiankral/ModelicaImages/blob/master/ModelicaImages/Electrical/Analog/Basic/SaturatingInductor/SaturatingInductor.jl). This scripts reads the data from a CSV file written by [SaturatingInductor.mo](https://gitlab.com/christiankral/ModelicaImages/blob/master/ModelicaImages/Electrical/Analog/Basic/SaturatingInductor/SaturatingInductor.mo). 

This was the simplest solution so far, which is reproducible by everyone. This workflow could, however, be improved as it is currently based on an intermediate CSV file. I developed a Julia interface some time ago to run Modelica files in OpenModelica and to read result files in MAT format. I will make this interface public soon.

beutlich:
Looks also strange in SimulationX:

grafik

christiankral:
Is "strange" a result of the HTML code or the HTML implementation of SimulationX and Dymola?

This is what it looks like in OpenModelica:

image

AHaumer:
@christiankral 
Just a guess: It's a tool issue, i.e. rendering of html5, not a html5 issue, since it's rendered properly by OM.

AHaumer:
@beutlich I think all change requests have been addressed. Please merge to master and cherry pick to maint/3.2.3,

beutlich:
In my opinion we neeed to wait for the two documentation issues #2519 (HTML5 `figcaption` tag) and #3033 (`figure` env) to be clarified/decided first.

christiankral:
> Documentation issues need to be clarified first before we can merge it, especially when back-porting to MSL v3.2.3.

On my opinion the documentation is improved and sufficient by now. All parameters, the characteristics of the implementation and two figures are added. I guess there is nothing left to do. @dietmarw and myself approved the PR.

beutlich:
My concern is not the documentation itself, but the new HTML features where we still have undecided discussions going on. If these discussions are resolved, we can merge this PR (and not vice versa). Additionally it might be good to have a second tool that properly displays the new documentation.

christiankral:
OK, I see your point. This is actually a matter of #2519, #3030 and https://github.com/modelica/ModelicaSpecification/issues/2395

dietmarw:
My reasoning is that MSL 4.0.0 and a new 3.2.3  maintenance release would be shipped with a new tool. Now if a tool in 2020 still does not support HTML 5 which by then would be more than 5 years old that truly this should be the tools problem and not "ours", right? Especially since it is not that the documentation is not readable but just does not look 100% "nice".

 E.g., browsers have supported figcaption as far back as Firefox 4, IE 9, Chrome 4. So making use of these features might have the positive influence on tool vendors to actually do something about the documentation rendering and get it into the 2020s.

beutlich:
I created a [test class](https://gist.github.com/beutlich/0dd78940243c41a6dc39c37f90ea6749) to test HTML features in different Modelica environments.

* In Dymola: `figcaption` is supported by the external Info window, but not in the internal Documentation view panel.

grafik

* In SimulationX: `figcaption` is neither supported by the Help window nor the Documentation view window, since it is displayed right to the image.

grafik

PR #3074, opened 2019-07-27
Declare variable t_start discrete in Modelica.StateGraph.Interfaces.PartialTransition

beutlich:
Avoid warning in Wolfram SystemModeler.

### DIFF
--- a/Modelica/StateGraph.mo
+++ b/Modelica/StateGraph.mo
@@ -2017,7 +2017,7 @@ package Interfaces "Connectors and partial models"
         "Vector of transition output connectors"
       annotation (Placement(transformation(extent={{10,-5},{20,5}})));
     protected
-    Modelica.SIunits.Time t_start
+    discrete Modelica.SIunits.Time t_start
         "Time instant at which the transition would fire, if waitTime would be zero";
     Real t_dummy;
   initial equation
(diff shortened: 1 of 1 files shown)


henrikt-ma:
Why not initialize to `time` rather than `0`?

beutlich:
> 
> 
> Why not initialize to `time` rather than `0`?

I tossed a coin, since not relevant anyway. But will do.

henrikt-ma:
I note that this model seems to have the interesting problem of sometimes dealing with a discrete variable that doesn't change.  I suppose it would also be possible to express this more clearly as:
```
initial equation
  pre(t_start) = time;
…
if enableTimer then
  …
else
  when false then
    t_start = time; /* Dummy equation for t_start. */
  end when;
```
Or is there some detail about initializing {{pre(t_start)}} vs {{t_start}} that I'm missing?

beutlich:
LGTM. I applied your proposed changes.

henrikt-ma:
It just occurred to me that the canonical way of providing a dummy equation for a `discrete` variable should probably be this:
```
    when false then
      t_start = pre(t_start); /* Dummy equation for t_start */
    end when;
```
If this would be recognized as just that, the comment wouldn't even be necessary.

Modelica/StateGraph.mo
```diff
@@ -2040,7 +2038,9 @@ package Interfaces "Connectors and partial models"
       t = if enableFire then t_dummy else 0;
       fire = enableFire and time >= t_start + waitTime;
     else
-      t_start = 0;
+      when false then
+        t_start = time; /* Dummy equation for t_start */
```
beutlich:
```suggestion
        t_start = pre(t_start);
```

beutlich:
@HansOlsson OK for you to merge?

PR #3077, opened 2019-07-29
Keep VS2005 support for C-Sources

beutlich:
VS2005 was broken with some recent commits.


thorade:
I also do not have 2005 installed. Could install it and try to compile; this is the only review I can contribute

beutlich:
VS2005 Express won't compile 64-bit binaries. Therefore I recently purchased (on behalf of the MA) a VS2005 Standard edition.

We still have the VS2005 build projects in the MSL and I can maintain them now.

Even if Dymola does not support VS 2005 as compiler, it still can link to the VS2005 compiled static libraries in a newer VS (provided it is pure C).

PR #3083, opened 2019-08-09
Add second derivatives to table blocks

### LINKED ISSUE #2901: Provide higher order derivatives for CombiTables
It would be appreciated if higher order derivatives are provided in the standard tables.

beutlich:
Close #2901.


beutlich:
> 
> 
> Would it be possible to add a test?

I used https://github.com/modelica/ModelicaStandardLibrary/issues/2901#issuecomment-489246504 as a test for CombiTimeTable after replaing the approximated second derivative by the symbolic second derivative.

beutlich:
> Would it be possible to add a test?

8af9758ab updates all test models.

PR #3084, opened 2019-08-09
Fix HTML tags in ModelicaReference.ModelicaGrammar

beutlich:


svorkoetter:
I can update the code that generated this in the first place, so the issue won't resurface.

PR #3085, opened 2019-08-09
Fix compilation of ModelicaMatIO.c if HAVE_ZLIB=0 is defined

beutlich:
If `HAVE_ZLIB=0` is defined, no zlib will be used.


beutlich:
And similar for HAVE_HDF5, right?

PR #3088, opened 2019-08-10
Add LinearDependency2 and deprecate LinearDependency

### LINKED ISSUE #1724: Modelica.Blocks.Math.LinearDependency has strange formulation

beutlich:
This is my second attempt to fix #1724.

### DIFF
--- a/Modelica/Blocks/Math.mo
+++ b/Modelica/Blocks/Math.mo
@@ -3607,13 +3607,13 @@ provide both values as output.
 </html>"));
   end MinMax;
 
-  block LinearDependency "Output a linear combination of the two inputs"
+  block LinearDependency2 "Output a linear combination of the two inputs"
     extends Modelica.Blocks.Interfaces.SI2SO;
     parameter Real y0=0 "Initial value";
     parameter Real k1=0 "Gain of u1";
     parameter Real k2=0 "Gain of u2";
   equation
-    y = y0*(1 + k1*u1 + k2*u2);
+    y = y0 + k1*u1 + k2*u2;
     annotation (Icon(coordinateSystem(preserveAspectRatio=false, extent={{-100,
               -100},{100,100}}), graphics={
           Line(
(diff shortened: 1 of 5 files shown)


JoeRiel:
I'm not real familiar with convertModifiers, but would the following work?
```convertModifiers("Modelica.Blocks.LinearDependency",
                 {"y0=0", "k1=0", "k2=0"}, {"y0=%y0%","k1=%y0%*%k1%","k2=%y0%*%k2%"});```
If so, that would allow directly replacing the model, hence no need to create a new name.

**Comment by beutlich** [comment]
> I'm not real familiar with convertModifiers, but would the following work?

Need to check it.

**Comment by beutlich** [comment]
> I'm not real familiar with convertModifiers, but would the following work?

Unfortunately, the following conversion fails in Dymola 2020, since `y0` cannot be evaluated.

```
convertModifiers("Modelica.Blocks.Math.LinearDependency",
                 {"k1"}, {"k1=%y0%*%k1%"});
convertModifiers("Modelica.Blocks.Math.LinearDependency",
                 {"k2"}, {"k2=%y0%*%k2%"});
```

The following conversion works in Dymola 2020, but unfortunately is not save w.r.t. the lookup of `y0`, since we do not have a self/this environment in Modelica (see https://github.com/modelica/ModelicaSpecification/issues/213).

```
convertModifiers("Modelica.Blocks.Math.LinearDependency",
                 {"k1"}, {"k1=y0*%k1%"});
convertModifiers("Modelica.Blocks.Math.LinearDependency",
                 {"k2"}, {"k2=y0*%k2%"});
```

Thus, currently complex modifiers cannot be converted in Dymola 2020.

beutlich:
Superseded by #3095.

PR #3093, opened 2019-08-14
maint/3.2.3: Fix uninitialized memory in ModelicaInternal_readLine, ModelicaStrings_scanIdentifier and ModelicaStrings_scanString

beutlich:
Back-port #3092 for maint/3.2.3.


beutlich:
> I noticed that one line still assumes that ModelicaAllocateString initializes the string

That one is already dealt with by #3068.

PR #3097, opened 2019-08-18
Remove Interfaces.ConditionalHeatPort from Modelica.Electrical.Analog and Modelica.Magnetic.FluxTubes

### LINKED ISSUE #359: Move conditional heat port models to Modelica.Thermal.HeatTransfer.Interfaces
**Reported by christiankral on 21 Apr 2010 15:44 UTC**
The conditional heat port models `Modelica.Electrical.Analog.Interfaces.ConditionalHeatPort` and `Modelica.Electrical.MultiPhase.Interfaces.ConditionalHeatPort` are not specific for electrical systems. Since these conditional heat port models can be applied to any model with one or more conditional heatports, the location in the electrical domain is not correct. Therefore, both conditional heat port models shall be moved to `Modelica.Thermal.HeatTransfer.Interfaces` when the next conversion script is due.

**Example**: Conditional heat port models shall also be applied in `Modelica.Magnetic`. 

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/359

beutlich:
Fix #359. 

@christiankral Please double-check Modelica.Electrical.Machines.BasicMachines.Components.DamperCage which was the only non-trivial conversion since it has both lossPower and LossPower declared.

Would also be good if one of the reviewers could run "Check model" in Dymola.


beutlich:
This is getting worse: 

* Unfortunately the connector position of the heatport is different in Modelica.Electrical.Analog.Interface.ConditionalHeatPort and Modelica.Thermal.HeatTransfer.Interfaces.PartialElementaryConditionalHeatPort.
* Unfortunately we cannot move the connector position after inheritance (see https://github.com/modelica/ModelicaSpecification/issues/377).
* Unfortunately, when using the IconMap/DiagramMap annotation of the [extends clause](https://specification.modelica.org/master/Ch18.html#extends-clause), I get three different results in Dymola, SimulationX and Wolfram SystemModeler.

1. Dymola requires 

```mo
extends Modelica.Thermal.HeatTransfer.Interfaces.PartialElementaryConditionalHeatPort
  annotation(IconMap(extent={{0,-100}, {200, 100}}), DiagramMap(extent={{0,-100}, {200, 100}}));
```
  
to display the heatport at the correct position for backward compatibility with Modelica.Electrical.Analog.Interface.ConditionalHeatPort. However also the connectors and connection lines from other base classes are moved now. Probably a tool issue @HansOlsson ?

2. SimulationX requires

```mo
extends Modelica.Thermal.HeatTransfer.Interfaces.PartialElementaryConditionalHeatPort
  annotation(IconMap(extent={{-200,-100}, {0, 100}}), DiagramMap(extent={{-200,-100}, {0, 100}}));
```

which is the opposite direction compared to Dymola. Is this a tool or specification issue @ooer?

3. Wolfram SystemModeler does not move the inherited heatport connector at all.

beutlich:
I took the time and crreated a [test model](https://gist.github.com/beutlich/0ebc141f6ed13a2d9b405776abecd855) comparing the four different implementations using a simple Resistor as an example. Compare the screen shots from Dymola and SimulationX:

1. Resistor1 is the original implementation of MSL v3.2.3 and is the reference.
grafik 
grafik
2. Resistor2 no longer extends from Modelica.Electrical.Analog.Interfaces.ConditionalHeatPort but from Modelica.Thermal.HeatTransfer.Interfaces.PartialElementaryConditionalHeatPort.
grafik 
grafik
3. Resistor 3 also sets the IconMap/DiagramMap annotation such that the heatPort is moved to the reference poistion (in Dymola). However, in Dymola also the blue electrical connection lines are modified as well.
grafik 
grafik
4. Resistor 4 also sets the IconMap/DiagramMap annotation such that the heatPort is moved to the reference poistion (in SimulationX).
grafik 
grafik

@HansOlsson @ooer Can you please confirm the tool issue or specification issue. Thank you.

christiankral:
I wonder if it is save to remove `Modelica.Electrical.Analog.Interfaces.PartialConditionalHeatPort` from the MSL which is still in. There is just `Modelica.Electrical.Analog.Batteries.BatteryOCV_SOCtable` relying on it. I guess it is no problem to keep the broken `Analog.Batteries` package, as it will be replaced by #3054 `Electrical.Batteries`.

christiankral:
@beutlich shall we remove the keyword `Partial` from the following models?
- `Modelica.Thermal.HeatTransfer.Interfaces.PartialElementaryConditionalHeatPort`
- `Modelica.Thermal.HeatTransfer.Interfaces.PartialElementaryConditionalHeatPortWithoutT`
- `Modelica.Thermal.HeatTransfer.Interfaces.PartialConditionalHeatPort`

christiankral:
@beutlich Even though the partial model `Modelica.Thermal.HeatTransfer.Interfaces.PartialConditionalHeatPort` is only used in polyphase models by now, but it may be used in any application which uses a vector of heat ports. It think it is therefore better to move this partial model to `Modelica.Thermal.HeatTransfer.Interfaces` as well. A possible naming of the vector heat port model (and the derivatives `Elementary...` and `...WithoutT` for full compatibility) is:

- `ElementaryConditionalVectorHeatPort`
- `ElementaryConditionalVectorHeatPortWithoutT`
- `ConditionalVectorHeatPort`

beutlich:
> 
> 
> @beutlich shall we remove the keyword `Partial` from the following models?
> 
>     * `Modelica.Thermal.HeatTransfer.Interfaces.PartialElementaryConditionalHeatPort`
> 
>     * `Modelica.Thermal.HeatTransfer.Interfaces.PartialElementaryConditionalHeatPortWithoutT`
> 
>     * `Modelica.Thermal.HeatTransfer.Interfaces.PartialConditionalHeatPort`

OK from my side, but in a seprate PR please.

beutlich:
> 
> 
> @beutlich Even though the partial model `Modelica.Thermal.HeatTransfer.Interfaces.PartialConditionalHeatPort` is only used in polyphase models by now, but it may be used in any application which uses a vector of heat ports. It think it is therefore better to move this partial model to `Modelica.Thermal.HeatTransfer.Interfaces` as well. A possible naming of the vector heat port model (and the derivatives `Elementary...` and `...WithoutT` for full compatibility) is:
> 
>     * `ElementaryConditionalVectorHeatPort`
> 
>     * `ElementaryConditionalVectorHeatPortWithoutT`
> 
>     * `ConditionalVectorHeatPort`

OK from my side, but in a separate PR please. What about ElementaryConditionalHeatPort**V**?

beutlich:
> 
> 
> @beutlich I wonder if it is save to remove `Modelica.Electrical.Analog.Interfaces.PartialConditionalHeatPort` from the MSL which is still in. There is just `Modelica.Electrical.Analog.Batteries.BatteryOCV_SOCtable` relying on it. I guess it is no problem to keep the broken `Analog.Batteries` package, as it will be replaced by #3054 `Electrical.Batteries`.

If so, we should move it to  ObsoleteModelica4.mo like I did with the other replaced classes.

beutlich:
> This is getting worse:
> 
>     * Unfortunately the connector position of the heatport is different in Modelica.Electrical.Analog.Interface.ConditionalHeatPort and Modelica.Thermal.HeatTransfer.Interfaces.PartialElementaryConditionalHeatPort.
> 
>     * Unfortunately we cannot move the connector position after inheritance (see [modelica/ModelicaSpecification#377](https://github.com/modelica/ModelicaSpecification/issues/377)).
> 
>     * Unfortunately, when using the IconMap/DiagramMap annotation of the [extends clause](https://specification.modelica.org/master/Ch18.html#extends-clause), I get three different results in Dymola, SimulationX and Wolfram SystemModeler.

@christiankral In case you missed my https://github.com/modelica/ModelicaStandardLibrary/pull/3097#issuecomment-522327170, I doubt we can merge this PR without a break in backward compatibility of the heatport position.

beutlich:
Closing, since impossible to fix for MSL v4.0.0.

Issue #712, opened 2017-01-14
Revise/revisit all SI-units listed in Modelica.SIunits


modelica-trac-importer:
**Reported by dietmarw on 12 Mar 2012 12:10 UTC**
The current version of `Modelica.SIunits` is based on quote

> Library of type and unit definitions based on SI units according to ISO 31-1992

The problem is that this ISO norm is deprecated by [ ISO 80000-1:2009](http://www.iso.org/iso/iso_catalogue/catalogue_ics/catalogue_detail_ics.htm?csnumber=30669). 

In addition to that has the current `Modelica.SIunits` been steadily "enhanced" by other library officers with unit names that are not defined in the ISO norm but where just _convenient_ to use. There should be a proper rule on what can be added and what not.

Related tickets: #545 (Molar units missing), #688 (Add Complex Units (Magnetic))

I'm scheduling this already for [milestone:MSL3.2.1] so that we can evaluate the new units that have been added recently. The main overhaul should probably be done in connection with [milestone:MSL4.0]

Related tickets: #545 (Molar units missing), #688 (Add Complex Units (Magnetic))

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/712

modelica-trac-importer:
**Comment by hansolsson on 17 Apr 2012 13:24 UTC**
To me it seems that if Modelica.SIunits is not based on the SI standard we should consider using another name. 

I know that one can argue that the other units are based on the SI-base units, but the same should be true for all models in the entire MSL, except for:
```
Modelica.SIunits.Conversions.NonSIunits
```
which does not seem like the most intuitive name.

The new standard is possibly not that important, I would assume it is just a slight revision with possibly some changed names etc; i.e. not a change with major impact.

modelica-trac-importer:
**Comment by wischhusen on 17 Apr 2012 14:19 UTC**
Perhaps one could stay with Modelica.SIunits but open another subpackage Modelica.SIunits.Derived to account for units that were derived from the standard unit system with multipliers. I think, it is useful to add somewhere in the Modelica Standard Library really commonly used units for many reasons (e. g. there is still no length unit in feet ... ). I know that this can (and is) already be done with local settings on each single machine but is that reasonable?

So what about that proposal:

Create a new subpackage Modelica.SIunits.Derived and put all those types inside which are really used by scientist and engineers :-).

modelica-trac-importer:
[...]

PR #3100, opened 2019-08-26
Add translational and rotational jerk to SIunits

### LINKED ISSUE #1988: Missing SIunits
**Reported by ahaumer on 21 Apr 2016 14:14 UTC**
In Modelica.SIunits both rotational and translational jerk is missing, i.e. the derivative of acceleration, unit = m/s3 resp. rad/s3.

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/1988

christiankral:
Close #1988.

### DIFF
--- a/Modelica/SIunits.mo
+++ b/Modelica/SIunits.mo
@@ -893,10 +893,11 @@ argument):</p>
   type AngularVelocity = Real (
       final quantity="AngularVelocity",
       final unit="rad/s");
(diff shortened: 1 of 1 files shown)


dietmarw:
So is this an SI unit or not? Maybe better wait with this until #712 is clarified.

beutlich:
> 
> 
> So is this an SI unit or not? Maybe better wait with this until #712 is clarified.

Well, it is not different from similar quantities velocity or acceleration and can be treated independently from #712.

Issue #3096, opened 2019-08-18
Invalid reference to T_heatPort in line models


beutlich:
The three line models Modelica.Electrical.Analog.Lines.{OLine, M_OLine, ULine} mention `T_heatPort` in their parameter descriptions and documentation though there is no such variable.

PR #3101, opened 2019-08-26
Refs #3096: Fix undocumented variable T_heatPort

christiankral:
In order to be more precise, T_heatPort is replaced by T_actual and T_actual is explained in the text, as this is mainly a documentation and not a code issue


dietmarw:
`T_actual` is not present in these models so no point in referencing them. See #3138 for an improved PR. Sorry I did not see this PR when creating that one but would suggest to continue with #3138

Issue #545, opened 2017-01-14
Molar units missing


modelica-trac-importer:
**Reported by @wischhusen on 2 May 2011 11:21 UTC**
SIunits require also molar specific enthalpy:

```
type MolarEnthalpy = Real (final quantity="MolarEnthalpy", final unit=
        "J/mol", nominal=5e4);
```

Moreover, a very popular Non-SIunit for molar mass is required e.g. in Modelica.Fluid.Dissipation.HeatTransfer.StraightPipe.kc_twoPhase_overall_KC (see ticket #502):

```
type MolarMass_gpmol = Real (final quantity="MolarMass", final unit=
          "g/mol", min=0, nominal=1) "Molar mass in gramm per mol";
```
----

Migrated-From: https://trac.modelica.org/Modelica/ticket/545

modelica-trac-importer:
**Comment by dietmarw on 2 May 2011 11:55 UTC**
Replying to [ticket:545 wischhusen]:
> Moreover, a very popular Non-SIunit for molar mass is required e.g. in Modelica.Fluid.Dissipation.HeatTransfer.StraightPipe.kc_twoPhase_overall_KC (see ticket #502):
> 
> ```
> type MolarMass_gpmol = Real (final quantity="MolarMass", final unit=
>           "g/mol", min=0, nominal=1) "Molar mass in gramm per mol";
> ```

Instead of introducing yet another non-SI unit which is simply a multiple of an existing unit I would rather suggest the use of `displayUnit`. E.g.,
```
parameter Modelica.SIunits.MolarMass GrammPerMol(displayUnit="g/mol");
```
This requires of course tool support (e.g., in Dymola you can add such conversions via script:
 ```
defineUnitConversion("kg/mol", "g/mol", 1e3, 0);
```
)

modelica-trac-importer:
**Comment by thorade on 14 Sep 2011 08:43 UTC**
I agree that it would be nice to have displayUnit g/mol for MolarMass.

Also it would be nice to have displayUnit mol/l for Modelica.SIunits.Concentration
The SI default unit for concentration is mol/m³ but mol/l is also frequently used.
Another molar quantity that would be nice to have is molar density (the inverse of molarVolume). It should have the same dimension as Concentration so it can be defined as an Alias of Concentration.

For Viscosity it would be nice to add displayUnit µPa·s
Default unit is Pa·s but that gives very low values.

modelica-trac-importer:
[...]

PR #3114, opened 2019-09-09
Add Molality/MolalConcentration and Molarity/MolarConcentrationto to SIunits

beutlich:
Note, that this changes the quanitity name for Modelica.SIunits.IonicStrength from "IonicStrength" to "Molality."

See #545.


beutlich:
> Looks very nice, except MolarConcentration=Molality :

Sorry, I misread it.

PR #3069, opened 2019-07-24
maint/3.2.3: Fix Media.Air.MoistAir.{density_derX, s_pTX, s_pTX_der}

beutlich:
Back-port of #2932 to fix #2874, #2872 and #2868 in maint/3.2.3 branch.

PR #3118, opened 2019-09-10
Bug in pressure influence in moist air function s_pTX and s_pTX_der

wischhusen:
See https://github.com/modelica/ModelicaStandardLibrary/pull/3069#issuecomment-529359817 for the issue analysis.

### DIFF
--- a/Modelica/Media/Air/MoistAir.mo
+++ b/Modelica/Media/Air/MoistAir.mo
@@ -1365,17 +1365,10 @@ The <a href=\"modelica://Modelica.Media.Air.MoistAir.ThermodynamicState\">thermo
       "Molar fraction";
 
   algorithm
-    s := Modelica.Media.IdealGases.Common.Functions.s0_Tlow(dryair, T)*(1 - X[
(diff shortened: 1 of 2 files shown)


dietmarw:
This PR is messed up. Please try again without deleting half of the MSL ;-)

beutlich:
I tried my best. Please double check.

beutlich:
@wischhusen  Can you please confirm the force-pushed changes!
@hubertus65 @thorade Can you please review!

thorade:
Just by looking at the code I do not immediately see what has changed or not, also partly due to formatting, but I assume it is correct. Do we have a minimal example to reproduce the problem that will be fixed by this?

wischhusen:
@thorade This should be the tester.

[testerMoistAir.zip](https://github.com/modelica/ModelicaStandardLibrary/files/3641215/testerMoistAir.zip)

beutlich:
@hubertus65 @thorade Can you please add your review.

beutlich:
Edit: Updated my previous comment (which was based on wrong simulation time).

Issue #1122, opened 2017-01-14
Adding yet another Complex record seems wrong


modelica-trac-importer:
**Reported by sjoelund&#x2e;se on 24 May 2013 06:41 UTC**
eeee36f4ba230b7153ddb19183346f79bacfe715 introduced a new record Modelica.Media...MyRecord, which is not only wrong but seems to be a weird design choice.
First of all, import MyRecord means the record's name has to be .MyRecord, but it does not exist on the top level, so the code is wrong.
And we anyway have operator overloading and a Complex class. So why would we introduce yet another class that does the same thing?
If this is due to some tool not supporting operator overloading, `[Note, the operator overloading as defined in this section is only a short hand notation for function calls.]`, which should mean you can use the operator record without implementing operator overloading, i.e.:
```mo
model M
  Complex compl1,compl2;
equation 
  compl1 = Complex.'constructor'(1,1);
  compl2 = Complex.'*'(compl1,compl1);
end M;
```

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/1122

modelica-trac-importer:
**Comment by otter on 24 May 2013 07:14 UTC**
Replying to [ticket:1122 sjoelund.se]:

> eeee36f4ba230b7153ddb19183346f79bacfe715 introduced a new record Modelica.Media...MyRecord, which is not only wrong

Sorry, there was a wrong import statement that I have just corrected in f8d78bf07cb3dc2939e13acb647eb884ab56a437.

> but seems to be a weird design choice.
> First of all, import MyRecord means the record's name has to be .MyRecord, but it does not exist on the top level, so the code is wrong.
> And we anyway have operator overloading and a Complex class. So why would we introduce yet another class that does the same thing?
> If this is due to some tool not supporting operator overloading, `[Note, the operator overloading as defined in this section is only a short hand notation for function calls.]`, which should mean you can use the operator record without implementing operator overloading, i.e.:
> ```mo
> model M
> Complex compl1,compl2;
> equation 
> compl1 = Complex.'constructor'(1,1);
> compl2 = Complex.'\*'(compl1,compl1);
> end M;
> ```

[...]

PR #3131, opened 2019-10-02
Replace Modelica.Media.Air.ReferenceMoistAir.Utilities.Ice09_Utilities.Basic.{MyComplex, MyComplexF} by operator record Complex

### LINKED ISSUE #2304: MyComplex vs. Complex
Left over from #1122: 
> But why not use Complex directly like the other packages already do?

Utilize the record Complex (and defined operators) instead of defining own record Modelica.Media.Air.ReferenceMoistAir.Utilities.Ice09_Utilities.Basic.MyComplex and functions MyComplexF.* to avoid redundancy.

Note, that there is Modelica.Media.Air.ReferenceMoistAir.Utilities.Ice09_Utilities.Basic.MyComplexF.'log' but no equivalent function in operator record Complex.

beutlich:
Closes #2304.

### DIFF
--- a/Modelica/Media/Air/ReferenceMoistAir.mo
+++ b/Modelica/Media/Air/ReferenceMoistAir.mo
@@ -2442,111 +2442,6 @@ package ReferenceMoistAir
           Modelica.SIunits.AbsolutePressure p0;
         end IceConstants;
 
-      protected
-        record MyComplex "Complex number with function"
-          extends Modelica.Icons.Record;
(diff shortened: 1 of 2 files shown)


beutlich:
Yes, see https://github.com/modelica/ModelicaStandardLibrary/issues/1122#issuecomment-272635099 from 2013-05-24:

> Hubertus suggested to not use Complex from MSL for that, because then tools not yet supporting operator overloading cannot handle Media.

Issue #2397, opened 2017-11-25
A better Multiphase power sensor


ceraolo:
The power sensor in the Electrical.Analog.MultiPhase sensor uses many variables and equations and requires adding a star and its ground.
Indeed because of Aron's theorem, power measure can be made in a simpler way, which uses much less variables and allows cleaner diagrams.
I enclose a package in which I propose the new AronSensor and two example models.
These two example models do exactly the same thing, but AronPower is simpler. Moreover, StdPower has 210 equations, while AronPow has 121.
Maybe AronSensor can be added to MSL?

[Pow3Sensors.zip](https://github.com/modelica/Modelica/files/1502944/Pow3Sensors.zip)

AHaumer:
To be correct;
This isn't the Aron connection we're used from threephase systems which uses 2 power meters instead of 3. https://de.wikipedia.org/wiki/Aronschaltung#/media/File:Leistung-AronschaltungWB.svg
The upper image shows active power measurement, the lower image shows reactive power measurement in a threephase system.
Your "AronSensor" calculates power in all phases and relies on the fact that either source or load are grounded. Using a power sensor in a real system means you have to connect not only the current branch but also the correct voltage branch. 
Regarding the count of equations, the difference can be found only looking at the alias variables i.e. trivial equations.
In your example, you can delete star1 and ground1 and connect powersensor.nv to either star.plug_p or star2.plug_p. So you just save 2 connections (to powerSensor.pv and powerSensor.nv).

ceraolo:
Thank you for considering this post. AHaumer.
Indeed I agree that the name "AronSensor" at least needs some explanation, and could be even changed, if that sensor (or a modified version of it) is adopted.

I did not comment on the name in my original post, because I first wanted to see if it stimulated discussion or not.
In the beginning I choose the name "AronSensor" because the icon of the proposed model on a diagram is like "as a set of Aron wattmeters": along the line, as  in the pictures from wikipedia you linked.
But that Aron connection gives correct results only for "balanced" three-wire three-phase systems (null sum of currents). So I resorted to the generalised formulation of Aron's theorem, that states that if a circuit is composed by two parts, connected through m+1 wires, the total power absorbed by one part is given by the sum of products of m voltages and  m currents, the voltages being taken with reference to the (m+1)-th wire (for which current is not measured).
So my "AronSensor" is an Aron Sensor in two ways:
[...]

Issue #2990, opened 2019-06-22
Missing User's Guides


christiankral:
I suppose that each Modelica package shall have a User's Guide

* to give a short introduction into the package,
* to host contact information,
* to host the revision history,
* and summarize all literature references. 

Such a User's Guide is currently missing in the following packages:

* Blocks
* Electrical.Analog
* Thermal.HeatTransfer
* Math
* ComplexMath
* Icons already provides the information does not host it in a subpackage User's Guide
![image](https://user-images.githubusercontent.com/4184218/59961739-0b9c0100-94dc-11e9-9378-668f799e2e87.png)

I guess we shall add User's Guides to these subpackags.

AHaumer:
I support this idea, But I don't understand:
"Icons already provides the information does not host it in a subpackage User's Guide"
I suppose those parts should be included in a UsersGuide?

One more question:
Should the parts of a UsersGuide be stord as one file or in a diretcory strudcture as single files?
Personally, I prefer the second one. 
I strongly dislike the style that UsersGuide and/or Examples are included in package.mo (like in Blocks)!

christiankral:
> I support this idea, But I don't understand:
> "Icons already provides the information does not host it in a subpackage User's Guide"
> I suppose those parts should be included in a UsersGuide?

My point is: every sub-library has a `UsersGuide` which contains at least `Contact` and `ReleaseNotes`. The `UsersGuide` explains what the sub-library is good for, `Contact` contains the library officers and their contact information, `Release Notes` contain...

We do not have such a structure in the `Icons` sub-library. The point of discussion is: Shall we also include a real `UsersGuide` -- not only the icons -- in the Icons sub-library. After looking at this case again I came to the conclusions that "no" is probably the better answer, as the library is better to understand if it is fully flat organized (as Constants and SIunits). The relevant information is hosted in the Documentation layer of `Icons`, so that shall be OK.

beutlich:
I'd prefer to have UsersGuide.mo as single file. For examples it might be preferrable to have each model as extra file.

christiankral:
User Guides are still missing for: 

- [ ] `Modelica.Blocks`
- [ ] `Modelica.Math`
- [ ] `Modelica.ComplexMath`

arunkumar-narasimhan:
As pointed out by @HansOlsson in the comment on issue #4247, it would be worth to discuss here if adding release notes to sub-libraries is necessary.

PR #3134, opened 2019-10-06
Update info of AronSensor.mo

max-privato:
See comment dated 5 Oct 2019 and 6 Oct 2019 on ticket #2397

### DIFF
--- a/Modelica/Electrical/Polyphase/Sensors/AronSensor.mo
+++ b/Modelica/Electrical/Polyphase/Sensors/AronSensor.mo
@@ -75,9 +75,10 @@ Contains two
 to measure total active power in a three-phase system.
 </p>
 <p>
-This device works only for sinusoidal voltages and currents in three-phase systems without neutral.
+This device works only in three-phase systems without neutral.
 </p>
 <p>
+Massimo Ceraolo: Electric Power Systems - University of PISA <br>
 Elmar Schr&uuml;fer, Leonhard Reindl, Bernhard Zagar: Elektrische Messtechnik. Carl Hanser Verlag.<br>
 Reinhard Lerch: Elektrische Messtechnik. Springer Vieweg.
 </p>


Modelica/Electrical/Polyphase/Sensors/AronSensor.mo
```diff
@@ -75,9 +75,10 @@ Contains two
 to measure total active power in a three-phase system.
 </p>
 <p>
-This device works only for sinusoidal voltages and currents in three-phase systems without neutral.
+This device works only in three-phase systems without neutral.
```
christiankral: This change is perfectly OK as sinusoidal voltages and currents are not an assumption. Solely the neutral must not be connected.

Modelica/Electrical/Polyphase/Sensors/AronSensor.mo
```diff
@@ -75,9 +75,10 @@ Contains two
...
 </p>
 <p>
+Massimo Ceraolo: Electric Power Systems - University of PISA <br>
 Elmar Schr&uuml;fer, Leonhard Reindl, Bernhard Zagar: Elektrische Messtechnik. Carl Hanser Verlag.<br>
 Reinhard Lerch: Elektrische Messtechnik. Springer Vieweg.
```
christiankral: I am unhappy with this literature, as these reference are not in line with the guidelines of the Modelica Standard Library, see `Modelica.UsersGuide.Conventions.Documentation.Format.References`. As we are lacking a User's Guide in the `Analog` package, which shall summarize all references, we are more and more getting inconsistent here.

beutlich:
>     2\. I will create a User's Guide for the `Analog` package

Make sure to address #2990. Thanks.

Modelica/Electrical/Polyphase/Sensors/AronSensor.mo
max-privato: Just to be sure I understand well: the latter comment of yours means that you agree with my correction without me to have to search for a reference?

The Aron connection for me is a straightforward consequence of the Tellegen's theorem, 
which just requires that a circuit can be formulated as the composition of two subcircuits connected to each other through three wires. No need to consider sinusoidal shapes.

In case you are curious or deem it useful I can add details of this.

christiankral: Yes, I fully agree with fixing line 78 and your arguments as the original documentation is wrong.

Modelica/Electrical/Polyphase/Sensors/AronSensor.mo
christiankral: On line 81 I want you to add a http(s) link to the original PDF document or the publisher where the document is available so that one does not have to search on the internet for the cited literature. I will also add the links for the two references in line 82-83.

max-privato: Line 81 is totally wrong. My mistake. 
In fact, I did not even understand that the bottom lines should be references (probably because the two existing ones they are incomplete; the German language did not help).

Personally I'm not sure the Aron connection really requires reference(s). The very name should recall users the meaning. 
I also don't like references in languages different from English, for general topics for which there exists large enough English literature.

What I propose now is one of the following options:
1. remove all references and don't substitute them with anything (i.e. stop the documentation with the word "neutral.")
2. remove all references and substitute them with a unique reference in English language, possibly openly available to anyone on the Internet
3. remove all references and substitute them with some formulas and a picture that, starting from Tellegen's theorem, illustrates the reasons why Aron connection works not only "in three-phase systems without neutral" as is already mentioned in the documentation for easy comprehension, but, more in general "in all circuits composed by two subcircuits connected to each other by three wires, typically three-phase systems without neutral".

I personally propose option 1, but in case there is agreement on option 3, I could prepare it myself and create a different PR.

christiankral: Different Modelica users have a different level of knowledge on electrical engineering so I suppose it is good to have some explanation or reference -- either one is fine with me. I agree that it is a lot better to have an English reference instead of a German one. However, if literature references are used, the models shall only refer to the reference by reference key which points to `References`. All `References` with are summarized the `User's Guide` (see `Modelica.UsersGuide.Conventions.Documentation.Format.References`) and shall have a http(s) link to the real book, paper, thesis, etc.

I am therefore in favor of option 2 and 3. I think we should also include @AHaumer's opinion here before creating an new PR. 

@max-privato thanks for taking care of this issue.

max-privato: Users of Electrical|MultiPhase Library have PowerSensor and AronSensor available.

@christiankral from your words I understand that you expect anyone to be at ease with PowerSensor, while more explanation is needed in AronSensor. According to this idea, I could, after @AHaumer opinion, open a new PR with this additional info to be added to Aron Sensor's. In that case, this PR should be taken as invalid.

I could add this just after "This device works only in three-phase systems without neutral."

beutlich:
@christiankral Ready to merge?

christiankral:
I assume @AHaumer agreed on @max-privato's proposal to for a new PR which includes more documentation. In this case and mark this PR as invalid instead.

christiankral:
> > ```
> > 2\. I will create a User's Guide for the `Analog` package
> > ```
> 
> Make sure to address #2990. Thanks.

OK, I created PR #3159 and included reference #2990. 

@max-privato In case you want to cite reference in English language for your new PR of the Aron sensor, you may included 

```html
<tr>
  <td>[<a href="http://brownwalker.com/book/1612335004">Malaric2011</a>]</td>
  <td>R. Malaric, <em>Instrumentation and Measurement in Electrical Engineering</em>, 
    BrownWalker Press, 2011</td>
</tr>
```

in `Modelica.Electrical.Polyphase.UsersGuide.References`.

max-privato:
Although this ticket is closed I added some comments in aprivate mail to @christiankral just because I did not know where was now the right place to post that comment. 

@christiankral In case you agree, at least partly, with my comment, please tell me by e-mail and I will share is, or part of it publicly in the place you suggest.

PR #3135, opened 2019-10-07
Fixes #778 by renaming "Commuting" to "TwoWay" 

dietmarw:
Also includes conversion script and test.
Note: The ticket suggests to replace "IdealCommuting" with  "TwoWay" instead of "IdealTwoWay" which is probably an oversight and would introduce inconsistency.

Fix #778.

PR #3137, opened 2019-10-07
Adds test models for TwoWay switch models.

dietmarw:


dietmarw:
> Should be merged after #3135.

Yes.

PR #3138, opened 2019-10-07
Fix #3096 by replacing T_heatPort with heatPort.T in documentation

### LINKED ISSUE #3096: Invalid reference to T_heatPort in line models

dietmarw:
Fix #3096.


dietmarw:
It looks like it and I missed that one. But #3101 is actually broken since there is no `T_actual` and some typos are also present. So I'd rather use this one.

PR #3133, opened 2019-10-04
Made engine designs consistent w.r.t. the whole design scenario

christoff-buerger:
This pull request fixes the inconsistencies between the `Engine1a`, `Engine1b` and `Engine1b_analytic` models of `Mechanics.MultiBody.Examples.Loops`. For details about these, please cf. Section 4.2 of [Modelica language extensions for practical non-monotonic modelling: on the need for selective model extension](https://modelica.org/events/modelica2019/proceedings/html/papers/Modelica2019paper3B1.pdf).

The pull request fixes #3132.

The changes are:
1. Consistent stop times:
   - Engine1b, Engine1b_analytic: stop time set from 0.5 to 5 (the time of Engine1a)
2. Consistent inertias:
   - Engine1bBase (=> Engine1b, Engine1b_analytic): set inertia(J = 1 "kg m^2") (the inertia of Engine1a)
3. Consistent FixedTranslation of cycle position:
   - Engine1bBase (=> Engine1b, Engine1b_analytic): set cylPosition(r = {0.15,0.45,0}) (the cycle position of Engine1a)
4. Consistent piston <-> cylinder connections:
   - Engine1b: connected pistion.a with cyclinder.b instead of rod3.a (the connection of Engine1a)
5. Consistent bearings:
   - Engine1b:
     - switched bearings b1 and b2 (according to Engine1a)
     - turned rod1 and connectingRod respectively
     - set rod1.r and connectingRod.r to {0,-0.2,0} (according to Engine1a)
6. Improved, consistent graphical layout

PR #3146, opened 2019-10-11
Fix for last-minute errors of #3133 pull request

christoff-buerger:
This is a fix for erroneous last-minute changes of the #3133 pull request. The problems of these are discussed at the end of #3133.

In summary, fa754542104e5042916d0cc5ef7c67dbd1033bbb made `Engine1a` also inherit from `Engine1bBase`, deleting the gas force in `Engine1bBase` and instead adding it separately to `Engine1b` and `Engine1b_analytic`. This is not right for the following reasons:
 1. As the name `Engine1bBase` suggests, it is the base-class for `Engine1b` and `Engine1b_analytic`, not  `Engine1a`.
 2. The description of `Engine1bBase` is erroneous now, because it refers to it introducing a gas-force.
 3. The idea of the whole scenario is to go from an engine without gas-force to one with it to finally an analytic solution. The challenge however is, that the required changes to do so are non-monotonic and therefore _cannot_ be modeled as a single inheritance sequence; this is problem inherent and explained in my [Modelica language extensions for practical non-monotonic modelling: on the need for selective model extension](https://modelica.org/events/modelica2019/proceedings/html/papers/Modelica2019paper3B1.pdf) paper. Any inheritance-based solution must split-off the inheritance tree at some point such that  `Engine1b` and `Engine1b_analytic` are not directly inheriting from `Engine1a` nor from each other.

This fix therefore does the following:
 1. `Engine1bBase` is the base class for `Engine1b` and `Engine1b_analytic` only. Its description is _"Common model of one cylinder engine with gas force."_ (**with** gas force). A gas force has been reintroduced in it.
 2. A new base class `Engine1aBase` is introduced. It encapsulates all common things and is the base for `Engine1a`. `Engine1bBase` inherits from it and just adds the gas force. The description of `Engine1aBase` is _"Common model of one cylinder engine without gas force."_ (**without** gas force).

Further, all code has been properly formatted.

### DIFF
--- a/Modelica/Mechanics/MultiBody/Examples/Loops/Engine1a.mo
+++ b/Modelica/Mechanics/MultiBody/Examples/Loops/Engine1a.mo
@@ -1,7 +1,7 @@
 within Modelica.Mechanics.MultiBody.Examples.Loops;
 model Engine1a "Model of one cylinder engine"
   extends Modelica.Icons.Example;
-  extends Utilities.Engine1bBase(inertia(phi(start=0), w(start=10)));
+  extends Utilities.Engine1Base;
   Modelica.Mechanics.MultiBody.Joints.Revolute b1(
     n={1,0,0},
     cylinderLength=0.02,
@@ -66,8 +66,7 @@ well-formed.
 </p>
 <p>
 An animation of this example is shown in the figure below.
-</p>
-
+</p><p>
 <img src=\"modelica://Modelica/Resources/Images/Mechanics/MultiBody/Examples/Loops/Engine.png\" alt=\"model Examples.Loops.Engine\">
-</html>"));
+</p></html>"));
 end Engine1a;
--- a/Modelica/Mechanics/MultiBody/Examples/Loops/Utilities.mo
+++ b/Modelica/Mechanics/MultiBody/Examples/Loops/Utilities.mo
@@ -798,8 +798,11 @@ solved analytically.
 </html>"));
   end EngineV6_analytic;
 
-  partial model Engine1bBase "Model of one cylinder engine with gas force"
+  partial block Engine1Base
+    "Base model for one cylinder engine"
 
+    inner Modelica.Mechanics.MultiBody.World world
+      annotation (Placement(transformation(extent= {{-100,-100},{-80,-80}})));
     Modelica.Mechanics.MultiBody.Parts.BodyCylinder piston(diameter=0.1, r={0,-0.1,0})
       annotation (Placement(transformation(
           origin={90,40},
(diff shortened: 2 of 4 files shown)


christoff-buerger:
**Note:** I am off on holiday the next two weeks.

Modelica/Mechanics/MultiBody/Examples/Loops/Engine1a.mo
```diff
@@ -1,50 +1,63 @@
 within Modelica.Mechanics.MultiBody.Examples.Loops;
-model Engine1a "Model of one cylinder engine"
+model Engine1a "Model of one cylinder engine without gas."
```
tobolar: This is not really needed as the other models say they add those gas force.

Modelica/Mechanics/MultiBody/Examples/Loops/Utilities.mo
```diff
@@ -798,99 +798,122 @@ solved analytically.
 </html>"));
   end EngineV6_analytic;
 
-  partial model Engine1bBase "Model of one cylinder engine with gas force"
+  partial block Engine1aBase
+    "Common model of one cylinder engine without gas force."
```
tobolar: Not necessary to say "without gas force" here. This is simply a base model of ANY one cylinder engine. So actually, the proper description shall be: "Base model for one cylinder engine".

Modelica/Mechanics/MultiBody/Examples/Loops/Utilities.mo
```diff
@@ -914,10 +937,20 @@ well-formed.
...
-</p>
-
+</p><p>
 <img src=\"modelica://Modelica/Resources/Images/Mechanics/MultiBody/Examples/Loops/Engine.png\" alt=\"model Examples.Loops.Engine\">
-</html>"));
+</p></html>"));
+  end Engine1aBase;
+
+  partial block Engine1bBase
+    "Common model of one cylinder engine with gas force."
```
tobolar: "Base model for one cylinder engine with gas force"

christoff-buerger:
> What is the reason to miss inertia(w(start=0)) in Engine1b? This setting is used in both Engine1b and Engine1b_analytic so it shall be defined in the base class Engine1aBase.

@tobolar: The start value of `inertia.w` is already set in `Engine1aBase` to `10`, i.e., `Engine1a` uses `inertia(w(start = 10))`, whereas `Engine1b` and `Engine1b_analytic` used to set `inertia(w(start = 0))` which I considered an inconsistency (I understood the model scenario to be -- at least conceptually -- a sequential improvement of a single case and therefore inconsistent start values to be an error). Now all models use the same inertia.

christoff-buerger:
@tobolar: Commit 40f3681b265403e19786ec3250e7d86373989296 incorporates your review comments:
 > 1. Do not use punctuation at the end of the model description.
 > 2. Following your argumentation, the lowermost base class shall actually be called Engine1Base, not Engine1aBase.

Regarding

 > In a PR, please try to omit changes like adding blanks around an equal sign (i.e. =). This is may be better for code readability but there is no such rule for MSL so far. Moreover, this doesn't really concern the PR and make it harder to follow your changes.

I will consider it in the future (at least split it into separate, formatting only commits), but did not undo the formatting improvements this time.

I hope it is Ok now?

beutlich:
> I hope it is Ok now?

@tobolar Do you agree?

christoff-buerger:
> 
> 
> This PR is far too noisy. I reverted all the dummy changes and only applied the relevant changes. The result is in my branch #3133-fix: [beutlich@2ae72fa](https://github.com/beutlich/ModelicaStandardLibrary/commit/2ae72fa16e3440bd11b19643fbf7d35ac4bd2d37)
> @tobolar @christoff-buerger If you agree I can force-push it here and merge afterwards.

Looks good to me. Just go ahead.

PR #3010, opened 2019-06-28
Merge HeatingDiode with Diode

beutlich:
@christophclauss Can you pease work on the following TODOs

- [x] Description for useTemperatureDependency -> Done.
- [x] Utilize useTemperatureDependency in the equations -> Done.
- [x] Dialog.enable annotation for N, EG, TNOM and XTI? -> Done.
- [x] Rename TNOM to Tnom (to match the parameter name of HeatingNMOS/HeatingPMOS)? -> Suspended.

See #2899.

PR #3167, opened 2019-10-23
Merge HeatingNMOS/HeatingPMOS with NMOS/PMOS

beutlich:
See #2899.

### DIFF
--- a/Modelica/Electrical/Analog/Examples/HeatingMOSInverter.mo
+++ b/Modelica/Electrical/Analog/Examples/HeatingMOSInverter.mo
@@ -21,9 +21,9 @@ model HeatingMOSInverter "Heating MOS Inverter"
         rotation=270)));
   Modelica.Thermal.HeatTransfer.Components.ThermalConductor TC1(G=0.01)
   annotation (Placement(transformation(extent={{0,-50},{20,-30}})));
-  Semiconductors.HeatingPMOS H_PMOS(useHeatPort=true)
+  Semiconductors.PMOS H_PMOS(useTemperatureDependency=true)
   annotation (Placement(transformation(extent={{-40,40},{-20,60}})));
-  Semiconductors.HeatingNMOS H_NMOS(useHeatPort=true)
+  Semiconductors.NMOS H_NMOS(useTemperatureDependency=true)
   annotation (Placement(transformation(extent={{-40,0},{-20,20}})));
   Modelica.Electrical.Analog.Sources.RampVoltage V(V=5, duration=1e-2)
   annotation (Placement(transformation(
--- a/Modelica/Electrical/Analog/Semiconductors.mo
+++ b/Modelica/Electrical/Analog/Semiconductors.mo
@@ -252,183 +252,70 @@ Stefan Vorkoetter - new model proposed.</li>
           Line(points={{30,-40},{20,-40}}, color={28,108,200})}));
   end ZDiode;
 
-model PMOS "Simple MOS Transistor"
+  model NMOS "Simple MOS Transistor with heating port"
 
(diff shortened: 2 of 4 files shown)


beutlich:
I guess, this also was overlooked in #3010 then. Easiest would be to set useHeatPort=useTemperatureDependency in the new models, right?

HansOlsson:
> I guess, this also was overlooked in #3010 then. Easiest would be to set useHeatPort=useTemperatureDependency in the new models, right?

That is one possibility, I tried to list all of them below:

- Don't set useHeatPort, let the user set both. Seems non-intuitive.
- Don't set useHeatPort, let the user set both; but make useTemperatureDependency conditionally enabled on useHeatPort. However, that means we cannot use a static temperature dependency.
- Add: useHeatPort=useTemperatureDependency (still allows all variant, but convenient default)
- Add: useTemperatureDependency=useHeatPort (still allows all variant, but convenient default)

To me the heat-port seems like the primary parameter, and thus the last option seems preferable; but I'm not an electrical engineer so I don't know which parameter is most natural.

beutlich:
If we have useTemperatureDependency=useHeatPort and have useHeatPort=true (as new default) we do not get the old Didoe/PMOS/NMOS behavior when using that new model (of v4.0.0).
But having useHeatPort=useTemperatureDependency and useTemperatureDependency=false (as default) and setting useTemperatureDependency=true when converting HeatingDiode etc. to Diode, we consider all use-case correctly.

HansOlsson:
> If we have useTemperatureDependency=useHeatPort and have useHeatPort=true (as new default) we do not get the old Didoe/PMOS/NMOS behavior when using that new model (of v4.0.0).
> But having useHeatPort=useTemperatureDependency and useTemperatureDependency=false (as default) and setting useTemperatureDependency=true when converting HeatingDiode etc. to Diode, we consider all use-case correctly.

Yes, I had just realized that as well. 

It's just that to me it seems natural to enable both temperature dependency and heat-port (if you want either of them), and thus that should be natural in the user-interface; i.e. the natural user experience is to set useTemperatureDependency. I don't know how we can do that in a simple way.

beutlich:
Parameters useTemperatureDependency and useHeatPort are independet, actually. useTemperatureDependency=false was introduced as new parameter for v4.0.0. Having useHeatPort=useTemperatureDependency enables a clean conversion. Users still can set useHeatPort to false/true since it is no final modification.

beutlich:
Is it actually a problem that useHeatPort is shown as check box in the UI?

beutlich:
There is a small UI issue in Dymola 2020 with depending parameters. This is how the new NMOS properties dialog looks:

grafik

If useTemperatureDependency is set to true, then implicitly useHeatPort is set to true implying that T should be drawn disabled (which is not the case):

grafik

HansOlsson:
Dymola uses a tri-state checkbox if its neither explicitly true nor false, and in this case the value is useTemperatureDependency (which happens to be false). Using the normal two-state would make it graphically difficult to see if a user changed the value between false and useTemperatureDependency.

Obviously we could use a different marker, but that adds complexity.
--
However, the point I wanted to make earlier was something else: can we make it easier for users to understand that it would be good to change useTemperatureDependency instead of useHeatPort? 

I don't see any amazing solutions, so just having it "earlier" in the dialog (as in this case), and reusing the logic in other models seems like the way forward.

beutlich:
> Dymola uses a tri-state checkbox if its neither explicitly true nor false

Which is a nice idea to avoid confusion.

> If useTemperatureDependency is set to true, then implicitly useHeatPort is set to true implying that T should be drawn disabled (which is not the case)

Still, you need to update the UI if the value of the tri-state checkbox changes due to changes of the connected parameter.

PR #3169, opened 2019-10-23
Fix #3165 by improving the icon design of the firing connector

### LINKED ISSUE #3165: Wrong position of fire input in GTO
The GTO fire input should be placed in contact or very near the actual gate wire. The present switch, instead, has the fuchsia fire input too far from the gate blue indicator: see the image below.

Immagine

dietmarw:
fix #3165


dietmarw:
@christiankral  Good point. I've corrected the semiconductor icon and also adjusted the position of the firing connection so it looks equally now:

image

dietmarw:
I did not touch the position of the name as this is not as per User's Guide anyway and would require a total overhaul of all components and is probably not worth it for these existing elements.

PR #3183, opened 2019-10-30
Add substrate (optional connector) to bipolar transistor models 

christiankral:
Refs #2146 and  #3178. This PR does not (yet) handle #3168.

This ticket covers two issues:

1. I actually consider using only implicit grounding to be a bug. This PR now includes a conditional substrate connector. The substrate connector is not available by default for compatibility reasons.  

2. I implemented initial conditions of the PNP transistor model and consider this as enhancement. Now the NPN and PNP transistor models are build with equal features.

PR #3187, opened 2019-11-01
Make the SimpleTriac model extend from conditional heatport model like the rest

dietmarw:
SimpleTriac was the only model in the Semiconductors package that did not extend from the conditionalHeatport model for some unknown reason. This fixes this. 

In addition some of the connectors were awkwardly placed inside this package.


dietmarw:
This should be merged after #3183

PR #3194, opened 2019-11-02
Add description comments to arguments of "built-in" mathematical functions

beutlich:
[This](https://stackoverflow.com/q/58627324/8520615) question on SO revealed that the arguments of the Math functions lack a description comment.


beutlich:
Can use log and log10 if you prefer.

PR #3197, opened 2019-11-04
Fix specific gas constant in BaseProperties of ReferenceAir

beutlich:
When reviewing #3190 for #2438 @HansOlsson and I noticed that the specific gas constant in the BaseProperties of ReferenceAir is wrong (as it assigns the universal gas constant instead). For better traceability (in the release notes) it is useful to have this bug-fix as separate pull request.

### DIFF
--- a/Modelica/Media/Air/ReferenceAir.mo
+++ b/Modelica/Media/Air/ReferenceAir.mo
@@ -146,7 +146,7 @@ The package Air_dT can be used as any other medium model (see <a href=\"modelica
         "Base properties of air"
 
       equation
-        MM = ReferenceAir.Air_Utilities.Basic.Constants.MM;
+        MM = Air_Utilities.Basic.Constants.MM;
         if dT_explicit then
           p = pressure_dT(d, T);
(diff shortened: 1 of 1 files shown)


thorade:
Modelica.Constants.R = 8.3144598 J/(mol.K)
Modelica.Media.Air.ReferenceAir.Air_Utilities.Basic.Constants.R = 287.117 J/(kg.K) 
Modelica.Media.Air.ReferenceAir.Air_Utilities.Basic.Constants.R_bar = 8.31451 J/(mol.K) 

So, this change would change the unit, and it is also a big change of the value. So, I cannot say right away that this is correct, instead I would have to look at the reference paper, and probably also compare values against another tool.

beutlich:
BaseProperties.R is the specific gas constant, see [doc.modelica.org](https://doc.modelica.org/Modelica%203.2.3/Resources/helpDymola/Modelica_Media_Interfaces_PartialMedium.html#Modelica.Media.Interfaces.PartialMedium.BaseProperties).

PR #3186, opened 2019-10-31
Update AronSensor.mo

max-privato:
Proposed the following documentation changes:
1) removed reference to the need to have sinusoidal voltages and currents (Aron sensor works with any shape of voltages and currents)
2) added basic explanation of the Aron connection and a reference.

PR #3217, opened 2019-11-10
Add DOI and ISBN number to reference examples

christiankral:
So far, DOI (and ISBN) numbers are not yet included in reference examples. This PR fixes this shortcoming. I therefore replaced [Andronov1973] (for which could neither find an ISBN nor a DOI) by my own book on Modelica. I searched for the DOIs and ISBNs of a couple other references already used in the MSL, but finally gave up as this took too long without having a result. If someone prefers another reference I am perfectly fine with any other proposal.


christiankral:
Refs #3186

Issue #3218, opened 2019-11-11
Improvements of FluxTubes


AHaumer:
Using FluxTubes to prepare examples of magnetic cricuits for teaching the basics of electrical engineering, I found some improvements:
- [x] Improvement of documentation and reference arrows of ElectroMagneticConverter (should be electric current and magnetic voltage resp. mmf) - see #3220 
- [x] Include a hint that correlation between electric and magnetic values is according to a "right-hand screw" - see #3220 
- [x] One ore more simple basic examples that also show reference arrows - see #3223
- [x] Improved icons for shapes, visualizing the shape - see #3219 
- [x] New shape CylinderSectionCircumferentialFlux, i.e. a section (w.r.t. angle) of a hollow cylinder, conducting flux in circumferential direction - see #3226 
- [x] New shape Toroid - see #3226 
- [x] Usage of SIunits.RelativePermeability mu_r instead of real mu_r
- [ ] annotation defaultComponentPrefixes = parameter for parameter records
- [ ] New electric-magnetic-mechanical element, representing a moving coil in an airgap, i.e. a combination of converter and force
- [ ] Find a backwards compatible solution for replacing or amending the Boolean parameter nonLinearPermeability with an enumeration to enable alternative  interpolation or approximation of material characteristics

I can't find Thomas Boedrich's colleague Johannes Ziske among the github users ...

AHaumer:
@christiankral should'nt we apply the same improvements to QuasiStatic.FluxTubes?

beutlich:
> I can't find Thomas Boedrich's colleague Johannes Ziske among the github users ...

I invited @jziske both to MA organizational acoount and MAP-Lib team years ago but he never accepted the invitations. That's the reason why he does not show up in the list of assignees or reviewers.

dietmarw:
@AHaumer Can you check if the last two open points are already fixed or can be fixed until tomorrow. Otherwise we should move them to a new ticket for the next version.

AHaumer:
Let's shift the last points to the next version ...

dietmarw:
@Ahaumer Please make a new ticket with the remaining parts. That way we can include this ticket in the release notes for MSL 4.0.0. Thanks

dietmarw:
Further improvements are dealt with by #3359. Thanks @AHaumer

PR #3220, opened 2019-11-11
Improve icon and documentation of ElectroMagneticConverter 

AHaumer:
Refers to #3218: 
improve icon (reference arrows) and documentation (include a hint about right-hand screw rule) of ElectroMagneticConverter


christiankral:
image

Ampere's law relates the terms **magnetomotive force** and **magnetic voltage**. This is a similar situation as with the relation of the **electromotive force** and the **electric voltage** in Kirchhoff's law. With these terms the two laws are:

- Ampere's law: `F = sum(V_m)`: The sum of all magnetomotive forces counted with the correct sign in a reference direction is equal to the sum of all magnetic voltages counted in the same reference direction 
- Kirchhoff's law: `E = sum(v)`: The sum of all electromotive forces counted with the correct sign in a reference direction is equal to the sum of all electric voltages counted in the same reference direction 

Modelica can neither treat magnetomotive nor electromotive forces directly in interconnected equivalent circuits. Instead the two laws are applied in a different way:

- Ampere's law: `sum(V_m) = 0`: The sum of all magnetic voltages counted with the correct sign in a reference direction is equal to zero
- Kirchhoff's law: `sum(v) = 0`: The sum of all electric voltages counted with the correct sign in a reference direction is equal to zero
 
Therefore each magnetomotive and electromotive is converted into an equivalent magnetic and electric voltage, respectively. In a magnetic circuit the electromotive force `N * I` is modeled as a magnetic voltage source `V_mF` oriented from plus to minus in the figure above. 

Therefore, there is a 
- right-hand assignment between the magnetomotive force `F` (thumb) and the current `I` (fingers), but a 
- left-hand assignment between the magnetic voltage of the source, `V_mF` (thumb) and the current `I` (fingers).

HansOlsson:
The new blue arrow from the screenshot above looks odd to me (and is easy to miss). Do we have any references showing such diagrams?

AHaumer:
@HansOlsson 
The old arrows are misleading: The blue arrow shows voltage, and voltage stems from derivative of flux. The orange icon shows magnetic potential difference. The direction of flux resp. derivative of flux depends on the rest of the magnetic circuit.
The "new" arrows show corresponding values: the current (blue arrow) drives a magnetic potential difference (the orange arrow). The other two arrows (induced voltage and magnetic flux) are omitted. 
We could enlarge the blue arrow (current) a little bit ...

AHaumer:
@DrMartinKuhn 
In fact in most domains Modelica follows Kirchhoff's laws (except stream variables):
The sum of potential differences in a closed loop is zero.
The sum of flow variables in a node is zero.
Of course this is true for magnetism, too:
Potential difference = magnetic potential difference
Flow variable = magnetic flux
And yes, in teaching and in some textbooks we use magnetic equivalent circuits and follow Kirchoff's laws. We also talk about magnetic conductance and magnetic resistance. Unfortunately, ferromagnetic materials show a nonlinear magnetic resistance resp. conductance due to saturation. 
As a summary: The term "Kirchhoff's laws" is correct and should support understanding of magentic circuits.

DrMartinKuhn:
> 
> 
> @DrMartinKuhn
> In fact in most domains Modelica follows Kirchhoff's laws (except stream variables):
> The sum of potential differences in a closed loop is zero.
> The sum of flow variables in a node is zero.
> Of course this is true for magnetism, too:
> Potential difference = magnetic potential difference
> Flow variable = magnetic flux
> And yes, in teaching and in some textbooks we use magnetic equivalent circuits and follow Kirchoff's laws. We also talk about magnetic conductance and magnetic resistance. Unfortunately, ferromagnetic materials show a nonlinear magnetic resistance resp. conductance due to saturation.
> As a summary: The term "Kirchhoff's laws" is correct and should support understanding of magentic circuits.

Sorry, I still have my problems with E=sum(v). Six lines below there is 0=sum(v). The later one I support.

HansOlsson:
> @HansOlsson
> The old arrows are misleading: The blue arrow shows voltage, and voltage stems from derivative of flux. The orange icon shows magnetic potential difference. The direction of flux resp. derivative of flux depends on the rest of the magnetic circuit.
> The "new" arrows show corresponding values: the current (blue arrow) drives a magnetic potential difference (the orange arrow). The other two arrows (induced voltage and magnetic flux) are omitted.
> We could enlarge the blue arrow (current) a little bit ...

My main concern was that we shouldn't try to invent some new graphical representation for this, but reuse what others do. To me there are two related issues with the design:

- The arrow is not easy to see, as it is shown in a busy section of the icon. Could it be moved to top-line of the icon instead? (I.e. horizontal arrow pointing right below the text "converter".)
- Having the blue line before the windings to me gives the impression that the current is input, i.e. a model with intended causality. (Similarly as electrical current sources have the arrow at the output indicating that current is an output of the sources.)

An alternative (used for transformers in many places - including Modelica.Electrical as I recall) is to not have arrows at all.

dietmarw:
One additional problem in the icon (old and new) is the direction of the V_m arrow. It follows the German [VZS (Verbraucherzählpfeilsystem)](https://de.wikipedia.org/wiki/Z%C3%A4hlpfeil) where voltage arrows always point from `+` to `-` but this convention confuses international users (especially in a non-German library) where the [PSC (Passive Sign Convention)](https://en.wikipedia.org/wiki/Passive_sign_convention) system defines voltage arrows always pointing towards `+`. Many international books don't use voltage arrows at all and simply indicate `+` and `-` which might be the safest bet.

DrMartinKuhn:
Considering the comments of Hans and Dietmar I propose to delete all arrows. For the proposed documentation, I have the feeling it does not improve the understanding but it is contradictory in some parts. Furthermore it tries to introduce a left or right hand system - while there was never defined any coiling direction in the original component and in the end it is all about definition since we transfer a 2D system to 1D. Therefore I propose to leave the old documentation and extend it by
"By definition, an electrical current into positive pin p effects a positive magnetomotive force V_m between magnetic port_p and port_n. This may cause a flux &Phi; out of port_p in a closed reluctance circuit. Coils with opposite winding direction can easily be considered by swapping the connection to their magnetic ports."

christiankral:
> Considering the comments of Hans and Dietmar I propose to delete all arrows. 

I agree to remove all arrows.

christiankral:
As of the icon, some suggestions and thoughts:

image

1. The reason for the symbol within the orange circle is to associate the **magnetomotive force** with the Greek letter Theta 

image

which is a purely German thing. As the magnetic voltage caused by the electric currents is a magnetic voltage source it may make sense to indicate it as a magnetic voltage source.

2. Does the dotted line really improve the icon?

AHaumer:
Ok @christiankral is correct: We have a definition of positive directions of voltage, current, mmf resp. magnetic potential difference and flux - there's no need for arrows.
I'd omit the "H" (theta) as well as the dotted line, but I'd draw the orange line through the orange circle (it is a source of magnetic potential difference, compare with electric voltage source).
@DrMartinKuhn 
We have to cases for Kirchhoff's voltage law:
If der(flux) is not present, we have the "original" law sum(v)=0.
But in case der(flux) is present, we can change this to sum(v)=E where E represents induced electromotive force.

christiankral:
One more comment: we could also use the following illustration and overcome all discussions and writing about signs and left and right hands:

image

christiankral:
The illustration is available at https://gitlab.com/christiankral/ModelicaImages/blob/master/ModelicaImages/Magnetic/FluxTubes/Basic/ElectroMagneticConverter/png/converter_signs.png

christiankral:
Hmmm... Actually, I just wonder what the new illustration shall show. In case all reference directions of the **converter** model shall be made clear, the flux arrow has to be changed. The flux `Phi` of the converter is **negative** in case of a passive reluctance network connected to the converter. 

image

AHaumer:
@dietmarw In MSL we generally follow VZS...

AHaumer:
@christiankral 
Either change the image, and note that the flux will be negative, or keep the image, and note that actual flux direction is opposite to Modelica's sign rules (VZS = Verbraucherzaehlpfeilsystem, i.e. power>0 designates power consumption).

christiankral:
@ahumer If you incorporate the update of the documentation of

- `Modelica.Magnetic.FluxTubes.Basic.ElectroMagneticConverter`
- `Modelica.Magnetic.FluxTubes.Basic.ElectroMagneticConverterWithLeakageInductance`

then please also consider the same changes in the `QuasiStatic` package:

- Modelica.Magnetic.QuasiStatic.FluxTubes.Basic.ElectroMagneticConverter

christiankral:
> @christiankral
> Either change the image, and note that the flux will be negative, or keep the image, and note that actual flux direction is opposite to Modelica's sign rules (VZS = Verbraucherzaehlpfeilsystem, i.e. power>0 designates power consumption).

In the image with the flux oriented counter clockwise, the VZS is fully obeyed both on the electrical and the magnetic side. The flux is negative if a passive reluctance network is connected. On my opinion this is the direction go.

dietmarw:
@AHaumer Referring to the **German** VZS does not help when the application is international and hence it causes confusion. So removing any arrows where we know that they are interpreted differently helps the user.  I personally would also prefer ONE system where everybody can agree on but that does not exist and probably will [never either](https://xkcd.com/927/). 
The one thing that can be agreed is the polarity as far as I know.

AHaumer:
@dietmarw 
referring to German VZS was only for discussion not for the documentation. 
I'd trust that Modelica users know about in which direction potential differences and flows are meant to be positive, therefore we omit all arrows.

DrMartinKuhn:
> 
> 
> Hmmm... Actually, I just wonder what the new illustration shall show. In case all reference directions of the **converter** model shall be made clear, the flux arrow has to be changed. The flux `Phi` of the converter is **negative** in case of a passive reluctance network connected to the converter.
> 
> image

To prevent any confusion, wouldn't it be better to just explain these ideas textually in the documentation layer? About the icons, I am in favour of converter3 which highlights the source of the magnetic potential variable. I also like the dots which indicate the coupling.

AHaumer:
Please review my proposal for clarifying the documentation of ElectroMagneticConverter (FluxTubes and QuasiStatic.FluxTubes) 0b5f736a9..fa576f4e3

dietmarw:
I'm still not sure if the arrow of the flux is correct. I'm used to this orientation:
FigC_1-5

Figure is taken from http://highered.mheducation.com/sites/0073529540/information_center_view0/index.html

dietmarw:
Btw, if the images come from a SVG source we should include the SVG directly and not bother with the PNGs. Then it is easier to do later edits.

beutlich:
> 
> 
> Btw, if the images come from a SVG source we should include the SVG directly and not bother with the PNGs. Then it is easier to do later edits.

Not before https://github.com/modelica/ModelicaSpecification/issues/174 is resolved.

dietmarw:
Right, I was assuming (wrongly) that SVG support was already official since Dymola and OM support it.

AHaumer:
@dietmarw You're right with your image, but flux is not directly associated to current. 
Modelica count's positive flow and potential difference from positive to negative connector. 
Positive current in the coil results in positive magnetic potential difference which in turn drives flux in the direction you indicated through the attached (passive) magnetic circuit, Therefore flux in the converter is negative (like current in a voltage source). 
Negative time derivative of flux results in a positive induced voltage in the coil.

dietmarw:
Would it be possible to also add the sources of the images besides the PNG? Preferable in SVG (there is a nice [tex text inkscape extension](https://textext.github.io/textext/) that lets you embed **editable** LaTeX code). I guess they are done in Ipe which is a nice but not so known tool.

HansOlsson:
> Would it be possible to also add the sources of the images besides the PNG? Preferable in SVG (there is a nice tex text inkscape extension that lets you embed editable LaTeX code). I guess they are done in Ipe which is a nice but not so known tool.

Since the images contain text this becomes more important when we want to support https://github.com/modelica/ModelicaSpecification/issues/302

beutlich:
> Since the images contain text this becomes more important when we want to support [modelica/ModelicaSpecification#302](https://github.com/modelica/ModelicaSpecification/issues/302)

Hm, it's only letters, right? I dod not think that we need to translate images in first place.

Might be rather related to https://github.com/modelica/ModelicaSpecification/issues/546.

AHaumer:
How do we proceed with that PR?

dietmarw:
Independent on any SVG support it would still be nice to have the sources of the images in the repository. Could you add them?

christiankral:
> Independent on any SVG support it would still be nice to have the sources of the images in the repository. Could you add them?

@dietmarw The currently used image and the source IPE file are stored at https://gitlab.com/christiankral/ModelicaImages/tree/master/ModelicaImages/Magnetic/FluxTubes/Basic/ElectroMagneticConverter

Shall we copy the IPE file to the folder `Modelica/Resources/Images/Magnetic/FluxTubes/Basic/` where the PNG file is stored?

dietmarw:
@christiankral Yes having the source files in the same location and same name (except the extension of course) makes it easy to find and edit at a later point in time.

beutlich:
Should we add them as `export-ignore` in .gitattributes? This should then also hold for other source files like:

* Modelica/Resources/Documentation/Clocked/Modelica_Synchronous.pptx
* Modelica/Resources/Documentation/Fluid/Stream-Connectors-Overview-Rationale.ppt
* Modelica/Resources/Documentation/Mechanics/Figure_PlanarLoopAnalytic.ppt

dietmarw:
Yes, that might be a good idea. @christiankral I suggest you add the IPE sources in `.ipe` xml format since then we can just globally exclude those files from export (for the release).

AHaumer:
@beutlich  Hope these ppt are avialabel as pdf, too.

beutlich:
> 
> 
> @beutlich Hope these ppt are avialable as pdf, too.

Yes, there are Modelica/Resources/Documentation/Clocked/Modelica_Synchronous.pdf and Modelica/Resources/Documentation/Fluid/Stream-Connectors-Overview-Rationale.pdf. File Modelica/Resources/Documentation/Mechanics/Figure_PlanarLoopAnalytic.ppt is only used to draw the figures, thus there is no PDF required here.

beutlich:
Please be aware, that I rebased the PR branch and resolved the merge conflict.

HansOlsson:
> Hm, it's only letters, right? I do not think that we need to translate images in first place.

The Ampere's Law image contains texts - not just single letters; and that raises the issue of translation.

HansOlsson:
> As for the translation since the image is part of the documentation than a documentation in a different language would simply have to come up with a different image if needed. In my eyes no need to hold off this PR because of that.

I agree with the conclusion. It was just that it implies that when translating we also need to consider this - e.g., allow different images for different languages, or revisit that part of the PR.

PR #3228, opened 2019-11-12
Utilize molar gas constant in Modelica.Media.Interfaces.{TemplateMedium, PartialLinearFluid}.BaseProperties

beutlich:
Utilize the molar gas constant `R` from Modelica.Constants instead of different approximations.

### DIFF
--- a/Modelica/Media/Air/ReferenceMoistAir.mo
+++ b/Modelica/Media/Air/ReferenceMoistAir.mo
@@ -2443,7 +2443,7 @@ package ReferenceMoistAir
         end IceConstants;
 
         constant IceConstants Constants(
-          R_bar=8.314472,
+          R_bar=Modelica.Constants.R,
           R_s=461.52364,
           MM=18.015268E-003,
           rhored=1.0,
@@ -2901,7 +2901,7 @@ package ReferenceMoistAir
         Real caaw=0;
         Real caww=0;
         Real cwww=0;
-        final constant Real R_bar=8.314472;
+        final constant Real R_bar=Modelica.Constants.R;
 
       algorithm
         //p_ws is the saturation pressure of water
--- a/Modelica/Media/Air/ReferenceMoistAir.mo
+++ b/Modelica/Media/Air/ReferenceMoistAir.mo
@@ -2443,7 +2443,7 @@ package ReferenceMoistAir
         end IceConstants;
 
         constant IceConstants Constants(
-          R_bar=Modelica.Constants.R,
+          R_bar=8.314472,
           R_s=461.52364,
           MM=18.015268E-003,
           rhored=1.0,
@@ -2901,7 +2901,7 @@ package ReferenceMoistAir
         Real caaw=0;
         Real caww=0;
         Real cwww=0;
-        final constant Real R_bar=Modelica.Constants.R;
+        final constant Real R_bar=8.314472;
 
       algorithm
         //p_ws is the saturation pressure of water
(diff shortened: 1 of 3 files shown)


thorade:
This would require reading the publication, because if a different value of the gas constant was used in the publication, it must also be used in the implementation to get matching results.

Modelica/Media/R134a.mo
```diff
@@ -9284,7 +9284,7 @@ Proceedings of the Joint Meeting of IIR Commissions B1, B2, E1, and E2, Padua, I
 
     record data
       constant SIunits.SpecificHeatCapacity R_s=81.4888564372;
-      // 8.314471/0.102032
+      // Modelica.Constants.R/MM
       constant SIunits.MolarMass MM=0.102032;
```
thorade: Instead of specifying both R_s and MM it would be better to calculate one of the two:
R_s = R/MM
MM = R/R_s

Modelica/Media/package.mo
```diff
@@ -3820,7 +3820,7 @@ package Interfaces "Interfaces for media models"
       h = cp_const*T;
       u = h - p/d;
       MM = 0.024;
-      R_s = 8.3144/MM;
+      R_s = Modelica.Constants.R/MM;
```
thorade: In the interface this is good!

Modelica/Media/package.mo
```diff
@@ -4764,7 +4764,7 @@ are described in
       p = state.p;
       T = state.T;
       MM = MM_const;
-      R_s = 8.3144/MM;
+      R_s = Modelica.Constants.R/MM;
```
thorade: This line looks good, it is in the interface.

Modelica/Media/Air/ReferenceMoistAir.mo
```diff
@@ -2443,7 +2443,7 @@ package ReferenceMoistAir
         end IceConstants;
 
         constant IceConstants Constants(
-          R_bar=8.314472,
+          R_bar=Modelica.Constants.R,
```
thorade: This will change results slightly. 
If we still want to change it like this, we should not specify both R_s and MM, but instead calculate one of the two from
R_s=R/MM

mestinso:
I agree with @thorade here. Changes in Modelica/Media/package.mo make sense and should be implemented. However, for the R134a and Moist air reference cases, it is arguably wrong since this will cause the results to change. The R values should be considered fossilized/internal factors that you shouldn't care about externally. You either need two changes (two wrongs making a right) or no changes at all to keep it consistent with the publication. I would vote for the latter.

Modelica/Media/R134a.mo
beutlich: Reverted.

Modelica/Media/Air/ReferenceMoistAir.mo
beutlich: Reverted.

Modelica/Media/package.mo
beutlich: Changing it in Modelica.Media.Interfaces.PartialLinearFluid.BaseProperties also affects Modelica.Media.CompressibleLiquids.LinearColdWater and Modelica.Media.CompressibleLiquids.LinearWater_pT_Ambient via inheritance.

beutlich:
> Changing it in Modelica.Media.Interfaces.PartialLinearFluid.BaseProperties also affects Modelica.Media.CompressibleLiquids.LinearColdWater and Modelica.Media.CompressibleLiquids.LinearWater_pT_Ambient via inheritance.

@casella Acceptable for you?

casella:
> 
> 
> > Changing it in Modelica.Media.Interfaces.PartialLinearFluid.BaseProperties also affects Modelica.Media.CompressibleLiquids.LinearColdWater and Modelica.Media.CompressibleLiquids.LinearWater_pT_Ambient via inheritance.
> 
> @casella Acceptable for you?

Absolutely!

In fact, those models do not really need a value for R, it is only provided because the interface asks for it. I'm not sure why we put some rounded numbers, but I guess that was the reason.

Please go ahead

PR #3233, opened 2019-11-14
Specify case with one coefficient and one variable in Modelica.Electrical.Spice3.Additionals.poly

### LINKED ISSUE #2794: Spice3 poly source
The Electrical.Spice3.Additionals.poly function has
```mo
    // case one coefficient
      if (na == 1) then
        v := a[1] * s[1];
        return;
      end if;
```
Am thinking that the condition should be `(na==1 and n==1)`. Neither is particularly useful when n>1, but the original has an asymmetry that is strange and which does not exist in SPICE, I think.

beutlich:
Only for one coefficient and one variable the polynomial is evaluated as linear function (with vanishing absolute term).

Closes #2794.

### DIFF
--- a/Modelica/Electrical/Spice3.mo
+++ b/Modelica/Electrical/Spice3.mo
@@ -3741,6 +3741,10 @@ If, e.g., time = 1.0, the current i =  0.0 (before event), 1.0 (after event)
 
     // case one coefficient and one variable
       if n == 1 and na == 1 then
+        /* Note: If the polynomial is one-dimensional and exactly one coefficient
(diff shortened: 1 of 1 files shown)


beutlich:
> 
> 
> Looks correct, but since it is so weird it would be helpful with additional documentation:
> E.g. https://www.macspice.com/ug/secC.html explains the special case as:
> 
> > Note: if the polynomial is one-dimensional and exactly one coefficient is specified, then SPICE assumes it to be p1 (and p0 = 0.0), in order to facilitate the input of linear controlled sources.

Yes, agreed. I just was not sure which Spice source we should cite here.

PR #3243, opened 2019-12-01
Addresses #3179 (battery parameters)

### LINKED ISSUE #3179: Change parameters of battery models
As discussed with @AHaumer, from the currently implemented parameter records some parameters shall be removed in order to use only physical parameters in the physical models. This change allows the strict separation of physical parameters and auxiliary calculation records which may be available from data sheets of batteries. Such an approach is already used in:

- `Modelica.Electrical.Machines.Utilities.SynchronousMachineData`
- `Modelica.Electrical.Machines.Utilities.TransformerData`

Changes to be applied: 

### `Modelica.Electrical.Batteries.ParameterRecords.CellData`

- Remove `Isc` as parameter `Ri` this the relevant parameter of the physical model
- Instead an auxiliary parameter record shall be implemented which computes the resistance `Ri` from the maximum open circuit voltage `OCVmax` and the short circuit current `Isc`

### `Modelica.Electrical.Batteries.ParameterRecords.RCData`  

- Remove time constant `T` as the physical parameters are fully determined by `R` and `C`
- Instead an auxiliary parameter record shall be implemented which utilizes the resistance and the time constant and calculates the capacitance from these parameters

AHaumer:
simplified battery parameterization (resolve #3179):
leave calculation of inner resistance from short circuit and calculation of RC-capacitance from time constant to pre-processing.


AHaumer:
adapted linewidth of icons (@dietmarw )

dietmarw:
Better but still not really nice. OM does do a slightly better job. But maybe the circuit just needs to get enlarged. 
OM: 
image
Dymola:
image

AHaumer:
@dietmarw here you go ... I think I won't become much better rendered. 
Do you have another suggestion to visualize the enclosed RC?

Modelica/Electrical/Batteries/BaseClasses/BaseCellStack.mo
```diff
@@ -87,11 +87,11 @@ equation
           textColor={0,0,255},
           textString="%name"),
         Rectangle(
-          extent={{-90,50},{90,-50}},
+          extent={{-90,60},{90,-60}},
           lineColor={0,0,255},
           radius=10),
         Rectangle(
-          extent=DynamicSelect({{70,-30},{-70,30}},{{70,-30},{70-140*SOC,30}}),
```
dietmarw: I guess this removal was accidentally (since Dymola just loves to overwrite dynamic select statements)

PR #3244, opened 2019-12-01
Fix various compiler issues or warnings

beutlich:
Tested with

* BCC 5.5
* GCC 7.4.0 (Cygwin)
* GCC 5.1.0 (tdm64-1)
* LCC
* MSVC 6.0
* VS2005
* VS2010
* VS2012
* VS2017


beutlich:
@HansOlsson Agreed and implemented as suggested. Just a slight correction: The LL prefix seems to be first supported by _MSC_VER >= 1300, which is VS2002), but I can longer test since I do not have it available. But since I have it also at different places (when I still had it available or found the proper reference), I believe we can leave it as proposed.

PR #3247, opened 2019-12-02
WIP: Add date and time utilities

beutlich:
Move getTime from System to Internal package and introduce a new package Time with function getTime returning a record.


HansOlsson:
The general idea is good, but:

- I don't like the abbreviations in the record (and as outputs): min, mon, ms, even if consistent with "struct tm". (As far as I understand the names should be singular: second, minute, hour, year.)
- I am unsure if we should have leap-year utilities. The reason is that it leads to people doing their own leap-year-calculations and that can lead to annoying bugs. Having a full set of utilities to avoid the need for that might be a better solution (that includes conversions between date-time and number, and handling of date-time-differences).

I tried to look at the Buildings-library. It uses full names in the records, but unfortunately it contains its own leap-year handling (hopefully correct):
https://github.com/lbl-srg/modelica-buildings/blob/master/Buildings/Utilities/Time/CalendarTime.mo

beutlich:
I started with an operator record for TimeType to also allow addition, subtraction etc. That meant, we also would need to consider a TimeDeltaType and conversion to absolute time, which is not straight-forward. Therefore, I rejected this first idea and went for the TimeType record as absolute point in time and without numerical conversion. (String conversion is still to do, but without handling of leap-seconds).

> I don't like the abbreviations in the record (and as outputs): min, mon, ms, even if consistent with "struct tm". (As far as I understand the names should be singular: second, minute, hour, year.)

OK, will rename.

> I am unsure if we should have leap-year utilities.

Hm, @GallLeo what do you think? @mwetter considers leap years only in range [2020, 2030] which might not be sufficient in a general purpose library.

sjoelund:
If you want to consider leap years, daylight savings, etc... You need to work on some absolute unit of s/etc relative from 1970... I think that's the only sane way of dealing with any kind of delta and ambiguity... Just convert between tm and time_t if you want to use that (and run the calculation in external C).

beutlich:
[Here](https://gist.github.com/beutlich/f090c7a010b109fa1418495f64ff2705) is a first (and incomplete) draft for conversion of date/time to string. Not sure if it is good enough ❔  . Any feedback is appreciated.

beutlich:
@HansOlsson @DagBruck @t-sommer Yesterday, I noticed by chance that there are Testing.Utilities.Time{DateTime, Duration} available in Dymola as operator records. This is exactly the way I wanted to go here for MSL, too. I believe these utilities came up as necessary side-product when meaasuring time in Modelica. Now, that I am aware of this library, I am also biased with the further development. My question is, if you could think of contributing or sharing Modelica code for further development as OSS, such that I am sure, no copyright will be violated?

m-kessler:
We would be happy to move the operator records from the Testing library to the MSL. 
@beutlich How shall we proceed? My recommendation would be, that I first push a purely MSL-based version of the operator records to this pull request. Then we can review the code and merge it with your work.

beutlich:
@m-kessler Thank you for the generous offer. I rebased the commits of this PR on master and resolved the merge conficts. This PR branch is now ready to be used. I added you as contributor (with access role = Triage) such that you can directly push to this branch (hopefully).

I already created a TimeType record. This can be now enhanced by using the two different operator records from the Testing package. It would be good if we can use the exsting names of the TimeType record components (which are different than the Testing package.)

Modelica/Utilities/Strings.mo
```diff
@@ -172,6 +172,50 @@ isEmpty(\"a\");      // returns false
...
+    input String searchString "Substring to look for";
+    input Boolean caseSensitive=true
+      "= false, if lower and upper case are ignored for the search";
+    output Boolean result "= true, if searchString can be found anywhere in string";
+
+  algorithm
+    if length(searchString) > 0 then
+      result := find(string, searchString, 1, caseSensitive) > 0;
+    else
+      result := string == searchString;
```
beutlich, line 189: We might define the case with empty searchString to always return `true` (as e.g., Python does).

beutlich:
@sjoelund Do you have an idea why 53dc84306e7f4ada6fa28db5535c0017a260b25d cannot be built by OMC Jenkins?

grafik

Modelica/Utilities/Time.mo
```diff
@@ -0,0 +1,1394 @@
...
+      import Modelica.Icons.Function;
+
+      function subtract "Return time delta between dt2 and dt1 as Duration"
+        extends Function;
+
+        import Modelica.Utilities.Time.Duration;
+
+        input DateTime dt1;
+        input DateTime dt2;
+        output Duration result "= dt1 - dt2";
```
beutlich, line 563: SimulationX raises the Check.TypeInconsistentEquation warning

```
The types of the left and right expression in the equation must be compatible. left: record Modelica.Utilities.Time.Duration, right: record Modelica.Utilities.Time.DateTime
```

Modelica/Resources/C-Sources/ModelicaTime.c
```diff
@@ -0,0 +1,245 @@
...
+        /* Avoid calling mktime for New Year 1970 in local time zone */
+        const time_t startTime = epoch(0, 0, 0, 2, 1, 1970);
+        return difftime(endTime, startTime) + 86400.0 + ms/1000.0;
+    }
+    else {
+        const time_t startTime = epoch(0, 0, 0, 1, 1, refYear);
+        return difftime(endTime, startTime) + ms/1000.0;
+    }
+#endif
+}
```
m-kessler, line 89: There seem to be some troubles. With Visual Sutdio 2015 in Dymola 2020x these example do not work:

    Modelica.Utilities.Internal.Time.diffTime(0, 0, 0, 0, 1, 1, 1970);       // does not work
    Modelica.Utilities.Internal.Time.diffTime(0, 0, 0, 0, 1, 1, 1969, 1969)  // does not work
    
For later years it works
    
    Modelica.Utilities.Internal.Time.diffTime(0, 0, 0, 0, 1, 1, 1971, 1971); // = 0.0
    Modelica.Utilities.Internal.Time.diffTime(0, 0, 0, 0, 1, 1, 2019, 2019); // = 0.0

Adding 1h also fixes the problem.
    
    Modelica.Utilities.Internal.Time.diffTime(0, 0, 0, 1, 1, 1, 1970);  // = 3600.0

Negative diffTime also has some troubles:
    
    Modelica.Utilities.Internal.Time.diffTime(0, 0, 0, 1, 1, 1, 1969);       // does not work    
    Modelica.Utilities.Internal.Time.diffTime(0, 0, 0, 0, 1, 1, 1970, 1971); // does not work

Again, adding 1h fixes the problem.

    Modelica.Utilities.Internal.Time.diffTime(0, 0, 0, 1, 1, 1, 1970, 1971); // = -31532400.0

m-kessler:
The new external function ModelicaTime_strftime is nice and definitely the way to go (using the existing C functions to format time, instead of rewriting it in Modelica). However, currently it has troubles with summertime:

    String(DateTime.now())                    // = 2020-04-14 15:09:41, should be 2020-04-14 14:09:41
    String(DateTime(2020, 4, 1, 0, 0, 0, 0)); // = 2020-04-01 01:00:00, should be 2020-04-01 00:00:00
    String(DateTime(2020, 3, 1, 0, 0, 0, 0)); // = 2020-03-01 00:00:00, as expected

maltelenz:
There seem to be conflicts in the `Duration` constructors.

When calling `Duration(1)`, it is not clear if `Modelica.Utilities.Time.Duration.'constructor'.fromInput` or `Modelica.Utilities.Time.Duration.'constructor'.fromSeconds` should be used. The specification says in 14.2 that:

>  there shall not exist any potential call that lead to multiple matches

m-kessler:
> There seem to be conflicts in the `Duration` constructors.

@maltelenz true. This commit would fix it https://github.com/m-kessler/ModelicaStandardLibrary/commit/e5477009808b5801560402c8652f3c8486e3c6d5 - making `fromSeconds` the new default constructor. @beutlich shall we merge it?

beutlich:
> @beutlich shall we merge it?

Sure, please go ahead. Push directly to this PR branch as you already did.

I'll answer the other issues you observed later.

Modelica/Resources/C-Sources/ModelicaTime.c
beutlich: Many of the corner cases are addressed by 2bc5990c08f32e10fc7944879f348c44cf6594c1, escept that `mktime` does not succeed for years earlier than 1970.

beutlich:
> 
> 
> The new external function ModelicaTime_strftime is nice and definitely the way to go (using the existing C functions to format time, instead of rewriting it in Modelica). However, currently it has troubles with summertime:
> 
> ```
> String(DateTime.now())                    // = 2020-04-14 15:09:41, should be 2020-04-14 14:09:41
> String(DateTime(2020, 4, 1, 0, 0, 0, 0)); // = 2020-04-01 01:00:00, should be 2020-04-01 00:00:00
> String(DateTime(2020, 3, 1, 0, 0, 0, 0)); // = 2020-03-01 00:00:00, as expected
> ```

Confirmed. ba9d075443ccc7a157a188b1182a5f285e9af818 should fix it, but only could test on Windows.

ModelicaTest/Math.mo
```diff
@@ -32,6 +32,14 @@ extends Modelica.Icons.ExamplesPackage;
     assert(Math.isPowerOf2(1), "isPowerOf2(1) is wrong");
     assert(Math.isPowerOf2(4), "isPowerOf2(4) is wrong");
     assert(not Math.isPowerOf2(9), "isPowerOf2(9) is wrong");
+
+    assert(Math.nearestInteger(1.4999)==1, "nearesetInteger(1.4999) failed");
+    assert(Math.nearestInteger(1.5)==2, "nearesetInteger(1.5) failed");
+    assert(Math.nearestInteger(-1.4999)==-1, "nearesetInteger(-1.4999) failed");
+    assert(Math.nearestInteger(-1.5)==-2, "nearesetInteger(-1.5) failed");
```
beutlich, line 39:
```suggestion
    assert(Math.nearestInteger(1.4999)==1, "nearestInteger(1.4999) failed");
    assert(Math.nearestInteger(1.5)==2, "nearestInteger(1.5) failed");
    assert(Math.nearestInteger(-1.4999)==-1, "nearestInteger(-1.4999) failed");
    assert(Math.nearestInteger(-1.5)==-2, "nearestInteger(-1.5) failed");
```

Modelica/Utilities/Strings.mo
m-kessler: Done with 6954f65 - Return true for empty search strings of Strings.contains

maltelenz:
Unfortunately this pull request exposed some shortcomings on operator records in our tool, so I'm not in a good position to do a thorough review. Apart from the issue I mentioned in the previous comment (which is now fixed), all issues I'm seeing so far are issues in the tool, not with the changes in the pull request.

HansOlsson:
Oh, and ModelicaTest.Utilities.DateTime fail due to dates before 1970 (for 1901, 1910, 1912).
The first failure seems to be for:
```
  ref_dt:= DateTime(1901, 12, 13, 20, 45, 53, 0);
  ref_r := -2147483647;

  act_dt:= DateTime(ref_r);
  act_r := DateTime.epoch(ref_dt);
```
It's slightly complicated to investigate this in Dymola since we normally pretty-print operator records and thus even `Modelica.Utilities.Time.DateTime(1901,12,13,20,45,53,0)` will fail.

Either we have to work around that - or just document it and don't test it.

beutlich:
> It should -2d -5min, right?

Correct. Fixed by ca583ce273096856d65c7964b216d33c0bb375eb.

maltelenz:
[Here](https://github.com/beutlich/ModelicaStandardLibrary/blob/d2dbe66732b40039450f7e459fc9229993016d8f/Modelica/Utilities/Time.mo#L732) an operator record is extended, but the specification says in the beginning of Chapter 14:

>  It is not legal to extend from an operator record, except as a short class definition 

This seems like invalid Modelica?

beutlich:
Hm, I only looked at the inheritance table of section [7.1.3](https://specification.modelica.org/v3.4/Ch7.html#restrictions-on-the-kind-of-base-class) where such an inheritance is legal.

m-kessler:
I usually declare icon-only-classes as class, as everyone can inherit from that. Should also work here, right? The pedantic check in Dymola 2021 does not complain.

maltelenz:
> I usually declare icon-only-classes as class, as everyone can inherit from that. Should also work here, right? The pedantic check in Dymola 2021 does not complain.

The checking in development versions of SystemModeler agrees `class` is fine.

beutlich:
> 
> We could make it a class then.

Also works in SimulationX without warnings.

Modelica/Utilities/Time.mo
```diff
@@ -0,0 +1,1443 @@
...
+
+      input Duration d1;
+      input Duration d2;
+      output Boolean result "= d1 <> d2";
+
+    algorithm
+      result := not d1 == d2;
+    end '<>';
+
+    encapsulated operator function '>' "Check if Duration d1 is larger than d2"
```
beutlich, line 1069: larger -> greater

Modelica/Utilities/Time.mo
```diff
@@ -0,0 +1,1443 @@
...
+      input Duration d1;
+      input Duration d2;
+      output Boolean result "= d1 >= d2";
+
+    algorithm
+      result := d1 == d2 or d1 > d2;
+
+    end '>=';
+
+    encapsulated operator function '<' "Check if Duration d1 is smaller than d2"
```
beutlich, line 1096: smaller -> less

beutlich:
* I spent many hours to work around the 1970 restriction that is in current code. The solution is not yet ready and not that nice. My proposal would be to accept the current implementation and refine once we got the requirement to support dates before 1970.
* Currently, ConvertModelica_from_3.2.3_to_4.0.0.mos is used for the conversion. This needs to be changed of course as MSL v4.00 is already released.
* The PR is more than 7 months old and has not seen reviews from @sjoelund (as library officer for C sources) or @GallLeo (who raised the idea initially). Would be very good to see some feedback from you, too. Thanks.

Modelica/Utilities/Internal.mo
```diff
@@ -240,4 +240,233 @@ especially if the operating system supports Unicode characters.
...
+      extends Modelica.Icons.Function;
+      input Integer ms "Millisecond";
+      input Integer sec "Second";
+      input Integer min "Minute";
+      input Integer hour "Hour";
+      input Integer day "Day";
+      input Integer mon "Month";
+      input Integer year "Year";
+      input Integer epoch_year = 1970 "Reference year";
+      output Real seconds "Elapsed seconds since epoch_year in the current time zone";
```
HansOlsson, line 257: Shouldn't it be since _start of_ epoch_year?

Modelica/Utilities/Internal.mo
```diff
@@ -240,4 +240,233 @@ especially if the operating system supports Unicode characters.
...
+      input String str "Formatted date and time string";
+      input String format = "%Y-%m-%d %H:%M:%S" "Format string passed to strptime";
+      output Integer ms "Millisecond";
+      output Integer sec "Second";
+      output Integer min "Minute";
+      output Integer hour "Hour";
+      output Integer day "Day";
+      output Integer mon "Month";
+      output Integer year "Year";
+      external "C99" ModelicaTime_strptime(ms, sec, min, hour, day, mon, year, str, format)
```
HansOlsson, line 373: I find it confusing that we have C, C90 (which should be the same as C, right? - note that the Modelica specification calls it C89 due to ANSI/ISO difference) and C99. It's not clear why, and since we only have one C-source file where the functions are just defined without anything extra I don't see how it will work.

Modelica/Math/nearestInteger.mo
```diff
@@ -0,0 +1,50 @@
...
+
+<h4>Example</h4>
+<blockquote><pre>
+import Modelica.Math;
+Math.nearestInteger(0.4);                     // = 0
+Math.nearestInteger(0.5);                     // = 1
+Math.nearestInteger(-0.4);                    // = 0
+Math.nearestInteger(-0.5);                    // = -1
+Math.nearestInteger(0.3999999999999999+0.1);  // = 0
+Math.nearestInteger(1.39999999999999999+0.1); // = 1 (erroneous border case, see note below)
```
HansOlsson, line 36: I don't see why we describe this special case as erroneous.
It's not a problem with nearestInteger in itself, but that the input is rounded before the call; so from the perspective of nearestInteger we are computing the nearest integer to 1.5.

Modelica/Resources/C-Sources/ModelicaTime.c
```diff
@@ -0,0 +1,260 @@
...
+        calendarTime = calendarTime + epoch(0, 0, 0, 1, 1, refYear);
+    }
+
+#if defined(_POSIX_)
+    tlocal = localtime_r(&calendarTime, &tres); /* Time fields in local time zone */
+#elif defined(_MSC_VER) && _MSC_VER >= 1400
+    localtime_s(&tres, &calendarTime);          /* Time fields in local time zone */
+    tlocal = &tres;
+#else
+    tlocal = localtime(&calendarTime);          /* Time fields in local time zone */
```
HansOlsson, line 144: I would have preferred if we used tres after this and only in this case used tlocal at all, and use tres=*tlocal; after the call of localtime as copying a struct has been in C since the early days (it might have been missing in some variants prior to C89/C90). (And time.h should, since C89/C90 define 'struct tm' in time.h.)

HansOlsson (approved):
Looks ok, except for the minor comments.
I would prefer that they are improved, but I don't find them blocking.

Modelica/Utilities/Internal.mo
beutlich: OK.

Modelica/Utilities/Internal.mo
beutlich: I set "C99" to explicitly enable all format cases of strftime (see http://www.cplusplus.com/reference/ctime/strftime/). I guess we can simply set "C" here, if this is OK for you.

HansOlsson: Yes, I think "C" would be better - since entirely removing the function for C89/C90 seems excessive. But possibly a comment about the new formats in C99.

Modelica/Math/nearestInteger.mo
beutlich: @m-kessler I cannot reproduce your results for Math.nearestInteger(0.3999999999999999+0.1) and Math.nearestInteger(1.39999999999999999+0.1) in Dymola 2021.

Modelica/Utilities/Internal.mo
```diff
@@ -240,4 +240,233 @@ especially if the operating system supports Unicode characters.
...
+      input Real seconds "Elapsed seconds since epoch_year";
+      input Integer epoch_year = 1970 "Reference year";
+      output Integer ms "Millisecond";
+      output Integer sec "Second";
+      output Integer min "Minute";
+      output Integer hour "Hour";
+      output Integer day "Day";
+      output Integer mon "Month";
+      output Integer year "Year";
+      external "C90" ModelicaTime_localtime(ms, sec, min, hour, day, mon, year, seconds, epoch_year)
```
beutlich, line 355: @HansOlsson I explicitly set "C90" here since the C implementation uses modf (http://www.cplusplus.com/reference/cmath/modf/) of C90. OK for you?

Modelica/Resources/C-Sources/ModelicaTime.c
beutlich: I see. FYI, the above implementation matches the current implementation of MSL 3.2.3 or 4.0.0

https://github.com/modelica/ModelicaStandardLibrary/blob/1adccd9a281c14d89abec1155a41c486984c1d74/Modelica/Resources/C-Sources/ModelicaInternal.c#L1297-L1304

HansOlsson: Ok, I still think the change would be good, but it's not that important.

Modelica/Utilities/Internal.mo
HansOlsson: Sure.

beutlich: OK.

Modelica/Math/nearestInteger.mo
beutlich: Removed the "border cases".

sjoelund (changes requested):
I think it's probably fine, but not easy to review it. There seems to be unrelated changes such as the nearestInteger function being added that should be its own pull request.

beutlich:
> I think it's probably fine, but not easy to review it. There seems to be unrelated changes such as the nearestInteger function being added that should be its own pull request.

There now are #4495 for adding Math.nearestInteger and #4493 for adding String.contains.

Modelica/Resources/C-Sources/ModelicaTime.c
```diff
@@ -0,0 +1,259 @@
...
+        free(format2);
+        if (retLen > 0 && retLen <= maxSize) {
+            return (const char*)timePtr;
+        }
+        else {
+            return "";
+        }
+    }
+#endif
+}
```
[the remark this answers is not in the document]
beutlich: `timePtr` was allocated by `ModelicaAllocateString` and thus its lifetime is managed by the runtime. It must not be freed using `free`.

GallLeo:
We should coordinate with @pstelzig
https://github.com/pstelzig/DateTime
(i.e. check this PR for completeness).

PR #3255, opened 2019-12-07
Move signal output of sensors to standard locations

christiankral:
The locations of the signal outputs of the `HeatTransfer` sensors

image

are moved to their standard locations outside the icon boundaries:

image


beutlich:
Was the issue also present in MSL 3.x.y? If yes, we should remove the `V: 4.0.0-dev ` label.

PR #3258, opened 2019-12-10
Fix waitTime visualization in transition icons

beutlich:


beutlich:
enableTimer is a parameter, thus there is no much dynamic behaviour during simulation.

PR #3266, opened 2019-12-15
Update issue links or revision links

beutlich:
Completely remove all links to trac.modelica.org.

This addresses the remaining ones from #2821.


beutlich:
> 
> 
> Looks good. Are you sure there aren't any more references to these pages?

Only check Modelica files so far. Still found some more in these ResolvedTracTickets.html files and in

https://github.com/modelica/ModelicaStandardLibrary/blob/1a711dc57fe5b2e979abe9070fdb411451f1eb14/Modelica/Resources/BuildProjects/autotools/configure.ac#L3

PR #3272, opened 2019-12-17
maint/3.2.3: Bump library version of libmodelicastandardtables

beutlich:
This was forgotten to update when releasing MSL v3.2.3.


Modelica/Resources/BuildProjects/autotools/configure.ac
```diff
@@ -1,6 +1,6 @@
 dnl Init
 AC_PREREQ([2.63])
-AC_INIT([Modelica Standard Library Tables],[3.2.2],[https://trac.modelica.org/Modelica],[libmodelicastandardtables],[https://modelica.org])
+AC_INIT([Modelica Standard Library Tables],[3.2.3],[https://trac.modelica.org/Modelica],[libmodelicastandardtables],[https://modelica.org])
```
dietmarw: What about the URL here?

beutlich: Will only fix in master.

PR #3274, opened 2019-12-18
Fix broken html tags

tobolar:
See also PR #3270.

Issue #3060, opened 2019-07-18
Code / Formatting guidelines checker


AHaumer:
In a discussion the idea came up to have tool independent checker for coding and formatting. 
I know that some parts of such a thing exist in some tools, but it would be nice to have a tool independent solution (like the regression check!). 
Let the tool run through all models of a library, getting a list of things tho improve, like:
parameter description missing, icon color is wrong, name in icon layer has wrong size and color
and many more - fortunately we have guidelines in the UsersGuide.

beutlich:
We can discuss it at the scheduled MAP-Lib strategy meeting in Nov 2019.

atrosinenko:
I have [recently implemented](https://github.com/pmd/pmd/pull/2041) an initial support of Modelica in [PMD](https://github.com/pmd/pmd/) static code analyzer. For now, it is _experimental_ (but already merged into `master`) and has only demo rules, but:
* PMD has language-agnostic "XPath rule" built-in, so you can write some XPath queries that mark matched nodes as violations - without any Java coding at all, just external XMLs with rules
  * one of these demo rules is

        //ClassSpecifier/*[SimpleName[1]/@Image != SimpleName[last()]/@Image]

    it basically checks for `model Name end OtherName;` mistakes
* PMD Modelica language module already has some more-or-less MLS-aligned flattening process to resolve type references
    * the other demo rule tries to find suspicious class name shadowing
    * requires Java coding to utilize
* PMD has Copy-Paste Detector module that can find large-enough copied parts
  * in some sense, it is even harder to review: if someone sends a bad pull request, it would probably be rejected or improved by maintainers, but if some other PR has parts of code copied from other places of the same repo (already reviewed and good quality!), then maintainer need to know **all codebase at once** to notice copy-paste
  * it seems, it already found a [180-line](https://github.com/pmd/pmd/pull/2041#issuecomment-539368857) copy-paste in MSL

It is still not published as part of the released Binary Distribution (the next release is planned at the end of January), so manual build of `master` branch is required. Also, it is still experimental, so some API still may change.

Last but not least: sometimes Modelica compilers crash with `Internal compiler error` or something like this. There exist many different tools to minimize source code that crash the compiler. I have implemented yet another as a tool for PMD that should support any "fully-featured" (that is, not CPD-only) PMD language (including Modelica). Its maintainers placed it to separate repo [pmd/pmd-scm](https://github.com/pmd/pmd-scm) as another experimental feature.

[...]

PR #3279, opened 2019-12-20
Update CI to check HTML documentation for tag validity

beutlich:
Related to #3060 and https://github.com/modelica/ModelicaStandardLibrary/pull/3274#issuecomment-567102255.

The python script has some troubles with self-closing void tags `<br/>` or `</br>` which I simply replaced by `<br>` in https://github.com/modelica/ModelicaStandardLibrary/pull/3274/commits/87d7fa9dfe907360056cd49b7f755c9e8bdd79bf for #3274.

One other false-positive are multi-line non-void tags. Luckily there is only one occurrence in all of MSL, which I edited in https://github.com/modelica/ModelicaStandardLibrary/pull/3274/commits/1b7ed59564d6d5122cf423d03eabc61b38e10b8b for #3274.

https://github.com/modelica/ModelicaStandardLibrary/blob/289ba3c99d7773065dfaee9cc5a8d2860616b8b0/Modelica/Media/IdealGases/Common/package.mo#L1058-L1059

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,13 +5,15 @@
 # to native line endings on checkout.
 *.c text
 *.h text
+*.htm* text
+*.md text
 *.mo text
 *.mos text
 *.order text
+*.py text
+*.yml text
 .gitattributes text
 .gitignore text
-*.md text
-*.htm* text
 .mailmap text
 
 # git archive keyword substitions for the main package files
@@ -29,4 +31,5 @@ ObsoleteModelica4.mo export-subst
 CONTRIBUTING.md export-ignore
 .mailmap export-ignore
 .travis.mos export-ignore
+.travis.py export-ignore
 .travis.yml export-ignore
--- a/.travis.py
+++ b/.travis.py
@@ -18,10 +18,13 @@ import sys
 void_tags = ('area', 'base', 'br', 'col', 'embed', 'hr', 'img', 'input', 'keygen', 'link', 'menuitem', 'meta', 'param', 'source', 'track', 'wbr')
 
 # See https://haacked.com/archive/2004/10/25/usingregularexpressionstomatchhtml.aspx/
-pattern = re.compile(r'</?\w+((\s+\w+(\s*=\s*(?:\\"(.|\n)*?\\"|\'(.|\n)*?\'|[^\'">\s]+))?)+\s*|\s*)/?>')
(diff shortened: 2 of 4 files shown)


beutlich:
CI will run successfully once #3274 is merged.

mtiller:
Wouldn't it be better to just use an existing library for this?  I'm not that familiar with the Python ecosystem, but a [simple search](https://www.google.com/search?q=python+tidy+html&oq=python+tidy+html&aqs=chrome..69i57j69i64.2058j0j1&sourceid=chrome&ie=UTF-8) turned up many options.  Seems like they will catch all the issues (which aren't really specified here) and more.

beutlich:
Yes, HTML Tidy is where I started, basically by running https://github.com/RWTH-EBC/HTML-Tidy-Modelica.

The problem with HTML Tidy is that it only detects invalid HTML tags as errros, e.g., 

```
[-- C:\Projekte\MSL\Modelica\Electrical\Batteries\ParameterRecords\TransientData\CellData.mo ]
line 31 column 93 - Error: <cde> is not recognized!


[-- C:\Projekte\MSL\Modelica\Electrical\Batteries\UsersGuide\Concept.mo ]
line 49 column 114 - Error: <nr> is not recognized!
```

Whenever you configure HTML Tidy to show warnings or information, it gets far too noisy. See the two log files with [warnings only](https://gist.github.com/beutlich/2643abf949768e01a6571622be95299e) and [both warnings and info](https://gist.github.com/beutlich/e4f360650035611f7b11056bea5b98c4) for example.

The proposed CI really is a fast, but poor-mans implementation of simple HTML tag validity checking according to our guidelines:
 * lower case tags
 * use `strong` and `em` tags (over `b` and `i` tags)
 * tags should be closed (though HTML might be more tolerant)

beutlich:
Hm, I believe we need to fix these warnings and really should run HTML Tidy with warnings disgnostics.

beutlich:
Doesn't look too bad if we mute the three HTML Tidy warnings `MISSING_ATTRIBUTE,REMOVED_HTML5,MISMATCHED_ATTRIBUTE_WARN` (requires HTML Tidy version 5.6). I'll care for the remaining HTML issues and update the CI job to also utilize HTMl Tidy.

beutlich:
@mtiller The HTML Tidy check is now also included. Thanks for your feedback which really pushed me (again) in the right direction.

PR #3282, opened 2019-12-21
#3280 pythagoras fix

### LINKED ISSUE #3280: Pythagoras block doesn't handle negative y2
With u1IsHypotenuse = true, Math.Blocks.Pythagoras doesn't properly handle the case where y2 goes negative. The code should use an if-expression (rather than an if-equation) wrapped in a noEvent to prevent evaluating the square root with a negative value. As it is, an event is created when y2 goes negative, but that doesn't prevent the problem.

JoeRiel:
My fix for #3280 (issue with Pythagoras block). Resolves #3280.

### DIFF
--- a/Modelica/Blocks/Math.mo
+++ b/Modelica/Blocks/Math.mo
@@ -3492,26 +3492,21 @@ so one can plot the result directly as frequency lines.
             thickness=0.5)}));
   end RealFFT;
 
-  block Pythagoras "Determines the hypotenuse or leg of a right triangle"
+  block Pythagoras "Determine the hypotenuse or leg of a right triangle"
     extends Interfaces.SI2SO;
-    parameter Boolean u1IsHypotenuse = false "If true, u1 is the hypotenuse and y is one leg";
-    Interfaces.BooleanOutput valid  "Is true, if y is a valid result" annotation (Placement(transformation(extent={{100,-70},{120,-50}})));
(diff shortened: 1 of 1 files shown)


Modelica/Blocks/Math.mo
```diff
@@ -3505,13 +3505,8 @@ so one can plot the result directly as frequency lines.
...
       y2 = u1^2-u2^2;
-      if y2>=0 then
-        y=sqrt(y2);
-        valid = true;
-      else
-        y=0;
-        valid = false;
-      end if;
+      valid = (y2 >= 0);  
+      y = noEvent(if y2 >= 0 then sqrt(y2) else 0);
```
beutlich: What about?

```
valid = noEvent(y2 >= 0);
y = if valid then sqrt(y2) else 0;
```

JoeRiel: That won't work, I believe, because valid is a discrete variable, so its equation must occur in an event.

JoeRiel: A nice variation would be
  `valid = y2 >= 0;
   y = noEvent(if valid then sqrt(y2) else 0);`
Alas, that doesn't work (at least not on the Dymola version I tried).  I didn't expect it to, but cannot say exactly why it should fail.

beutlich:
I force-pushed to your branch with a fix that no longer raises a floating-point exception in Dymola or SimulationX.

PR #3286, opened 2019-12-23
Add labeler for pull requests via GitHub Actions

beutlich:


mtiller:
Do I understand correctly that the purpose here is to add labels to commits to indicate which sub-packages have been modified by the commit?  I'm not quite sure what I'm supposed to review here given that I haven't used Github Actions (although I've used Gitlab CI which is quite similar).  In other words, what is the "question" to be answered here?

beutlich:
Yes, the idea is that with every new PR (from external contibutors) the labels (for that PR) are set according to the modified files via GitHub Actions.

PR #3285, opened 2019-12-22
Some small fixes to Batteries

AHaumer:
I had to fix a few typos as well broken links (didn't notice them during the last change regarding the parameters). Additionally I created a (sub-)icon indicating the usage of transient RC elements.

PR #3289, opened 2019-12-24
Batteries extensions

AHaumer:
As mentioned in #3285 I'v implemented some extensions to Batteries (based on #3285):
cells and stacks equipped with sensors
stacks built as Ns x Np matric of different cells
Thus, one is able to investigate the influence of a degraded cell on the bahviour of the whole stacks. as well as to design a battery management system.
For active cell balancing, the stack models have to get equipped additionally with appropriate actors.


casella:
I'll take care of this ASAP, I guess it will take some time to do a proper evaluation

beutlich:
I resolved the pseudo merge conflicts and rebased the branch.

christiankral:
I changed variable name `p` to `power` in 91702ff3a11793c1d0c097a3ad8911d61c9e5170 and c0ad18323d454836bcb93c02c9fdd8b68fd83d0b to be in line with the usual naming convention of the Electrical domain. 

@AHaumer I wonder if we shall also change the variable name `losses` to `lossPower` or `LossPower` to avoid an additional variable name for a loss term.

AHaumer:
@christiankral you're right with losses vs. lossPower

AHaumer:
@christiankral 
followed your advice and renamed losses to lossPower
resolved the bug introduced by renaming p to power (we have to correct the conntection to the bus variable, too - no error message since new bus variable p is introduced)

PR #3294, opened 2020-01-07
Normalize inputs to from_nxy and from_nxz

### LINKED ISSUE #3201: Need to normalize input vectors to Modelica.Mechanics.MultiBody.Parts.FixedRotation if vector length small
One of our customers reported this issue to us. They were creating a inboard anti-roll bar model from components in the Modelica standard library. To get the mechanism to initialise in the correct arrangement they needed to normalize the vector inputs, **n_x** and **n_y**, to the Modelica.Mechanics.MultiBody.Parts.FixedRotation named **blade_frame_b** in the mechanism. The customer believe that this is because the lengths of the vectors n_x and n_y are small (10^-4 and 10^-6) and this causes a problem in the function  Modelica.Mechanics.MultiBody.Frames.from_nxy which is used in the FixedRotation model unless the input vectors are normalized.  There is nothing in the documentation to say this could be an issue. 
Attached is the example which used MSL version 3.2.3, and these were simulated with Dymola 2020. 
[KinematicLoopTest_Issue.zip](https://github.com/modelica/ModelicaStandardLibrary/files/3808605/KinematicLoopTest_Issue.zip)
The experiment **Modelica_test** shows the results of the model where the vectors **n_x** and **n_y** of the component **blade_frame_b** are not normalized. If you look at the animation you will see that the two sides of the mechanism start in different position. 
Modelica_test

The experiment **Modelica_test_Normalize** shows the results when the vectors are normalized, in the animation both sides start in the same position. 
Modelica_test_Normalize

beutlich:
Resolves #3201

### DIFF
--- a/Modelica/Mechanics/MultiBody/Frames.mo
+++ b/Modelica/Mechanics/MultiBody/Frames.mo
@@ -3076,9 +3076,10 @@ Optionally, residues are returned as well if <code>withResidues=true</code>.
         "Orientation object to rotate frame 1 into frame 2";
     protected
       Real e_x[3](each final unit="1")=if length(n_x) < 1e-10 then {1,0,0} else normalize(n_x);
-      Real n_z_aux[3](each final unit="1")=cross(e_x, n_y);
-      Real n_y_aux[3](each final unit="1")=if n_z_aux*n_z_aux > 1.0e-6 then n_y
-             else (if abs(e_x[1]) > 1.0e-6 then {0,1,0} else {1,0,0});
+      Real e_y[3](each final unit="1")=if length(n_y) < 1e-10 then {0,1,0} else normalize(n_y);
+      Real n_z_aux[3](each final unit="1")=cross(e_x, e_y);
+      Real n_y_aux[3](each final unit="1")=if n_z_aux*n_z_aux > 1.0e-6 then e_y
+             else if abs(e_x[1]) > 1.0e-6 then {0,1,0} else {1,0,0};
       Real e_z_aux[3](each final unit="1")=cross(e_x, n_y_aux);
       Real e_z[3](each final unit="1")=normalize(e_z_aux);
     algorithm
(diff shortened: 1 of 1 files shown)


MartinOtter:
I changed the label from bug to enhancement, because this pull request improves the numerics (the original code is correct, but, as any other code, may fail due to numerical issues).

beutlich:
> 
> 
> I changed the label from bug to enhancement, because this pull request improves the numerics (the original code is correct, but, as any other code, may fail due to numerical issues).

OK. That implies that I won't back-port the changes to maint/3.2.3.

Modelica/Mechanics/MultiBody/Frames.mo
```diff
@@ -3076,9 +3076,10 @@ Optionally, residues are returned as well if <code>withResidues=true</code>.
         "Orientation object to rotate frame 1 into frame 2";
     protected
       Real e_x[3](each final unit="1")=if length(n_x) < 1e-10 then {1,0,0} else normalize(n_x);
-      Real n_z_aux[3](each final unit="1")=cross(e_x, n_y);
-      Real n_y_aux[3](each final unit="1")=if n_z_aux*n_z_aux > 1.0e-6 then n_y
-             else (if abs(e_x[1]) > 1.0e-6 then {0,1,0} else {1,0,0});
+      Real e_y[3](each final unit="1")=if length(n_y) < 1e-10 then {0,1,0} else normalize(n_y);
+      Real n_z_aux[3](each final unit="1")=cross(e_x, e_y);
+      Real n_y_aux[3](each final unit="1")=if n_z_aux*n_z_aux > 1.0e-6 then e_y
```
tobolar: @MartinOtter As the length of e_y is always 1, `if n_z_aux*n_z_aux > 1.0e-6` only assures the two input vectors aren't collinear now. Could we simply change it to `if length(n_z_aux) > 1e-3`?

Modelica/Mechanics/MultiBody/Frames.mo
```diff
@@ -3076,9 +3076,10 @@ Optionally, residues are returned as well if <code>withResidues=true</code>.
         "Orientation object to rotate frame 1 into frame 2";
     protected
       Real e_x[3](each final unit="1")=if length(n_x) < 1e-10 then {1,0,0} else normalize(n_x);
```
tobolar: The function normalize() uses eps=100*ModelicaServices.Machine.eps to avoid a division by zero. Theoretically, there could be a mess if such eps>1e-10. Could it be reasonable to better set
`if length(n_x) < 1e-10 then {1,0,0} else normalize(n_x, 1e-10)`
or
`if length(n_x) < 100*ModelicaServices.Machine.eps then {1,0,0} else normalize(n_x)`
?

dietmarw: in addition. Should not rather the higher level `Modelica.Constants.eps` be used instead of `ModelicaServices.Machine.eps`?

Modelica/Mechanics/MultiBody/Frames.mo
beutlich: n_z_aux*n_z_aux avoids the sqrt operation of function length. Though sqrt converges fast, I prefer the current notation.

Modelica/Mechanics/MultiBody/Frames.mo
beutlich: normalizes already uses Modelica.Constants.eps.

tobolar: > normalize already uses Modelica.Constants.eps

This is correct, sorry for the confusion. 

Actually, my question rather is whether there shall better be the same numerical value for the if-clause threshold and for "normalize" evaluation. (This should of course be a multiple of Modelica.Constants.eps)

Modelica/Mechanics/MultiBody/Frames.mo
tobolar: I understand. We could apply the same rule on `if length(n_x) < 1e-10 then ...` as well since the 'length' is not really needed there. But having a threshold of 1e-20 would probably lead to numerical problems.

Modelica/Mechanics/MultiBody/Frames.mo
MartinOtter: I would not change these thresholds now, because existing models might be influence in rare cases. Note, whatever threshold is used, there might be (very rare) cases where this heuristics is not good enough.

PR #3296, opened 2020-01-08
Avoid code duplication of functions g2, cond_dT and visc_dTp in Modelica.Media.Air.ReferenceMoistAir.Utilities.IF97_new

beutlich:
Avoid code duplication of functions g2, cond_dT and visc_dTp by introduction of short-class functions in Modelica.Media.Air.ReferenceMoistAir.Utilities.IF97_new.

I checked that all 7 example models of Modelica.Media.Examples.ReferenceAir still give identical results.

As reported by https://github.com/modelica/ModelicaStandardLibrary/issues/3060#issuecomment-571520891.

### DIFF
--- a/Modelica/Media/Air/ReferenceMoistAir.mo
+++ b/Modelica/Media/Air/ReferenceMoistAir.mo
@@ -1834,11 +1834,11 @@ package ReferenceMoistAir
       end rho_pT_der;
 
       function visc_dT = Modelica.Media.Water.IF97_Utilities.BaseIF97.Transport.visc_dTp(
-        p=0, phase=0, final checkLimits=false)
+        final p=0, final phase=0, final checkLimits=false)
         "Dynamic viscosity eta(d,T), industrial formulation";
 
       function cond_dT = Modelica.Media.Water.IF97_Utilities.BaseIF97.Transport.cond_dTp(
-        p=0, phase=0, final industrialMethod=true, final checkLimits=false)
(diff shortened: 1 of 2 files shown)


beutlich:
@MartinOtter I added you as reviewer since you added the IF97_new package by ba2f041349fe1333d2124bef5ccdcf68a033faa3.

HansOlsson:
Shouldn't it be "final p=0, final phase=0" as well as "final checkLimits=false"?

MartinOtter:
> 
> 
> @MartinOtter I added you as reviewer since you added the IF97_new package by [ba2f041](https://github.com/modelica/ModelicaStandardLibrary/commit/ba2f041349fe1333d2124bef5ccdcf68a033faa3).

I only included this package into the MSL. The package was developed by LTX Simulation GmbH. Therefore, I suggest that @wischhusen should approve this change

wischhusen:
Makes sense to me. I did not find obstacles, because the calculation is now equivalent to the state we had after the first release of ReferenceMoistAir.

beutlich:
> 
> 
> Makes sense to me. I did not find obstacles, because the calculation is now equivalent to the state we had after the first release of ReferenceMoistAir.

@wischhusen Can you approve (or reject) the PR then since we require a second review. Thanks.

beutlich:
Thanks, perfect timing 😃 

grafik

wischhusen:
Jippieh :-D

PR #3275, opened 2019-12-18
Fix two-phase calculation of setState_dTX for R134a

beutlich:
closes #3163

PR #3298, opened 2020-01-09
maint/3.2.3: Fix two-phase calculation of setState_dTX for R134a

### LINKED ISSUE #3163: Modelica.Media.R134a problem
**Reported by CTG on 15 Oct 2019 07:56 UTC**

Although it is not an OpenModelica related problem, I would like to share the following experience.
While checking my own media libraries compatible with Modelica.Media, I have found what I think is a bug in the Modelica.Media.R134a package. The calculation fail is shown in the attached file. Coming from a density between bubble and dew densities, it fails in the calculation of the vapor quality. The origin of the problem seems a bad implementation of the setState_dTX function: the calculation of Helmholtz derivatives and specific enthalpy is done before to check the phase fractions, as if the given density would be a one phase density. This produces a complete fail for calculations in the two phase region coming from density and temperature.
I do not know if this is a known bug, or I am wrong in some point.

[ModelicaTests.mo](https://github.com/modelica/ModelicaStandardLibrary/files/3750324/ModelicaTests.zip)

---

Originally reported at https://openmodelica.org/forum/default-topic/2801-modelica-media-r134a-problem#p9417

beutlich:
Back-port of #3275 to fix #3163 in maint/3.2.3.

### DIFF
--- a/Modelica/Media/R134a.mo
+++ b/Modelica/Media/R134a.mo
@@ -328,19 +328,27 @@ Example:
 
     algorithm
       R := R134aData.data.R;
-      f := f_R134a(d=d, T=T);
-      state.p := d*R*T*f.delta*f.fdelta;
-      state.h := R*T*(f.tau*f.ftau + f.delta*f.fdelta);
-      sat.psat := state.p;
-      state.T := T;
-      state.d := d;
-      sat.Tsat := state.T;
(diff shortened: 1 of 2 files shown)


wischhusen:
Many thanks Francesco!

PR #3307, opened 2020-01-12
Fix #3300 LeakageWithCoefficient

### LINKED ISSUE #3300: LeakageWithCoefficient should have a port for reluctance input
The model Modelica.Magnetic.FluxTubes.Basic.LeakageWithCoefficient has an input signal, R_mUsefulTot, that provides the total reluctance of the path in which it is connected. To make this part usable in a GUI, that input should be a port to which a real signal can be connected. 

Actually, it's not clear to me why this quantity is an input and not a parameter. Are there examples of a model where it varies?  The doc-string for the component ("... not for dynamic simulation of actuators") suggests it isn't intended to be used in that application.

AHaumer:
fixes #3300

### DIFF
--- a/Modelica/Magnetic/FluxTubes/Basic/LeakageWithCoefficient.mo
+++ b/Modelica/Magnetic/FluxTubes/Basic/LeakageWithCoefficient.mo
@@ -1,25 +1,25 @@
 within Modelica.Magnetic.FluxTubes.Basic;
 model LeakageWithCoefficient
   "Leakage reluctance with respect to the reluctance of a useful flux path (not for dynamic simulation of actuators)"
-
   extends BaseClasses.Leakage;
-
-  parameter SI.CouplingCoefficient c_usefulFlux=0.7
+  import Modelica.Constants.eps;
+  parameter SI.CouplingCoefficient c_usefulFlux(final min=eps, final max=1-eps)
     "Ratio useful flux/(leakage flux + useful flux) = useful flux/total flux";
-
-  input SI.Reluctance R_mUsefulTot
-    "Total reluctance of useful flux path as reference" annotation (Dialog(
-        group="Reference reluctance", groupImage=
-          "modelica://Modelica/Resources/Images/Magnetic/FluxTubes/Basic/LeakageWithCoefficient.png"));
-
+  Blocks.Interfaces.RealInput R_mUsefulTot(quantity="Reluctance", unit="H-1")
+    annotation (Placement(transformation(
+        extent={{-20,-20},{20,20}},
+        rotation=270,
+        origin={0,120})));
 equation
   (1 - c_usefulFlux)*R_m = c_usefulFlux*R_mUsefulTot;
   // Generalized Kirchhoff's current law
-
   annotation (defaultComponentName="leakage", Documentation(info="<html>
 <p>
 Differently from the flux tube elements of package <a href=\"modelica://Modelica.Magnetic.FluxTubes.Shapes.Leakage\">Shapes.Leakage</a>
 that are calculated from their geometry, this leakage reluctance is calculated with reference to the total reluctance of a useful flux path. Please refer to the <strong>Parameters</strong> section for an illustration of the resulting magnetic network. Exploiting <em>Kirchhoff</em>'s generalized current law, the leakage reluctance is calculated by means of a coupling coefficient c_usefulFlux.
+</p>
+<p>
+<img src=\"modelica://Modelica/Resources/Images/Magnetic/FluxTubes/Basic/LeakageWithCoefficient.png\" alt=\"Leakage with coefficient\">
 </p>
 
 <h4>Attention:</h4>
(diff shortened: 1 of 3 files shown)


Modelica/Magnetic/FluxTubes/Basic/LeakageWithCoefficient.mo
```diff
@@ -1,26 +1,26 @@
 within Modelica.Magnetic.FluxTubes.Basic;
 model LeakageWithCoefficient
   "Leakage reluctance with respect to the reluctance of a useful flux path (not for dynamic simulation of actuators)"
-
   extends BaseClasses.Leakage;
-
-  parameter SI.CouplingCoefficient c_usefulFlux=0.7
+  import Modelica.Constants.eps;
+  parameter SI.CouplingCoefficient c_usefulFlux(final min=eps, final max=1-eps)
     "Ratio useful flux/(leakage flux + useful flux) = useful flux/total flux";
```
beutlich: I believe it is not a good idea to remove the default value here.

Modelica/Magnetic/FluxTubes/Examples/BasicExamples/QuadraticCoreAirgap.mo
```diff
@@ -57,59 +57,75 @@ parameter Integer N=500 "Number of turns of exciting coil";
...
   connect(electricGround2.p, voltageSensor.n)
-    annotation (Line(points={{50,-10},{60,-10}}, color={0,0,255}));
+    annotation (Line(points={{60,-10},{70,-10}}, color={0,0,255}));
   connect(excitingCoil.port_n, magFluxSensor.port_n)
-    annotation (Line(points={{-20,-10},{-20,-20}}, color={255,127,0}));
+    annotation (Line(points={{-30,-10},{-30,-20}}, color={255,127,0}));
   connect(magFluxSensor.port_p, magneticGround.port)
-    annotation (Line(points={{-20,-40},{-20,-50}}, color={255,127,0}));
+    annotation (Line(points={{-30,-40},{-30,-50}}, color={255,127,0}));
+  connect(leakage.port_n, airGap.port_n) annotation (Line(points={{-1.77636e-15,
```
beutlich:
```suggestion
  connect(leakage.port_n, airGap.port_n) annotation (Line(points={{0,
```

Modelica/Magnetic/FluxTubes/Examples/BasicExamples/QuadraticCoreAirgap.mo
```diff
@@ -57,59 +57,75 @@ parameter Integer N=500 "Number of turns of exciting coil";
...
+          -10},{0,-10},{0,-40},{30,-40}}, color={255,127,0}));
+  connect(measuringCoil.port_p, leakage.port_p)
+    annotation (Line(points={{30,10},{1.77636e-15,10}}, color={255,127,0}));
```
beutlich:
```suggestion
    annotation (Line(points={{30,10},{0,10}}, color={255,127,0}));
```

AHaumer:
@christiankral 
In fact the component only models the leakage, not the main path - but it depends on the main path. 
I'm not sure what was the intention of Thomas Bödrich - this component has not been used in the examples ... I could guess: model a leakage defined by a fixed stray factor dependent on a variable main reluctance.

JoeRiel:
While I'm happy with the change, it's still not clear how this would be usefully employed. An example would be helpful. How would the varying reluctance of the main path be modeled?

AHaumer:
@JoeRiel 
Simply with a shape with nonlinear material characteristic

JoeRiel:
Maybe I'm missing something obvious, but how does one generate the reluctance, so it can be passed to a LeakageWithCoefficient block, for a shape with a nonlinear magnetic characteristic?

christiankral:
> @christiankral
> In fact the component only models the leakage, not the main path - but it depends on the main path.
> I'm not sure what was the intention of @ThomasBoedrich - this component has not been used in the examples ... I could guess: model a leakage defined by a fixed stray factor dependent on a variable main reluctance.

Yes, you're right. It is just a model of the leakage path.

AHaumer:
@JoeRiel 
The shapes calculate the reluctance resp. permeance dependent on the actual point of operation. Have a look at the Example `BasicExamples.QuadraticCoreAirgap` of my PR.

JoeRiel:
Thanks, Anton, for that example. That also suggests why R_usefulTotal was originally not designed as a port; it has to come from an internal signal.

christiankral:
It makes sense to include variable reluctance and permeance models, too, if we provide a signal input for `LeakageWithCoefficient`: see #3311

AHaumer:
@JoeRiel 
One could provide a signal output at the shapes for that purpose. Looks a little bit ugly ...

christiankral:
I guess we are ready to merge now.

PR #3310, opened 2020-01-13
Fix HTML Tidy warning "MISMATCHED_ATTRIBUTE_WARN" in documentation 

beutlich:
See #3060.


dietmarw:
I'm also a bit on the fence. Unless we specifically require HTML5(+) only we should probably just ignore the HTML4 warnings.

beutlich:
I can live with that feedback. No need to keep it open.

Issue #3276, opened 2019-12-18
Possible Modelica.Media bug


beutlich:
**Reported by CTG on 14 Dec 2019 19:53 UTC**

While checking my media models compatibility against ThermoPower, I observed a too high difference in the results obtained in the TestRefrigeranEvaporator test model, using the standard Modelica.Media.R134a.R134a_ph medium and my own medium. I arrived to the conclusion that it was due to the physical properties calculation and checked it. The result I obtain with the Modelica medium seems incorrect regarding density, as you can check with the attached model. The calculation done at 25ºC and 30 bars gives a liquid density result much lower than the saturated one, what seems very strange. I checked the calculation with CoolProp and it seems that the Modelica calculation is erroneous. The calculation at 25ºC is OK till 18 bars.
Can someone check this?

[ModelicaTest5.zip](https://github.com/modelica/ModelicaStandardLibrary/files/3980401/ModelicaTest5.zip)

---

Originally reported at https://openmodelica.org/forum/default-topic/2841-possible-modelica-media-bug#p9560

beutlich:
**Reported by CTG on 15 Dec 2019 20:02 UTC**

Well, I think I have found the origin of the problem. The medium is giving negative values for  isothermal compressibility, when they should be possitive. Reviewing the code it seems that a term is missing inside the brackets of the kappa calculation.
I hope not to be wrong and that this can help.

---

Originally reported at https://openmodelica.org/forum/default-topic/2841-possible-modelica-media-bug#p9561

beutlich:
@wischhusen Can you please analyze? Thank you.

beutlich:
@thorade Can you please have a look. I guess he is referring to https://github.com/modelica/ModelicaStandardLibrary/blob/4b560e50519d7b80036550fbbe380f6e1396a00f/Modelica/Media/R134a.mo#L1629-L1630 in Modelica.Media.R134a.R134a_ph.isothermalCompressibility, which even is repeated at https://github.com/modelica/ModelicaStandardLibrary/blob/4b560e50519d7b80036550fbbe380f6e1396a00f/Modelica/Media/R134a.mo#L183

thorade:
I believe the analysis is correct and the kappa formula should be:
```
kappa := 1/(state.d*R134aData.data.R_s*state.T*f.delta*(1.0/f.delta + 2.0*f.fdelta + f.delta*f.fdeltadelta)); 
```
or equivalent
```
[...]

PR #3317, opened 2020-01-14
Fix Modelica.Media.R134a.R134a_ph.setState_pTX (only applicable for one-phase)

beutlich:
The property tables, especially for setState_pTX, created by [T3276.mo](https://gist.github.com/beutlich/ccd3505920466e99d0446b1f6d6f2b78) at least look as expected now (compared to https://github.com/modelica/ModelicaStandardLibrary/issues/3276#issuecomment-574316755):

**setState_phX()** (ref props in parentheses)

| row | T | p | d | h |
|---|---|---|---|---|
| 1 | 374.21 | 4.05911e+06 | 511.899<br>(511.9) | 389639<br>(389639) |
| 2 | 169.85 | 389.563 | 1517.14<br>(1591.11) | 71455.2<br>(71455.2) |
| 3 | 298.15 | 3e+06 | 1220.13<br>(1220.13) | 234687<br>(234687) |
| 4 | 305.987 | 3e+06 | 1190.76<br>(1190.76) | 245776<br>(245776) |
| 5 | 273.15 | 292803 | 1294.76<br>(1294.78) | 200000<br>(200000) |
| 6 | 298.15 | 665380 | 969.978<br>(969.916) | 235741<br>(235741) |

**setState_pTX()** (ref props in parentheses)

| row | T | p | d | h |
|---|---|---|---|---|
| 1 | 374.21 | 4.05911e+06 | 501.2<br>(511.9) | 390934<br>(389639) |
| 2 | 169.85 | 389.563 | n.a.<br>(1591.11) | n.a.<br>(71455.2) |
| 3 | 298.15 | 3e+06 | :heavy_check_mark: 1220.13<br>(1220.13) | :heavy_check_mark: 234687<br>(234687) |
| 4 | 305.987 | 3e+06 | :heavy_check_mark: 1190.76<br>(1190.76) | :heavy_check_mark: 245776<br>(245776) |
| 5 | 273.15 | 292803 | n.a.<br>(1294.78) | n.a.<br>(200000) |
| 6 | 298.15 | 665380 | n.a.<br>(969.916) | n.a.<br>(235741) |

Except when looking at the results, I am not sure if this derivative-free appoach is appropriate at all or if the limits used for the specific enthalpy 71455.2, 400000 are valid choices.

This issue with isentropic exponent was not addressed, currently.

### DIFF
--- a/Modelica/Media/R134a.mo
+++ b/Modelica/Media/R134a.mo
@@ -361,7 +361,7 @@ Example:
 parameter Medium.Density d = 4;
 parameter Medium.Temperature T = 298;
 
-Medium.SpecficEntropy s;
+Medium.SpecificEntropy s;
 
 <strong>equation</strong>
 
@@ -408,9 +408,9 @@ Example:
 </p>
 <blockquote><pre>
 parameter Medium.AbsolutePressure p = 3e5;
-parameter Medium.SpecficEntropy s = 1.7e3;
+parameter Medium.SpecificEntropy s = 1.7e3;
 
-Medium.SpecficEnthalpy h;
+Medium.SpecificEnthalpy h;
 
 <strong>equation</strong>
 
@@ -423,16 +423,15 @@ h = Medium.specificEnthalpy(setState_psX(p, s, fill(0, Medium.nX)));
       "Set state for pressure and temperature (X not used since single substance)"
 
     protected
-        Modelica.SIunits.Pressure delp=1.0e-2
-        "Relative error in p in iteration";
+      SpecificEnthalpy h = hofpT(p, T) "Specific enthalpy";
 
     algorithm
       Modelica.Media.R134a.R134a_ph.phaseBoundaryAssert(p, T);
 
       state := ThermodynamicState(
-              d=dofpT(p, T, delp),
+              d=density_ph(p, h),
               T=T,
-              h=hofpT(p, T, delp),
+              h=h,
               p=p,
               phase=1);
       annotation (Inline=true, Documentation(info="<html>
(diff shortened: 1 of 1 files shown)


beutlich:
> Can you please summarize what is the improvement brought by this PR?

I marked the relevant changes by :heavy_check_mark: in the above table. Without this patch, i.e., with the implemented Newton methode of dofpT, the values are calculated incorrectly.

beutlich:
Would be good if the enthalpy limits for hofpT are cross-checked, since the limits are not set in the medium.

https://github.com/modelica/ModelicaStandardLibrary/blob/4864a3a7c86aa91fb3c251ba730bb9c7816bbf6e/Modelica/Media/R134a.mo#L215-L216

wischhusen:
I checked the modifications in the Diff but I didn't get the meaning of:

y := temperature_ph(p=p, h=u) - T;

Is u the specific internal energy or something else. I would have expected h instead of u ...?

beutlich:
`u` is the input of Modelica.Math.Nonlinear.Interfaces.partialScalarFunction. `y` is its output.

thorade:
If I read the PR correctly, you are suggesting to call density_ph from density_pT. That would be the wrong way to go. The underlying equation of state is explicit in d and T, p and h are calculated as output. So pT is a 1D iteration (one unknown: density), but ph is a 2D Newton (two unknowns, d and T).

The required change here would be instead to improve the 1D pT Newton iteration: As it is a 1D root finding, it can be improved to check bounds, and it can be combined with bisection to update the bounds after each Newton step. Bounds means: density values should be outside the saturated/two-phase region density values, not sure whether you want to use spline values (plus some percentage?) to check the bounds, or a full equilibrium calculation based on the full EoS.

beutlich:
@thorade Thanks for the detailed feeback. I already had the feeling that the change was too simple and only created this PR when I surprisingly obtained the correct results. Unfortunately we have no test models for setState_pTX and setState_psX.

There is no need to keep this PR longer. I'll let the fix to the library officers. If the fix is too tough ot too late, I am thinking to visually mark setState_pTX and setState_psX (e.g., by inheritance from Modelica.Icons.UnderConstruction or some other icon) and update the documention accordingly.

PR #3303, opened 2020-01-11
Remove obsolete drawings of diagram layer

christiankral:
Incorporating #3056

In order to review this PR it makes sense to have a look at the diagram layers of models in the following packages:
- Modelica.Blocks.Continuous
- Modelica.Blocks.Discrete
- Modelica.Blocks.Continuous
- Modelica.Blocks.Discrete
- Modelica.Blocks.Interfaces
- Modelica.Blocks.Logical
- Modelica.Blocks.Math
- Modelica.Blocks.Math
- Modelica.Blocks.Nonlinear
- Modelica.Blocks.Routing
- Modelica.Blocks.Sources
- Modelica.Blocks.Tables
- Modelica.ComplexBlocks
- Modelica.Magnetic.FluxTubes.Basic
- Modelica.Magnetic.FluxTubes.Sources
- Modelica.Magnetic.FundamentalWave.Components
- Modelica.Magnetic.QuasiStatic.FundamentalWave.Components
- Modelica.Thermal.HeatTransfer.Components
- Modelica.Thermal.HeatTransfer.Sources
- Modelica.Thermal.HeatTransfer.Components
- Modelica.Thermal.HeatTransfer.Celsius
- Modelica.Thermal.HeatTransfer.Fahrenheit
- Modelica.Thermal.HeatTransfer.Rankine
- Modelica.Mechanics.Rotational.Components
- Modelica.Mechanics.Rotational.Sources
- Modelica.Mechanics.Translational.Components
- Modelica.Thermal.FluidHeatFlow.BaseClasses
- Modelica.Math
- Modelica.Electrical.Analog.Lines
- Modelica.Electrical.Analog.Semiconductors.Thyristor
- Modelica.Electrical.Analog.Sources

PR #3318, opened 2020-01-14
Split Mechanics.Rotational into single files

AHaumer:
refs #2975
see also #3319


AHaumer:
As far as I see there are no open PRs for Rotational (and Translational)

beutlich:
> 
> 
> As far as I see there are no open PRs for Rotational (and Translational)

Except #3310.

dietmarw:
This needs to be reverted because of #3303

PR #3319, opened 2020-01-14
Split Mechanics.Translational into single files

AHaumer:
refs #2975
see also #3318


dietmarw:
This needs to be reverted because of #3303

Issue #2695, opened 2018-08-01
Storage of PNG files (figures)


tobolar:
There is a convention concerning figures' storage which states that
> The PNG files should be placed in a folder which exactly represents the package structure. 

(see also Modelica.UsersGuide.Conventions.Documentation.Format.Figures).

According to this, the figures collected in `Modelica/Resources/Images/UsersGuide` should be stored appropriately. This is already the case for figures used in Modelica.UsersGuide.Conventions, i.e. there is a directory `Modelica/Resources/Images/UsersGuide/Conventions` containing corresponding subdirectories.

tobolar:
@AHaumer @christiankral I'm going to proceed here with Electrical and Magnetic in the next days. In order to omit conflicts please let me know when I could start. Or you can of course fix this issue for those two packages on your own.

dietmarw:
As far as I can tell there are no more open PRs left that would conflict with changes to be done in Electrical and Magnetic

tobolar:
You are right. But working on this issue is really time consuming and I will not find time to fix this till MSL 4.0.0 fix.

dietmarw:
Yes. Hopefully @christiankral and @AHaumer can take care of those.

christiankral:
OK then. @AHaumer I guess we should always indicate in advance what packages we are working on to avoid redundant work.

beutlich:
> 
> 
> You are right. But working on this issue is really time consuming and I will not find time to fix this till MSL 4.0.0 fix.

Remvong milestone.

PR #3322, opened 2020-01-15
Fix figures location for Fluid

tobolar:
refs #2695


tobolar:
@beutlich 
> Should Modelica/Resources/Images/Fluid/Dissipation/heatTransfer/ be renamed to Modelica/Resources/Images/Fluid/Dissipation/HeatTransfer/ (and similar for all other occurrences)?

Yes and no. This PR concerns only PNG files used in the **class documentation**. But the images in "heatTransfer" directory are used only in icons. So yes, it should be renamed, and no, not in this PR (IMO).
Actually, there are far more images left in the directories of Fluid which aren't used anyway. I will open separate issue for this.

tobolar:
@casella 
> Did you check with at least one tool that everything shows up correctly?

I have tried it. And have also checked that all images used in the documentation exist in the /Modelica/Resources/Images/Fluid directory.

PR #3332, opened 2020-01-16
Remove default diagram size annotations Rotational and Translational

dietmarw:
Using ttws


dietmarw:
@AHaumer this is already on the splitted version. @christiankral took care of the re-splitting today.

PR #3334, opened 2020-01-16
Split ComplexBlocks into single files

christiankral:
Refs #2975


dietmarw:
@christiankral This split has to be redone too since #3047 was modifying the original single file package.

beutlich:
> dietmarw closed this 2 hours ago

Usually there is no need to close since force-pushing to the same branch updates the pull request.

dietmarw:
You are right but thought it would be easier for the LOs

Issue #3335, opened 2020-01-16
Modelica.Electrical.QuasiStatic.Polyphase.Blocks.ToSpacePhasor fails Dymola check


beutlich:
@christiankral Can you please check why Modelica.Electrical.QuasiStatic.Polyphase.Blocks.ToSpacePhasor does no longer check in Dymola 2020? I noticed that you tried some alternative implementations to please OpenModelica.

```
Checking block Modelica.Electrical.QuasiStatic.Polyphase.Blocks.ToSpacePhasor:

Function input had incorrect type in the equation
c.re = 1.4142135623730951/m*(Modelica.ComplexMath.sum(u[k].re*cos(phi[k])-u[k]. im*sin(phi[k]) for k in 1:m));

Function input had incorrect type in the equation
c.im = 1.4142135623730951/m*(Modelica.ComplexMath.sum(u[k].re*sin(phi[k])+u[k]. im*cos(phi[k]) for k in 1:m));

Missing declarations.

Check aborted.
```

beutlich:
The original implementation still works

```mo
c = sqrt(2)/m*sum({u[k]*exp(j*phi[k]) for k in 1:m});
```

PR #3337, opened 2020-01-16
Clarify #3335

christiankral:
I added an additional comment to better understand where the import problem (conflict) came from

### DIFF
--- a/Modelica/Electrical/QuasiStatic/Polyphase/Blocks/ToSpacePhasor.mo
+++ b/Modelica/Electrical/QuasiStatic/Polyphase/Blocks/ToSpacePhasor.mo
@@ -4,6 +4,8 @@ block ToSpacePhasor "Conversion: m phase -> space phasor"
   // import Modelica.ComplexMath.j;
   // import Modelica.ComplexMath.exp;
   // import Modelica.ComplexMath.sum;
+  // The import of sum goes along with the original complex equation:
+  // c = sqrt(2)/m*sum({u[k]*exp(j*phi[k]) for k in 1:m});
   parameter Integer m(min=1) = 3 "Number of phases";
   Modelica.ComplexBlocks.Interfaces.ComplexInput u[m]
     annotation (Placement(transformation(extent={{-140,-20},{-100,20}})));


dietmarw:
What are those commented imports for anyway? Whenever I see commented code it looks like unfinished WIP models which probably should not be inside the MSL (ideally)

beutlich:
> 
> 
> What are those commented imports for anyway? Whenever I see commented code it looks like unfinished WIP models which probably should not be inside the MSL (ideally)

Yes, may be appropriate to remove this source code comment at all.

christiankral:
> > What are those commented imports for anyway? Whenever I see commented code it looks like unfinished WIP models which probably should not be inside the MSL (ideally)
> 
> Yes, may be appropriate to remove this source code comment at all.

In this particular case I would keep the commented code, as the original code was re-written due to a OpenModelica issue. Further below in this model:

```modelica
  // c = sqrt(2)/m*sum({u[k]*exp(j*phi[k]) for k in 1:m});
  // Alternative implementation due to https://trac.openmodelica.org/OpenModelica/ticket/4496
  c.re = sqrt(2)/m*sum(u[k].re*cos(phi[k])-u[k].im*sin(phi[k]) for k in 1:m);
  c.im = sqrt(2)/m*sum(u[k].re*sin(phi[k])+u[k].im*cos(phi[k]) for k in 1:m);
```

I guess one day we will switch back to the original code; so the comment is the reminder on what the original code was.

Modelica/Electrical/QuasiStatic/Polyphase/Blocks/ToSpacePhasor.mo
```diff
@@ -4,6 +4,8 @@ block ToSpacePhasor "Conversion: m phase -> space phasor"
   // import Modelica.ComplexMath.j;
   // import Modelica.ComplexMath.exp;
   // import Modelica.ComplexMath.sum;
```
dietmarw: But these can be removed, right?

beutlich: No, as they belong to the commented eq. `c = sqrt(2)/m*sum({u[k]*exp(j*phi[k]) for k in 1:m});`.

dietmarw: ah no I see now it's just for the particular equation. Never mind.

PR #3340, opened 2020-01-17
Fix broken URIs in the documentation

dietmarw:
Most could be fixed automatically but there are some remaining one which need to be looked at manually. Ticket coming up.


dietmarw:
See #3341 for remaining issues.

PR #3347, opened 2020-01-18
Rebuild binaries

beutlich:
* MSVC VS 2005
* GCC (TDM64-1) 5.1.0
* Linux GCC 4.8.5


dietmarw:
Any advice how to test those. Well could do for Linux 64 port only anyway.

PR #3348, opened 2020-01-18
Add comparison signals of examples and test models

beutlich:
The relevant signals of the examples and test models should be part of the repository and maintained along with the models. **They will not be part of the distributed Modelica package though (as configured in .gitattributes)**.


Modelica/Resources/Reference/Modelica/Mechanics/MultiBody/Examples/Constraints/PrismaticConstraint/comparisonSignals.txt
```diff
@@ -0,0 +1,11 @@
+time
+bodyOfConstraint.v_0[1]
+bodyOfConstraint.v_0[2]
+bodyOfConstraint.v_0[3]
+constraint.frame_b.r_0[1]
```
tobolar: Instead of `constraint.frame_b.r_0[1:3]` we can use `bodyOfConstraint.r_0[1:3]`.

henrikt-ma:
A comment about the general idea of having the comparison signals in the repository: I like the idea of putting library officers in charge of maintaining the selection of comparison signals, and believe that the selection then needs to be in the repository to make this feasible.

Having _comparisonSignals.txt_ files is probably the best way to start, but that's just because we still don't have figure annotations (we will start collecting reviews on the MCP any day now).  With figure annotations, I would argue that those should be used as basis for comparison signals to ensure that we only compare signals that someone is (or at least have) actually looking at.  If comparison signals will be determined this way in the end, it could mean that the era of _comparisonSignals.txt_ in the repository becomes a short one, but having those files in the repository would likely also a be great way to start the transition to figure-based selection.

(Regarding @casella's recent complaint about [Magnetic.FundamentalWave.Examples.BasicMachines.AIMC_withLosses.diff.aimc.stator.zeroInductor.i0](https://libraries.openmodelica.org/branches/newInst/Modelica_3.2.3/files/Modelica_3.2.3_Modelica.Magnetic.FundamentalWave.Examples.BasicMachines.AIMC_withLosses.diff.aimc.stator.zeroInductor.i0.html): in case someone wants to have such a signal in a figure, we can simply use a vendor annotation such as {{__MA_comparisonSignal=false}} to exclude it from the set of comparison signals.)

tobolar:
It could be helpful if the files contain a comment on model name or even complete URI.

beutlich:
> 
> 
> It could be helpful if the files contain a comment on model name or even complete URI.

This would impede the copying of comparisonSignals files between similar models.

tobolar:
@beutlich 
What is the idea behind the "relevant signals"? I suppose you added in particular model states in this first draft, so there shall yet be enough information.
So my question is how shall I focus my review?

beutlich:
My original motivation (24 days ago) was that library officers create the comparisonSignals.txt files of the 146 newly introduced example and test models of MSL v4.0.0. This plan did not work out and I created all comparisonSignals.txt files on my own. :sweat:

Now, it would be good to get approval if the general idea of having teh comparisonSignals.txt files within the MSL resources is a good idea.

tobolar:
> if the general idea of having the comparisonSignals.txt files within the MSL resources is a good idea

I guess https://github.com/modelica/ModelicaStandardLibrary/pull/3348#pullrequestreview-353734146 intends right opposite idea.

IMO it is reasonable to define/select signals for comparison. But I haven't sufficient skills to evaluate eglibililty of the suggested idea introducing comparisonSignals.txt

beutlich:
> I guess [#3348 (review)](https://github.com/modelica/ModelicaStandardLibrary/pull/3348#pullrequestreview-353734146) intends right opposite idea.

Yes, that was after I finished adding the comparisonSignals.txt files.

Meanwhile, this PR got the two required approvals.

casella:
For the record, I'm fine with library officers being in charge of the maintenance. On the other hand, creating 146 such files from scratch is a lot of work, I think this should not be carried out by volunteers.

tobolar:
> On the other hand, creating 146 such files from scratch is a lot of work,...

I don't believe they all were generated from scratch: https://github.com/modelica/MAP-LIB_ReferenceResults

beutlich:
> 
> 
> > On the other hand, creating 146 such files from scratch is a lot of work,...
> 
> I don't believe they all were generated from scratch: https://github.com/modelica/MAP-LIB_ReferenceResults

Well, I created the 146 new comparisonSignals.txt from scratch. That was one reason, why I originally wanted to distribute the work load.

PR #3351, opened 2020-01-20
Rename sensor outputs

### LINKED ISSUE #3022: Unify names of sensor signal outputs
In the different domains we have different names of the signal output connectors. For example, the voltage sensor:

Sensor | Output connector name | Comment
---------- | ------------------------------ | --------------
`Analog.Sensors.VoltageSensor` | `v` | Instantaneous voltage, lower case  
`Machines.Sensors.VoltageQuasiRMSSensor` | `V` | Quasi RMS voltage, upper case
`QuasiStatic.SinglePhase.Sensors.VoltageSensor` | `y` | Complex quasi static
`QuasiStatic.MultiPhase.Sensors.VoltageSensor` | `y[m]` | Complex qusai static vector
`QuasiStatic.MultiPhase.Sensors.VoltageQuasiRMSSensor` | `V` | Complex quasi RMS voltage, upper case 

So this is a very non-uniform look. Change proposals
1. Change `y` to `v` and `y[m]` to `v[m]`
   * Pro: a voltage is clearly indicated as voltage `v`, characteristic voltage quantities are indicated by upper case `V`
   * Con: Partial sensors with `RealOutput y` cannot be re-used
2. Change every output to `y`
    * Pro: Clear rule: `y` is always the output of any sensor
    * Con: From the connector output name it cannot be told what kind of quantity it is

christiankral:
This PR resolves #3022. The following changes are applied both in `QuasiStatic.SinglePhase` and `QuasiStatic.Polyphase`:

- Rename all sensor outputs to match their physical quantity (e.g., `i` for current sensor)
- Introduce new partial models `TwoPinElementary` with connectors but without declaration of voltage and current
- Introduce new partial model `RelativeSensorElementary` to be used with voltage and current sensor(s only)
- Move obsolete partial models `RelativeSensor` to `ObsoleteModelica4.mo` as these models have only been used for voltage and current sensors; so there is no need to keep them
- Update models of MSL on which the changed sensor outputs have an impact
- Create conversion script
- Create test models in `ModelicaTestConversion4.mo`


beutlich:
I see potential merge conflicts: Renaming sensor outputs will likely require to update the comparisonSignals.txt files as part of #3348, too. Thus, whichever PR gets merged first ...

Modelica/Electrical/QuasiStatic/Polyphase/Sensors/CurrentQuasiRMSSensor.mo
```diff
@@ -20,32 +20,30 @@ model CurrentQuasiRMSSensor
...
-        Line(points={{-90,0},{-70,0}}, color={0,0,255}),
-        Line(points={{70,0},{90,0}}, color={0,0,255}),
         Line(points={{0,-70},{0,-100}}, color={85,170,255}),
         Text(
           extent={{150,-100},{-150,-70}},
           textString="m=%m"),
-        Text(
-          textColor={0,0,255},
-          extent={{-150,80},{150,120}},
-          textString="%name"),
```
dietmarw: Why is the %name removed here?

dietmarw: Never mind.

Modelica/Resources/Scripts/Conversion/ConvertModelica_from_3.2.3_to_4.0.0.mos
```diff
@@ -1826,3 +1826,35 @@ convertElement("Modelica.Media.Interfaces.PartialSimpleMedium.BaseProperties", "
 convertElement("Modelica.Media.Interfaces.PartialTwoPhaseMedium.BaseProperties", "R", "R_s");
 convertElement("Modelica.Media.R134a.R134a_ph.BaseProperties", "R", "R_s");
 convertElement("Modelica.Media.Water.WaterIF97_base.BaseProperties", "R", "R_s");
+
```
beutlich: I propose to put these lines before line 1795. i.e., before the automatically generated conversions.

Issue #3354, opened 2020-01-20
Update icons of QuasiStatic.SinglePhase and QuasiStatic.Polyphase


christiankral:
Some of the icons of `QuasiStatic.SinglePhase` and `QuasiStatic.Polyphase` are not in line with guidelines and shall be updated. In order to perform such update properly, #3351 has to be merged prior to a PR.

dietmarw:
#3351 is merged.

PR #3358, opened 2020-01-21
Fix icons of electrical sensors

christiankral:
Refs #3354 and extends it to transient electrical sensors. This PR fixes

- size of `%name`
- match color of connectors and connection lines of icon
- unify appearance of electrical flow sensors (current, flux) with one solid line from the left to right connector, indicating the there is no potential difference from one connector to the other (this was implemented for some electrical sensors already, and is unified now)

image


beutlich:
Oops, why do I see the comparisonSignals.txt of #3348 here?

christiankral:
Bad news :-( 

I created the branch not from master apparently. Definitely my fault. @beutlich Do you see a way to fix this issue in an elegant way or do I have to start over and re-apply all the changes manually.

beutlich:
Hang on. Can do.

christiankral:
Would cherry-picking help?

christiankral:
Thanks

beutlich:
Yes, that tis the way to go.

beutlich:
Fixed.

christiankral:
Thanks a lot!

PR #3388, opened 2020-02-03
Incorporate the new Modelica Libraries logo

dietmarw:
This adds the new official logo to the documentation ~~and also modifies the package icon.~~ 

Now there are two things I'd like some feedback first:
**UPDATE: Item 1 is no longer part of this PR**
1. Package logo 
This looks currently like this:
* OM
image
 * Dymola:
image
and after the update would look like this:
 * OM:
 image
 * Dymola: 
image

So the new logo has the now correct colour but might not be as distinctive. 

2. Use of SVG as logo. For now I've only included the SVG version as both Dymola and OMEdit manage to display them. Dymola 2020 does not seem to support the `style="width:250"` part (simply ignores it) but this might perhaps not be an issue by the time Dymola ships MSL 4.0.0 (i.e., when they fixed the HTML rendering). Feedback from other tool vendors is welcome.


Modelica/package.mo
```diff
@@ -8698,25 +8698,18 @@ revisionId="$Format:%h %ci$",
...
-Modelica Association, see
-<a href=\"https://www.Modelica.org\">https://www.Modelica.org</a>.
+<img src=\"modelica://Modelica/Resources/Images/Logos/Modelica_Libraries.svg\" style=\"width:250\">
+</p>
+
+<p>
+The package <strong>Modelica&reg;</strong> is a <strong>standardized</strong> and <strong>free</strong> package
+that is developed by the \"<strong>Modelica Association Project - Libraries</strong>\".</p>
+<p>
+Its development is coordinatted with with the Modelica&reg; language from the
```
maltelenz: I think this should be `coordinated with the`, with a single `t` and a single `with`

svorkoetter:
I'll check if it works in our browser. From the pictures posted so far though, the old one definitely looks better at that size. I've found that when resizing any sort of image to thumbnail size, it's often necessary to increase the contrast to maintain the same visual impact.

dietmarw:
Yes. I am a bit on the fence about the icon myself. On the one side the **official** colour is that grey but it too find it harder to recognise. Hence I'm thankful for the feedback.

Modelica/package.mo
dietmarw: Fixed in 1b4507b6e051fabf62d0f738c30547e3256ffecb

HansOlsson:
>  For now I've only included the SVG version as both Dymola and OMEdit manage to display them. Dymola 2020 does not seem to support the style="width:250" part (simply ignores it) but this might perhaps not be an issue by the time Dymola ships MSL 4.0.0 (i.e., when they fixed the HTML rendering).

A working alternative is to use `` instead of ``. It seemed to work reliably - and is even shorter.

For the second the results in Dymola were:
- Info-viewer, works.
- Documentation-viewer, ignored.
- Exported html, included in the generated html and then ignored by Firefox and Edge.

svorkoetter:
I checked with our GUI folks, and MapleSim just uses the user's default browser. Since those all support SVG these days, it should be fine.

HansOlsson:
> 
> 
> I'll check if it works in our browser. From the pictures posted so far though, the old one definitely looks better at that size. I've found that when resizing any sort of image to thumbnail size, it's often necessary to increase the contrast to maintain the same visual impact.

I also agree with this.

svorkoetter:
> the **official** colour is that grey but it too find it harder to recognise. 

I think we might need an official decision to use a darker colour for smaller scales.

dietmarw:
> I think we might need an official decision to use a darker colour for smaller scales.

OK I will raise this with the board. 

How are we on using SVG here. The (rather old, in this respect) specification says 

> 18.6.5.6: The supported file formats include PNG, BMP and JPEG, other 
supported file formats are unspecified.

which kind of allows for others too (or what does "supported file formats are unspecified" actually mean). And I'd say as long as the tool vendors are OK with it should be fine to use SVG here.

HansOlsson:
> 
> 
> > I think we might need an official decision to use a darker colour for smaller scales.
> 
> OK I will raise this with the board.
> 
> How are we on using SVG here. The (rather old, in this respect) specification says
> 
> > 18.6.5.6: The supported file formats include PNG, BMP and JPEG, other
> > supported file formats are unspecified.
> 
> which kind of allows for others too (or what does "supported file formats are unspecified" actually mean). And I'd say as long as the tool vendors are OK with it should be fine to use SVG here.

I agree that the text doesn't make sense, but I would assume that it was intended to mean other file formats are not officially supported. As far as I can see this was added in 2.0, and haven't been changed since then.

maltelenz:
I am, on pure principle, opposed to start using SVG in the `Bitmap` primitive until it is explicitly written in the specification that is is a supported format.

It seems there was some kind of decision here, but it doesn't look like it is in the specification yet? https://github.com/modelica/ModelicaSpecification/issues/174

dietmarw:
> A working alternative is to use `` instead of ``. It seemed to work reliably - and is even shorter.

OK updated in d481fbe1c3961cf81e96874690ee2c0515e2b7cf

`` (note the quotes btw)  would be HTML4 only. Whilst HTML4 and 5 would 

> 
> For the second the results in Dymola were:
> 
>     * Info-viewer, works.
> 
>     * Documentation-viewer, ignored.
> 
>     * Exported html, included in the generated html and then ignored by Firefox and Edge.

svorkoetter:
> I am, on pure principle, opposed to start using SVG in the `Bitmap` primitive until it is explicitly written in the specification that is is a supported format.
Is it being used in a Bitmap primitive here? I only saw it in an  tag in the HTML. But I agree that using it in a Bitmap primitive should not be allowed. If you want vector graphics in an icon, just use annotations to draw it.

dietmarw:
No it is only used in HTML inside the `img` tag. So no bitmap primitive.

maltelenz:
I think there was some confusion. I'm completely fine with this pull request as it stands.

I interpreted Dietmars comment above about mentioning `18.6.5.6` as talking about the icon. As I read it, that section is only relevant for discussions about icons (and diagrams).

beutlich:
Have a look at [README.md](https://github.com/beutlich/ModelicaStandardLibrary/blob/new-logo/README.md) where I added the scaled SVG (similar as MAP-FMI does in [README.adoc](https://github.com/modelica/fmi-standard/blob/master/README.adoc)).

dietmarw:
OK so just to summarise the status of this PR draft:
1. old (current) Dark logo for the package is to be preferred. Pending feedback from MA board if OK with them
2. Including SVGs in **documentation**  via `` is not forbidden (never has been) as we do not specify the allowed image formats used inside HTML in the Modelica Specification (correct @HansOlsson ?). So this means it is down to "us" as MAP-LIB agreeing if it is OK to start using SVGs inside the MSL's documentation. And it is already supported in three tools as we have established: Dymola, OMEdit, Wolfram SystemModeler. 

I'm going to open this PR up out of Draft mode now.

HansOlsson:
> 
> 
> OK so just to summarise the status of this PR draft:
> 
>     1. old (current) Dark logo for the package is to be preferred. Pending feedback from MA board if OK with them
> 
>     2. Including SVGs in **documentation**  via `` is not forbidden (never has been) as we do not specify the allowed image formats used inside HTML in the Modelica Specification (correct @HansOlsson ?). 

Correct.

dietmarw:
I think this is a good plan. So if the board can not come to an agreement it will simply stay like it was before. I'll update the PR to remove the modification of the package icon.

Issue #3389, opened 2020-02-04
Split Electrical.Analog into single files


christiankral:
Once #3291 is merged the Analog package shall be split into single files before releasing MSL 4.0.0. This ticket serves as reminder to not forget about it.

PR #3396, opened 2020-02-07
Split Mechanics.MultiBody into single files

tobolar:
refs #2975


tobolar:
I checked:
- all classes saved in single files
- pedantic check of MultiBody in Dymola 2020x
- all examples simulates (no regression check)
- all test models in ModelicaTest.MultiBody simulates (no regression check)

tobolar:
> I'll just put this on hold for now

I'm in favor of it. 
I would then prefer to wait with proceeding of other MultiBody-related issues till this PR is merged, to prevent possible conflicts.

beutlich:
> 
> 
> I'll just put this on hold for now since I'm not sure we should introduce this change now after alpha has been tagged. I would like to get @beutlich 's opinion on this. After all it is a feature which helps the development but is not really important for the released version. So can be added right after 4.0.0 release is done.

Hm, @christiankral opened #3389 recently to get Electrical.Analog split for MSL 4.0.0. Thus, if we put it on hold for MultiBody we should also avoid it for Electrical.Analog.

dietmarw:
Yes I personally would put this also on hold there so it does not interfere with the testing. But I leave that up to you @beutlich.

PR #2782, opened 2018-11-21
Fix mass flow rate calculation

wischhusen:
Proposal for a fix of issue #2770.

Close #2770.

PR #3405, opened 2020-02-11
maint/3.2.3: Fix mass flow rate calculation

### LINKED ISSUE #2770: Wrong mass flow rate calculation in Modelica.Fluid.Fittings.GenericResistances.VolumeFlowRate
The mass flow rate of volumeFlowRate is always zero if the pressure drop is a linear function of volume flow rate. 

The volume flow of is calculated by solving the quadric equation dp=a*V_flow^2+b*V_flow. This fails if a is zero. The actual problem is in Modelica.Fluid.Dissipation.PressureLoss.General.dp_volumeFlowRate_MFLOW:

1) a is silently replaced by max(a,Modelica.Constants.eps) (the same for b, but this is not an issue here)
2) The volume flow rate is calculated by (-b/(2*a) + ((b/(2*a))^2 + dp/a)^0.5   (I simplified a bit)
The point is, for small a the right term sqrt(X^2 + eps) --> X, and this leads to V_flow=0 if b>0. 

Proposed solution:
1) The replacement  a,b=max(Modelica.Constants.eps, abs(IN_con.a,b)) should be removed. Instead, for (a*a)<=0 the linear solution dp/b should be returned. This should be relevant only if a is given as a parameter. 
2) One should calculate only the solution with the greater absolute value with the formula above. In the case b>0 (which is the common case) this is the negative solution  
    x1= (-b/(2*a) - ((b/(2*a))^2 + dp/a)^0.5
The second root should be calculated by Vieta's formula
    x2 = c/a/x1
In the common case b>0 the second solution x2 is the physically meaningful.
I removed the regularized power function SmoothPower here only for simplicity of presentation. It should be kept in the actual code.

beutlich:
Back-port #2782 to fix #2770 in maint/3.2.3.

### DIFF
--- a/Modelica/Fluid/Dissipation.mo
+++ b/Modelica/Fluid/Dissipation.mo
@@ -4530,37 +4530,42 @@ This record is used as <strong> input record </strong> for the pressure loss fun
         output SI.Pressure DP "Output for function dp_volumeFlowRate_DP";
 
       protected
-        Real a=max(Modelica.Constants.eps, abs(IN_con.a));
-        Real b=max(Modelica.Constants.eps, abs(IN_con.b));
+        Real a=abs(IN_con.a);
+        Real b=abs(IN_con.b);
 
         SI.VolumeFlowRate V_flow=m_flow/max(Modelica.Constants.eps, IN_var.rho)
           "Volume flow rate";
-        SI.Pressure dp_min=IN_con.dp_min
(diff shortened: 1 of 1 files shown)


beutlich:
@thorade @wischhusen Any objections with back-porting the bug-fix?

wischhusen:
No I have no objections for that

Issue #3409, opened 2020-02-13
Not fully specified initial system in Modelica.Magnetic.FluxTubes.Examples.Hysteresis.SinglePhaseTransformerWithHysteresis1


modelonrobinandersson:
In order to setup the initial equation system to Modelica.Magnetic.FluxTubes.Examples.Hysteresis.SinglePhaseTransformerWithHysteresis1 additional equations are needed in order to obtain a balanced system.

That the equations below are added is natural as a consequence of the start values and fixed=true for these states:
```
winding1.i = 0;
winding2.i = 0;
core.MagRel = 0.5;
```
However, the fourth equation that is needed is not clearly defined, and different compiler heuristics will lead to different initial equations.
We suggest to explicitly add 
```
initial equation
    core.derHstat = 0.0;
```

hubertus65:
I think that a complete specification of the initialization is necessary to achieve the same results across different tools. It would be awesome if CI could check for this in some way and flag this as a QA issue. A first step could be to make this a clearly formulated best practice. I think it is already, but I am not sure.

beutlich:
Resolved in master and maint/3.2.3 branches.

PR #3410, opened 2020-02-13
Fix initial conditions of Magnetic.FluxTubes hysteresis examples

modelonrobinandersson:
This is related to issue #3409 

In order to setup the initial equation system to Modelica.Magnetic.FluxTubes.Examples.Hysteresis.SinglePhaseTransformerWithHysteresis1 additional equations are needed in order to obtain a balanced system.

That the equations below are added is natural as a consequence of the start values and fixed=true for these states:
```
winding1.i = 0;
winding2.i = 0;
core.MagRel = 0.5;
```
However, the fourth equation that is needed is not clearly defined, and different compiler heuristics will lead to different initial equations.
We suggest to explicitly add 
```
initial equation
    core.derHstat = 0.0;
```

### DIFF
--- a/Modelica/Magnetic/FluxTubes.mo
+++ b/Modelica/Magnetic/FluxTubes.mo
@@ -2709,7 +2709,7 @@ The differences between these two models in static behaviour can be analysed and
               5,500; 6,-24; 7,24; 8,-24; 9,24; 10,-24; 11,24; 12,-24; 13,24; 14,
               -24; 15,24])
           annotation (Placement(transformation(extent={{-90,50},{-70,70}})));
-        Basic.ElectroMagneticConverterWithLeakageInductance winding1(N=1) "Winding 1" annotation (Placement(transformation(extent={{30,50},{50,70}})));
+        Basic.ElectroMagneticConverterWithLeakageInductance winding1(i(start=0, fixed=true), N=1) "Winding 1" annotation (Placement(transformation(extent={{30,50},{50,70}})));
         Modelica.Electrical.Analog.Basic.Ground elGnd1
           annotation (Placement(transformation(extent={{0,30},{20,50}})));
         Basic.Ground magGnd1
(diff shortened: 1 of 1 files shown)


dietmarw:
Is this also relevant for master (i.e., 4.0.0) or only for 3.2.3 as maintenance bugfix.

christiankral:
`Examples.Hysteresis.SinglePhaseTransformerWithHysteresis1` is not the only model with only partly specified initial conditions. So I will update this PR to also cover the other hysteresis models.

christiankral:
> Is this also relevant for master (i.e., 4.0.0) or only for 3.2.3 as maintenance bugfix?

Yes this is also relevant to master.

christiankral:
@dietmarw @beutlich I actually wonder why this issue was not yet discovered when releasing MSL 3.2.3 -- or was it possibly overlooked (by myself).

christiankral:
In the three phase transformer model `protected` was remove to access all the magnetic components from outside the model. Particularly, as `protected` is not needed and as the three phase transformer model is clearly intended to be used in the example only, this change is justified on my opinion.

beutlich:
> 
> 
> @dietmarw @beutlich I actually wonder why this issue was not yet discovered when releasing MSL 3.2.3 -- or was it possibly overlooked (by myself).

It was reported [here](https://github.com/modelica/MAP-LIB_ReferenceResults/blob/v3.2.3/Modelica/Magnetic/FluxTubes/Examples/Hysteresis/SinglePhaseTransformerWithHysteresis1/translate_passed.log), but it is not detected since `translateModel` returns `true` even in pedantic mode of Dymola.

```
Advanced.PedanticModelica:=true;
translateModel("Modelica.Magnetic.FluxTubes.Examples.Hysteresis.SinglePhaseTransformerWithHysteresis1");
 = true
```

beutlich:
I will update the [release check](https://github.com/modelica-tools/msl-release) to improve the error detection.

beutlich:
> 
> 
> I will update the [MSL release checks](https://github.com/modelica-tools/msl-release/blob/master/ReleaseChecks.mo) to improve the error detection.

* The updated translation verification is available by https://github.com/modelica-tools/msl-release/pull/2.
* The updated MSL v4.0.0 Beta.1 translation check failures are reported by https://github.com/modelica/ModelicaStandardLibrary/issues/3435. You can see that the issues with the  Magnetic.FluxTubes hysteresis examples are now reported, too.

PR #3416, opened 2020-02-14
Added Clocked to Modelica.UsersGuide.Overview

bernhard-thiele:
The library overview in Modelica.UsersGuide.Overview lacks a row for the Modelica.Clocked package. This PR adds this missing row.


sjoelund:
SVGs were allowed in the HTML documentation, right? That should look nicer...

bernhard-thiele:
Okay, the picture should be better now. Seemed like zooming out in Dymola and doing a screenshot gives much better results than exporting the picture and resizing it afterwards.

PR #3427, opened 2020-02-15
Fix MSL 4.0.0 initial conditions of Magnetic.FluxTubes hysteresis examples

christiankral:
This issue is related to #3409 and serves as equivalent PR to #3410 for (branch master) MSL 4.0.0. 

In the three phase transformer model `protected` was remove to access all the magnetic components from outside the model. Particularly, as `protected` is not needed and as the three phase transformer model is clearly intended to be used in the example only, this change is justified on my opinion.


christiankral:
@dietmarw @beutlich Are missing initial conditions considered to get the **bug** tag?

christiankral:
@modelonrobinandersson for your information

beutlich:
> 
> 
> @dietmarw @beutlich Are missing initial conditions considered to get the **bug** tag?

It depends. It is not teh case if only an example is affected. It is the case if a library component is affected. And since this issue already is merged to maint/3.2.3 (and we usually only back-port bug-fixes) there is good reason to label it as bug.

Issue #3164, opened 2019-10-22
Add an arrow to icon of SweptVolume?


sergniko:
Is it a good idea to add an arrow to **Fluid.SweptVolume** as for most of **Translational.Components**?
In class documentation it is said: "The force of the flange results from the pressure difference between medium and ambient pressure and the cross sectional piston area". But nothing about sign convention.
For me, as a new user, it took a lot of time to understand why the force is negative while the pressure is above ambient pressure.
Adding arrow can help to get better understanding. For example **Translational.Examples.WhyArrows** is very helpful.

beutlich:
@christiankral As we no longer have active library officers for Fluid and as @sergniko referred to Mechanical.Translational, could you please take care of it? Thanks a lot.

christiankral:
I somehow missed this issue because of the focus on MSL 4.0.0.

From investigating `ModelicaTest.Fluid.TestComponents.Machines.TestSweptVolume` 

![image](https://user-images.githubusercontent.com/4184218/74586550-f6369280-4fe8-11ea-8824-05da8a0baaf8.png)

![image](https://user-images.githubusercontent.com/4184218/74586291-62fc5d80-4fe6-11ea-8604-cdb9f98e7a77.png)
 
So I had a look at `Modelica.Fluid.Machines.SweptVolume` and created PR #3428 to include an arrow.

dietmarw:
fixed in #3428

sergniko:
The arrow is missing again in latest build/release.

PR #3428, opened 2020-02-15
Add translational reference direction arrow to SweptVolume model

christiankral:
Refs #3164


christiankral:
image

sergniko:
The arrow is missing again in latest OpenModelica 1.18.0~dev-24-g54771e5.

dietmarw:
@sergniko You need to check what version of MSL you have opened in OpenModelica. The issue was fixed for MSL 4.0.0. So if you have still 3.2.3 opened in OpenModelica then you would not see the fix.

PR #3431, opened 2020-02-17
Fix unconsidered change of parameter c_usefulFlux from default to start

### LINKED ISSUE #3300: LeakageWithCoefficient should have a port for reluctance input

christiankral:
Fixes #3300
Refs #3424

@beutlich: I suppose there is some scheme on how the conversion script commands are sorted and arranged in `ConvertModelica_from_3.2.3_to_4.0.0.mos`, respectively. I hope you agree with on where I put the code.

I also added a test model to `ModelicaTestConversion4.mo` to test both cases: with and without default parameter. Both cases are converted OK.


beutlich:
> @beutlich: I suppose there is some scheme on how the conversion script commands are sorted and arranged in `ConvertModelica_from_3.2.3_to_4.0.0.mos`, respectively. I hope you agree with on where I put the code.

No, there is no special sorting in the conversion script.

Issue #3417, opened 2020-02-14
Unnecessary event generation in Modelica.Electrical.Analog.Examples.OpAmps.SignalGenerator


modelonrobinandersson:
In the BTL of the model Modelica.Electrical.Analog.Examples.OpAmps.SignalGenerator we have the following equation block:
```
Torn variables:
0     r1.v
1     opAmp1.v_in
2     r2.v
3     opAmp1.v_out

Iteration variables:
0     r2.i (start=Vps / r2.R)

Torn equations:
0     r1.v := r1.R_actual * (- r2.i)
1     opAmp1.v_in := r1.v + opAmp2.v_out
2     r2.v := r2.R_actual * r2.i
3     opAmp1.v_out := smooth(0, if opAmp1.V0 * opAmp1.v_in > opAmp1.Vps then opAmp1.Vps elseif opAmp1.V0 * opAmp1.v_in < Vns then Vns else opAmp1.V0 * opAmp1.v_in)

Residual equations:
0     r2.v = opAmp1.v_in - opAmp1.v_out

```
According to the Modelica language specification, there is no guarantee that the smooth-operator does not generate any events.
When this model is simulated with Optimica Compiler Toolkit, the event generation resulting from torn equation 3 with the smooth-operator causes an infinite loop in the event iteration, and this is due to minor numerical noise. There is no reason to have event generation here since r2.i varies to such a small degree. Therefore we suggest setting the parameter "strict" on the components opAmp1 and opAmp2, equal to "true" instead of the current value "false", using the more appropriate noEvent branch instead.

beutlich:
See https://github.com/modelica/ModelicaStandardLibrary/pull/3432#issuecomment-590191439.

PR #3432, opened 2020-02-19
Avoid unnecessary event generation in SignalGenerator example

dietmarw:
Applied #3417 to master. See #3417


beutlich:
@gkurzbach Can you please analyze why BDF simulation in SimulationX 4.0 no longer converges and confirm if it is a tool-issue?

gkurzbach:
@beutlich @dietmarw  
I can **not** confirm that this is a tool issue of SimulationX, but maybe of the Optimica toolkit.

The equation 
`opAmp1.v_out := smooth(0, if opAmp1.V0 * opAmp1.v_in > opAmp1.Vps then opAmp1.Vps elseif opAmp1.V0 * opAmp1.v_in < Vns then Vns else opAmp1.V0 * opAmp1.v_in)`
is part of a block so the existence of an event handling is essential for step size control and to be able to solve that. This is what event generations and the special behavior of conditions on events (only change at event instants) is good for. The information provided by `smooth(0,...)` is that this is only a kink (0 times continously differentiable) and no full event iteration is necessary. When this is taken into account there should be no problems in solving this block.

Issue #3425, opened 2020-02-15
Proposal to split Modelica.Blocks into single files


christiankral:
I am asking the library officers @AHaumer and @MartinOtter if they have any intentions to have the package `Modelica.Blocks` split into single files. In case you agree, it makes sense to have this done before MSL 4.0.0 gets released.

AHaumer:
It makes sense, but I'd ask for @MartinOtter 's opinion.

christiankral:
@MartinOtter If you agree, I can prepare a PR to get the package split into individual files. On my opinion the great advantage of single files is, that a GitHub history can directly be created for each Modelica class of the package. This way it is very easy to check what changes have been applied to each Modelica class.

beutlich:
Since there is an open PRs for "Blocks" I am removing the milestone.

MartinOtter:
> @MartinOtter If you agree, I can prepare a PR to get the package split into individual files. On my opinion the great advantage of single files is, that a GitHub history can directly be created for each Modelica class of the package. This way it is very easy to check what changes have been applied to each Modelica class.

This is fine from my side

PR #3411, opened 2020-02-13
Split Electrical.Analog into single files

christiankral:
and applied `ttws` to cleanup white spaces

Resolves #3389

PR #3433, opened 2020-02-20
Split Modelica.Clocked into individual physical directories and files

christoff-buerger:
Likewise to tickets #3411, #3396 or for example #3425, this PR is only about splitting the `Modlica.Clocked` packages into physical independent directories and files.

There are no further changes.


christoff-buerger:
@dietmarw: Thanks a lot for the blank-line and superfluous annotation removal! I missed that ones.

Issue #3412, opened 2020-02-13
Analysis of the MSL v4.0.0-beta.1 regressions vs. v3.2.3-rc.2


beutlich:
This is the public [report](https://beutlich.github.io/MSL400BETA1FAILED/index.html) of the 32 failed examples models for MSL v4.0.0-beta.1 vs. v3.2.2-rc.2.

I believe many of the changes are false-positives due to updated examples. Please check, analyze and give feedback.

 - [x] Modelica.Blocks.Examples.NoiseExamples.ActuatorWithNoise.ActuatorWithNoise
 - [x] Modelica.Blocks.Examples.NoiseExamples.DrydenContinuousTurbulence.DrydenContinuousTurbulence
 - [x] Modelica.Electrical.Analog.Examples.OpAmps.Differentiator.Differentiator
 - [x] Modelica.Electrical.Analog.Examples.OpAmps.HighPass.HighPass
 - [x] Modelica.Electrical.Analog.Examples.OpAmps.Integrator.Integrator
 - [x] Modelica.Electrical.Analog.Examples.OpAmps.LowPass.LowPass
 - [x] Modelica.Electrical.Machines.Examples.InductionMachines.IMC_InverterDrive.IMC_InverterDrive
 - [x] Modelica.Electrical.PowerConverters.Examples.ACDC.RectifierCenterTap2mPulse.ThyristorCenterTap2mPulse_R.ThyristorCenterTap2mPulse_R
 - [x] Modelica.Electrical.PowerConverters.Examples.ACDC.RectifierCenterTap2mPulse.ThyristorCenterTap2mPulse_RL.ThyristorCenterTap2mPulse_RL
 - [x] Modelica.Electrical.PowerConverters.Examples.ACDC.RectifierCenterTap2mPulse.ThyristorCenterTap2mPulse_RLV.ThyristorCenterTap2mPulse_RLV
 - [x] Modelica.Electrical.PowerConverters.Examples.DCDC.ChopperStepDown.ChopperStepDown_R.ChopperStepDown_R
 - [x] Modelica.Electrical.PowerConverters.Examples.DCDC.ChopperStepDown.ChopperStepDown_RL.ChopperStepDown_RL
 - [x] Modelica.Electrical.PowerConverters.Examples.DCDC.ChopperStepUp.ChopperStepUp_R.ChopperStepUp_R
[...]

PR #3434, opened 2020-02-22
Ensure reproducible simulation results of NoiseExamples

beutlich:
See #3412: Set fixed local seed to 5009189 to obtain identical simulation results as of MSL v3.2.3


beutlich:
I consider this PR kind of optional. If we can live with the new simulation results due to the new automatic local seed we can simply ignore this PR.

beutlich:
No need to merge though

PR #3441, opened 2020-02-24
Rename NoiseExamples -> Noise

### LINKED ISSUE #3440: NoiseExamples package has odd name
The package `Modelica.Blocks.Examples.NoiseExamples` contains examples using the `Modelica.Blocks.Noise` components.

Since it is already placed inside the `Examples` package, would it make more sense to call it `Modelica.Blocks.Examples.Noise`? The extra `Examples` seems redundant. I could not find similar examples elsewhere either.

dietmarw:
Fixes #3440

### DIFF
--- a/Modelica/Resources/Scripts/Conversion/ConvertModelica_from_3.2.3_to_4.0.0.mos
+++ b/Modelica/Resources/Scripts/Conversion/ConvertModelica_from_3.2.3_to_4.0.0.mos
@@ -1938,6 +1938,6 @@ convertElement("Modelica.Media.Interfaces.PartialTwoPhaseMedium.BaseProperties",
 convertElement("Modelica.Media.R134a.R134a_ph.BaseProperties", "R", "R_s")
 convertElement("Modelica.Media.Water.WaterIF97_base.BaseProperties", "R", "R_s")
 
-// Rename of NoiseExample -> Noise (see #3440)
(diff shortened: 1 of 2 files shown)


Modelica/Resources/Scripts/Conversion/ConvertModelica_from_3.2.3_to_4.0.0.mos
```diff
@@ -1938,6 +1938,6 @@ convertElement("Modelica.Media.Interfaces.PartialTwoPhaseMedium.BaseProperties",
 convertElement("Modelica.Media.R134a.R134a_ph.BaseProperties", "R", "R_s")
 convertElement("Modelica.Media.Water.WaterIF97_base.BaseProperties", "R", "R_s")
 
-// Rename of NoiseExample -> Noise (see #3440)
+// Rename of NoiseExamples -> Noise (see #3440)
 convertClass("Modelica.Blocks.Examples.NoiseExamples",
              "Modelica.Blocks.Examples.Noise")
```
beutlich: Conversion is not required for examples, but still not wrong to have it. Please move it up before the auto-generated conversion rules.

dietmarw:
One reason I added the conversion rule was that (unfortunately) the Noise (example) package contains quote a lot of special models that users might have used.

Issue #3435, opened 2020-02-23
Analysis of example models of MSL v4.0.0-beta.1 failing a pedantic translation


beutlich:
The following models/blocks with StopTime annotation fail translation if pedantic=true:

- [x] Modelica.Blocks.Examples.PID_Controller
- [x] Modelica.Clocked.Examples.Systems.EngineThrottleControl
- [x] Modelica.Clocked.Examples.Elementary.ClockSignals.RotationalSample
- [x] Modelica.Clocked.Examples.Elementary.ClockSignals.LogicalSample
- [x] Modelica.Electrical.Spice3.Examples.Oscillator
- [x] Modelica.Electrical.Spice3.Examples.CoupledInductors
- [x] Modelica.Magnetic.FluxTubes.Examples.Hysteresis.HysteresisModelComparison
- [x] Modelica.Magnetic.FluxTubes.Examples.Hysteresis.SinglePhaseTransformerWithHysteresis1
- [x] Modelica.Magnetic.FluxTubes.Examples.Hysteresis.SinglePhaseTransformerWithHysteresis2
- [x] Modelica.Magnetic.FluxTubes.Examples.Hysteresis.ThreePhaseTransformerWithRectifier
- [x] Modelica.Magnetic.QuasiStatic.FluxTubes.Examples.Leakage.CylinderLeakage
- [x] Modelica.Magnetic.QuasiStatic.FluxTubes.Examples.Leakage.GeneralLeakage
- [ ] Modelica.Fluid.Examples.HeatingSystem
- [ ] Modelica.Fluid.Examples.AST_BatchPlant.Test.OneTank
- [ ] Modelica.Fluid.Examples.NonCircularPipes
- [x] Modelica.Fluid.Examples.TraceSubstances.RoomCO2WithControls
- [x] Modelica.Fluid.Examples.InverseParameterization
- [ ] Modelica.Fluid.Examples.Explanatory.MeasuringTemperature

beutlich:
Hysterisis models are addressed by #3427.

beutlich:
Spice3 models are addressed by #3438.

casella:
@beutlich, I just checked all the Modelica.Fluid examples on branch v4.0.0-beta.1 using Dymola 2020x and pedantic mode, but I don't get any error or warning. Can you please report which errors do you get?

beutlich:
Here the warnings:

```
Translation of Modelica.Fluid.Examples.HeatingSystem:

[...]

PR #3448, opened 2020-02-25
Refs #3435: Fully specify initial conditions of examples in QuasiStatic.FluxTubes.Examples.Leakage

christiankral:


christiankral:
It is not a component issue as there may be different sources involved in an example where only one source reference angle shall be fixed.

PR #3449, opened 2020-02-25
Update version management section to reflect semver based versions

dietmarw:
versionBuild is no longer advised to use and hence removed.

### DIFF
--- a/Modelica/package.mo
+++ b/Modelica/package.mo
@@ -2236,12 +2236,12 @@ Further development and maintenance of the Modelica Standard Library is performe
 two branches on the public <a href=\"https://github.com/modelica/ModelicaStandardLibrary.git\">GitHub repository</a> of the Modelica Association.
 </p>
 <p>
-Since version 4.0.0 the Modelica Standard Library uses semantic versioning following the 
+Since version 4.0.0 the Modelica Standard Library uses semantic versioning following the
 convention:
 </p>
  <blockquote><strong><code>MAJOR.MINOR.BUGFIX</code></strong></blockquote>
-<p> 
-This provides a mechanism for maintaining releases and bug-fixes in a well defined way and is inspired 
+<p>
+This provides a mechanism for maintaining releases and bug-fixes in a well defined way and is inspired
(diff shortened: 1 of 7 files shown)


Modelica/package.mo
```diff
@@ -2335,7 +2336,7 @@ more of the following changes.
 <li> Introducing a new name in the public section of a class
          (model, package, ...) or in any section of a partial class is <strong>not</strong> allowed.
          Since otherwise, a user might use this new name and when storing its model
-         and loading it with an older build-version, an error would occur.</li>
+         and loading it with an older bugfix-version, an error would occur.</li>
```
beutlich:
```suggestion
         and loading it with an older bug-fix version, an error would occur.</li>
```

dietmarw:
@henrikt-ma You might want to add your (valuable) input above to the MCPs in question since they might get lost here in this PR.

dietmarw:
I'll update the documentation with your input but would still mention semver since that is were we got the original idea from anyway. So might as well state that it is inspired by that.

beutlich:
> 
> 
> I'll update the documentation with your input but would still mention semver since that is were we got the original idea from anyway. So might as well state that it is inspired by that.

Please update, too:

https://github.com/modelica/ModelicaStandardLibrary/blob/3dfe8836cbacdd4a1fd9ea4b383259bfd6657c3e/Modelica/package.mo#L2359

henrikt-ma:
> Instead state we use semantic versioning (MAJOR.MINOR.BUGFIX), where bug-fixes are restricted as described in this document; and state that it will be clarified later.

Sounds like the right thing to do given the tight schedule.  I hope that we will later be able to interpret this as MSL 4 conforming to the future definitions of MCP0016, even for the versions of MSL 4 that were released before MCP0016 was accepted.  (If we would have had more time, I would have suggested to actually state in MSL 4.0.0 what we think should go into MCP0016 — that way, there would be no need in the future to retroactively change the version semantics of already released MSL versions.)

dietmarw:
I've updated the description and have hopefully covered the concerns.

PR #3457, opened 2020-02-28
Removed Dymola-specific directory

casella:
The [Modelica/Resources/Scripts](https://github.com/modelica/ModelicaStandardLibrary/tree/master/Modelica/Resources/Scripts) contains a Dymola directory. I understand the MSL should be shipped free of tool-specific annotations and resources, which every tool vendor can then add at will in the version shipped with the tool.


casella:
I guess this could be re-added in the future when plot annotations are standardized.

casella:
I thought we had different rules, but it's not me leading the game at this moment :)

PR #3480, opened 2020-03-03
Refs #3425: Split Blocks into single files

christiankral:
I also applied 
- `ttws -b` to suppress repeated empty lines 
- `ttws -s` to strips leading or trailing white spaces


christiankral:
OK, so we also wait for #3459 to be merged (which does not have the MSL4.0.0 milestone (yet))

PR #3490, opened 2020-03-07
Addresses #3412 (OpAmp-examples regression)

AHaumer:
If desired, this PR changes the 4 OpAmp-examples so that the original reference results are met.
The alternative would be (as mentioned in #3412): Eccaept the changes in the examples and store new reference results.


beutlich:
Should we rename "der_" of Differentiator to "derivative" to match HighPass?

PR #3496, opened 2020-03-10
Rename der_ to derivative in Modelica.Electrical.Analog.Examples.OpAmps.Differentiator

beutlich:
This is a proposal to rename der_ to derivative to be consistent with the HighPass example model.


AHaumer:
To be consistent with Modelica.Blocks:
The Derivative-block is not ideal 1/(1+sT), the Der-Block is ideal.
The OpAmp application is an ideal derivative, therefore Der would be the right name.

beutlich:
I see.

PR #3503, opened 2020-03-15
Improve fault-tolerance of Modelica.Utilities.Strings.substring w.r.t. index arguments and remove undocumented case with endIndex=-999

beutlich:
I missed the feature of Modelica.Utilities.Strings.substring to return an empty string. This is why I started this PR.

* Do not assert on incorrect arguments of substring, but try to correct them and raise a warning instead.
* substring was also improved to return an empty string, in case endIndex is less than startIndex.

### DIFF
--- a/.CI/Test/ModelicaStrings.c
+++ b/.CI/Test/ModelicaStrings.c
@@ -7,5 +7,12 @@ const char* ModelicaStrings_substring(const char* string,
                                       int startIndex, int endIndex);
 
 int main(int argc, char **argv) {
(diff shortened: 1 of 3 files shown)


beutlich:
@sjoelund @DagBruck Can you please do the review. Thanks.

casella:
@sjoelund is on holiday

beutlich:
> I agree with @HansOlsson that you should replace "negative" with "non-positive" or "less than one" in the warning when startIndex < 1

That should be the case already.

PR #3504, opened 2020-03-16
Fix initial conditions of quasi-static Leakage examples

christiankral:
Back-port #3448 also to maint/3.2.3.


christiankral:
Yes, this issue is fixed in master

PR #3507, opened 2020-03-16
Remove obsolete text located in Diagram layer

christiankral:
The text of the ~Icon~ Diagram layer about quasi-static machine models does not belong to the transient package `Magnetic.FundamentalWaves`


christiankral:
Yes, you are certainly right

PR #3508, opened 2020-03-17
#3479 fix

### LINKED ISSUE #3479: Modelica.Clocked.Examples.Systems.Utilities.ComponentsThrottleControl.Engine uses shiftSample with parametric resolution on an event clock
_Modelica.Clocked.Examples.Systems.Utilities.ComponentsThrottleControl.Engine_ has an event clock (_engine.crankshaftPositionEvent_) defined in the component _engine_. This clock is used as input to _engine.inductionToPowerDelay_, where it defines the clock of _engine.inductionToPowerDelay.delay_, which is an instance of _Modelica.Clocked.RealSignals.NonPeriodic.FractionalDelay_.

Now, _FractionalDelay_ uses a shiftSample with resolution, using the resolution set by the parameter _resolution_. A shiftSample can only be applied to an event clock if the resolution is set to 1 (See Modelica specification 3.4, page 206). In this case, the parameter _resolution_ is set to 1, but it is not indicated that the parameter should be evaluated during translation with the `Evalute = true` annotation.

It also seems to me that the parameter _resolution_ should have `Evalute = true` since it is used to determine the size of the array _u_buffer_ (the same goes for the parameter _shift_).

christoff-buerger:
Fix #3479; please check the issue for a rationale of the final modification.

### DIFF
--- a/Modelica/Clocked/Examples/Systems/Utilities/ComponentsThrottleControl/InductionToPowerDelay.mo
+++ b/Modelica/Clocked/Examples/Systems/Utilities/ComponentsThrottleControl/InductionToPowerDelay.mo
@@ -3,22 +3,24 @@ block InductionToPowerDelay
   "Accounts for the induction-to-power stroke lag."
   extends Modelica.Blocks.Icons.Block;
 
-  Modelica.Blocks.Interfaces.RealInput m_a(unit="g")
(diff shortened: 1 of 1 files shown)


beutlich:
Can someone from DLR please cross-check. Thanks.

PR #3512, opened 2020-03-18
Remove questionable geometries louver fin and slit fin from ModelicaTest.Fluid.Dissipation.Verifications.HeatTransfer.HeatExchanger.kc_roundTube

beutlich:
Alternative PR for #3499 and #3509 to address #2969.


beutlich:
I tested this model also in SimulationX where it now runs for the first time.

PR #3519, opened 2020-03-22
Add link checker

beutlich:
Checker does not run in Travis CI since it takes too long. Instead it can be run locally whenever it might be necessary.

### DIFF
--- a/.CI/.travis.py
+++ b/.CI/.travis.py
@@ -9,248 +9,14 @@ Check Modelica HTML documentation for validity
 python .travis.py [check_function] [path]
 '''
 
-import re
-import os
+import importlib
 import os.path
-import urllib2
-import ssl
 import sys
-
-from tidylib import tidy_document
-from concurrent.futures import ProcessPoolExecutor as PoolExecutor
-
-# See https://html.spec.whatwg.org/#void-elements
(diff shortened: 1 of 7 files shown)


beutlich:
@sjoelund OK for you?

beutlich:
> To run faster, you could make the requests threaded to not wait for each one in series (but not really necessary if we only run this on releases).

Good point. The requests are run in parallel now which results in an acceptable performance (about 37s). I configured the link checker to only run as cron job (once a week).

beutlich:
See https://travis-ci.org/github/beutlich/ModelicaStandardLibrary/builds/668011895#L332 for a successful CI run as cron job.

beutlich:
> Any idea why travis runs the job so slowly? It's 24 seconds for me to run the links check.

It's about 80 seconds for me. 😢 

```
C:\Projekte\MSL>Powershell -Command "Measure-Command {C:\Python27\python.exe C:\Projekte\MSL\.CI\.travis.py checkLinks C:\Projekte\MSL\ | Out-Default}"


Days              : 0
Hours             : 0
Minutes           : 1
Seconds           : 19
Milliseconds      : 618
Ticks             : 796180495
TotalDays         : 0,000921505202546296
TotalHours        : 0,0221161248611111
TotalMinutes      : 1,32696749166667
TotalSeconds      : 79,6180495
TotalMilliseconds : 79618,0495
```

dietmarw:
FWIW: just under 20 secs here.

Issue #2728, opened 2018-10-04
Remove comparison signal from regression test 


beutlich:
The nextTimeEvent variable is a protected, discrete state variable, that should not be used as comparison signal of the ModelicaTest.Tables.CombiTimeTable tests, since it depends on Modelica.Constants.inf = ModelicaServices.Machine.inf, i.e., is tool-dependent (see #2435).

GallLeo:
Removed nextTimeEvent for all test cases in `ModelicaTest.Tables.*`

Ticket can be closed after upload of new reference results.

beutlich:
Please also remove ModelicaTest.MultiBody.PlanarLoopWithMove.PlanarLoopWithMove.combiTimeTable.nextTimeEvent.

GallLeo:
1. Removed ModelicaTest.MultiBody.PlanarLoopWithMove.PlanarLoopWithMove.combiTimeTable.nextTimeEvent
from my test setup, as requested.
2. I searched through all comparison signals and could not find string "nextTimeEvent" anymore.

This will be publiclty available, as soon as we create the final reference results based on v3.2.3.

PR #3522, opened 2020-03-24
Remove combiTimeTable.nextTimeEvent from the list of comparison signals.

maltelenz:
The signal is problematic, since it jumps to Modelica.Constants.inf at the end of the simulation. This value is not well defined across tools.
The outputs from the combiTimeTable component are already tested which should be enough.


beutlich:
See also #2728.

PR #3533, opened 2020-04-05
Make vector quantity a parameter

### LINKED ISSUE #3529: Shouldn't quantity in PartialVector be a parameter or constant?
https://github.com/modelica/ModelicaStandardLibrary/blob/aaec4268d27b4677088ffa792b439b24f7981f78/Modelica/Utilities/Internal.mo#L61

Why is `quantity` in `PartialVector` an `input`? To me it really doesn't make any sense to allow it to be time dependent.

beutlich:
* Change variability of vector quantity to parameter
* Improve documentation

Resolves #3529.

### DIFF
--- a/Modelica/Mechanics/MultiBody/Types/VectorQuantity.mo
+++ b/Modelica/Mechanics/MultiBody/Types/VectorQuantity.mo
@@ -6,4 +6,30 @@ type VectorQuantity = enumeration(
     Acceleration,
     AngularVelocity,
     AngularAcceleration,
-    RelativePosition);
+    RelativePosition)
+  "Enumeration defining the kind of physical quantity represented by the vector"
+    annotation (Documentation(info="<html>
+<table border=\"1\" cellspacing=\"0\" cellpadding=\"2\">
+<tr><th><strong>Types.VectorQuantity.</strong></th><th><strong>Meaning</strong></th></tr>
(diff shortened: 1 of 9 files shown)


Modelica/Mechanics/MultiBody/Types/VectorQuantity.mo
```diff
@@ -6,4 +6,30 @@ type VectorQuantity = enumeration(
     Acceleration,
     AngularVelocity,
```
beutlich, line 7: Just noticed, that description of the alternatives might also be valuable. Will do later...

beutlich:
Well, I'd be similar to other types, e.g., https://github.com/modelica/ModelicaStandardLibrary/blob/master/Modelica/Mechanics/MultiBody/Types/ResolveInFrameA.mo

maltelenz:
> Well, I'd be similar to other types, e.g., https://github.com/modelica/ModelicaStandardLibrary/blob/master/Modelica/Mechanics/MultiBody/Types/ResolveInFrameA.mo

I am aware. I also linked to how the documentation looks for such an enumeration type in my previous comment, and the redundancy it introduces.

My comment was an attempt at poking at the current state of things, and see if we can improve for the future. Test the waters to see if others agree, as it were.

PR #3540, opened 2020-04-08
Fix component name forgotten in pull request #3532

maltelenz:
In #3532, all the components with names equal to synchronous operators were changed, but this one was missed in the model itself. The comparisonSignals.txt was already updated:
https://github.com/modelica/ModelicaStandardLibrary/blob/4b7e8aa4775fb5da2d5eca49ce3e45b818598ea3/Modelica/Resources/Reference/Modelica/Clocked/Examples/Elementary/IntegerSignals/Hold/comparisonSignals.txt#L3


beutlich:
> Don't know why I missed it.

Strange indeed, since I verified the comparisonSignals.txt when running the Clocked examples.

PR #3549, opened 2020-05-10

beutlich:
This PR improves the rendering of images in Dymola documentation view to avoid images being displayed next to the previous block.

img

### DIFF
--- a/ModelicaReference/package.mo
+++ b/ModelicaReference/package.mo
@@ -433,7 +433,7 @@ Automatically display a list of matching choices in a graphical parameter menu.
 Choices menus of replaceable elements can be automatically constructed showing the names of all classes that are either directly or indirectly derived by inheritance from the constraining class of the declaration.
 </p>
 <p>
-This can be recommended by having  <code><strong>annotation</strong> choicesAllMatching = true;</code> and disabled by having  <code><strong>annotation</strong> choicesAllMatching = false;</code>.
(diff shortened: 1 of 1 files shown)


maltelenz:
Isn't this just a bug in Dymola? If I'm understanding correctly, `blockquote` is [supposed to be](https://html.spec.whatwg.org/multipage/rendering.html#flow-content-3) `display: block`, which means there is already a line break, and the `div` is not needed.

henrikt-ma:
> Isn't this just a bug in Dymola? If I'm understanding correctly, `blockquote` is [supposed to be](https://html.spec.whatwg.org/multipage/rendering.html#flow-content-3) `display: block`, which means there is already a line break, and the `div` is not needed.

I think this comment deserves a reply.  I suppose standardizing on the WHATWG interpretation would be a possible choice for the Modelica documentation, but I don't see it stated anywhere.  This means that W3C resources seem just as relevant, and at [www.w3schools.com](https://www.w3schools.com/tags/tag_blockquote.asp) they just mention `display:block` as typical browser behavior:
> Most browsers will display the &lt;blockquote> element with the following default values: …

Anyway, regardless which HTML specification we have in mind, I think wrapping these images in `<div>` looks like good style, expressing clearly that these images should have line breaks both before and after.

maltelenz:
After some more reading, it turns out that the `img` tag, being an `inline` element, can not occur directly in the `body` element, since `body` only accepts `block` elements. As such, I see why it makes sense to wrap a `div` around `img` tags.

>  This means that W3C resources seem just as relevant, and at www.w3schools.com ...

w3schools.com is not affiliated with W3C, and as such is not a authoritative source.

PR #3550, opened 2020-05-10
Add noEvent in example function

beutlich:
Somehow related to #3547.


beutlich:
Hm, I guess it is correct as it is, since noEvent is not to be used in functions as domain guard.

PR #3554, opened 2020-05-13

beutlich:
See also #3549.

### DIFF
--- a/.CI/checkHTMLDoc/checkTags.py
+++ b/.CI/checkHTMLDoc/checkTags.py
@@ -33,6 +33,7 @@ def _checkFileTags(file_name):
     found_invalid = []
     found_mismatch = []
     found_case = []
+    found_inline = []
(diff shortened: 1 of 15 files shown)


beutlich:
Something like this?

> Figures shall be placed outside of paragraphs.
> Figures shall not be placed directly within the body context.

tobolar:
@christiankral Are you able to contribute to this discussion? You are in the end the Library officer of UsersGuide.

To be clear: I like the idea of `<div>`environment. But not sure how this works within another environments, e.g. figures within lists.

beutlich:
@tobolar OK for you to merge this PR anyway and update the guide lines later in a separate issue?

tobolar:
I don't know the procedure in detail but wouldn't it lead to lot of warnings being simply ignored? The check itself is fine to me, so ok to merge.

PR #3560, opened 2020-05-24
Fix syntax errors in documentary models of Modelica.Fluid.UsersGuide.ComponentDefinition.FluidConnectors

### LINKED ISSUE #3559: Fix the mistakes in the help document of Modelica.Fluid
In the `Modelica.Fluid.UsersGuide.ComponentDefinition.FluidConnectors`, there are two minor issues, 
1. In the example code, `FluidPort port_a, port_b;` should be `FluidPort port_a(redeclare each package Medium = Medium), port_b(redeclare each package Medium = Medium);`

2. Missing dot in the code fragment `Medium density d_a, d_b`.

beutlich:
fixes #3559

### DIFF
--- a/Modelica/Fluid/package.mo
+++ b/Modelica/Fluid/package.mo
@@ -160,8 +160,8 @@ are defined that contain medium specific values for the min, max and
 nominal attributes. Furthermore, Medium.MassFlowRate is defined as:
 </p>
 <blockquote><pre>
-<strong>type</strong> MassFlowRate =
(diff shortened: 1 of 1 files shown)


beutlich:
> I'm not sure if the arguments to f had to be changed, but the new variant works so it's ok.

Many (if not all) function variants in MSL take the pressure drop as (first) argument. This was my motivation to change it also in the call (though the function f is nowhere defined).

PR #3579, opened 2020-06-17
Update landing page doc.modelica.org

### LINKED ISSUE #3287: Improve doc.modelica.org
The web page [doc.modelica.org](https://doc.modelica.org/) is the welcome page for the generated Modelica HTML documentation.

Currently it contains the generated documentation for MSL v3.2.3 from four different tools:
* [Dymola](http://www.3ds.com/products-services/catia/products/dymola/latest-release/)
* [MapleSim](https://de.maplesoft.com/products/maplesim/)
* [OpenModelica](https://openmodelica.org/)
* [Wolfram SystemModeler](https://www.wolfram.com/system-modeler/)

I identified three tasks (see below) that could improve the value of the provided MSL documentation.

All HTML sources are available from the [gh-pages](https://github.com/modelica/ModelicaStandardLibrary/tree/gh-pages) branch of the MSL repository.

There is no strict deadline to meet here, but it would be most convenient if it can be managed in Q1/2020.

# Tool/version switch

It is expected that the number of tools providing their documentation will increase. It also is required that the documentation of MSL v3.2.3 and upcoming releases (e.g., MSL v4.0.0) should be accessible in parallel. Therefore, the idea is to have a tool selection and a version selection, similar as [doc.python.org](http://doc.python.org/) offers for language and version or [cmake.org](https://cmake.org/cmake/help/latest/) offers for version.

grafik grafik

This should be easily extensible whenever a new tool is added or an existing tool provides a new version.

We need to decided what happens in the case if a selected tool + version is not available. I have no strong opinion here, but we could either fall-back to a default case, do nothing or even raise a message.

# Quick search

A quick search (as available by the above mentioned two sites) would be very helpful. This should be based on a statically generated search index and enable a fulltext search of the selected tool- and version specific HTML content. See for example: https://docs.python.org/3/search.html?q=tuple&check_keywords=yes&area=default

# Logos

Some minimal refurbishment of the landing page would also be very nice, e.g., to incorporate the newly designed [logo](https://github.com/modelica/MA-Logos) for Modelica Libraries. Tool logos will be provided.

beutlich:
This is a shared contribution from @dietmarw, @filip-jezek, @HonzaMatosik, @pylypko and myself to improve the layout of and add algolia powered full-text search to doc.modelica.org based on a paid contract.

See https://doc.modelica.org/index-new.html for how it will look. It has been successfully tested on different displays and with different browsers (with IE9 no longer being supported).

Resolves #3287.


casella:
LGTM

PR #3580, opened 2020-06-18
Save the CFLAGS already passed to the script

sjoelund:
The last change to the lines was similar (e1c03f0002), but didn't catch this other error


sjoelund:
That would refer to the CFLAGS passed to the configure script, not the modified macro

beutlich:
> That would refer to the CFLAGS passed to the configure script, not the modified macro

Sorry, cannot follow. I cannot judge what was wrong before and what is fixed now.

beutlich (approved):

adrpo (approved):
OK by me!

PR #3585, opened 2020-06-25
Modified homotopy parameter to ensure correct initialization

casella:
The default value for homotopy is Linear, which leads OpenModelica to select an initial condition in the linear regime which is physically valid (although unstable) but not the desired one. This commit changes the homotopy parameter in specific instance of opAmp of this example to ensure the desired initialization of the opAmp output at the lower saturation limit of -15 V.


casella:
> 
> 
> I see that this is a partial back-port of #3291, but there you also set strict=true and added the documentation. Would be good to have it consistent.

Will do

casella:
@beutlich, done.

casella:
@christiankral, yet another back-port of a MSL 4.0.0 bugfix into MSL 3.2.3 maintenance for you to review.

Thanks!

Issue #3461, opened 2020-02-29
Development process with git workflow, branches and semver in mind


beutlich:
@dietmarw When writing https://github.com/modelica/ModelicaStandardLibrary/pull/3459#issuecomment-592926539, I noticed that we need to clarify the development process.

1. We should state that any changes requiring conversion got to MAJOR versions, here: MSL v5.0.0 wil be next. I noticed that you introduced the `conversion` label to mark/identify such issues
1. If we have a MINOR release scheduled as next release (say MSL v4.1.0 in fall 2020), we are not allowed to merge PRs with conversions to master, as long as both MINOR and MAJOR development happens on master. This means that we need to postpone this PR until release of MSL 4.1.0. As this might be seen unfortunate we might rethink the git workflow.

dietmarw:
I think there is a misunderstanding probably also due to the [Recommendation for a Release Process for Libraries](https://github.com/modelica/MAP-LIB_ProjectRules/blob/master/MAP-LIB_ProjectRules.md#annex-6-recommendation-for-a-release-process-for-libraries) using a vocabulary than what SemVer would use. 

Applying the SemVer vocabulary than a "maintenance" version is a new `BUGFIX` version. So Autumn 2020 would be a new v4.0.1 due (we can still decided to release earlier since it is no rule but an recommendation).

Then a new main version does not automatically mean `MAJOR` but could just as well mean `MINOR`. Here it is basically up to MAP-LIB to decided if for example Spring 2021 it should already be a MSL v5.0.0 with conversion or a MSL v4.1.0. 

And as I said the schedule is a **recommondation** and we can still decide to speed the process up or delay it. Due to our avaiiable resources it does not make sense to start development in parallel branches in parallel. We would simply not manage that (in my opinion).

beutlich:
Agreed.

* New "maintenance" = `BUGFIX` releases for MSL 4.0.0 happen from [maint/4.0.x](https://github.com/modelica/ModelicaStandardLibrary/tree/maint/4.0.x) branch.
* The master branch contains the new "main" version, either `MAJOR`or `MINOR`, depending if backward compatibility breaking changes are added or not.

PR #3586, opened 2020-06-25
Corrected homotopy operator to be consistent with provided reference

casella:
The OpAmps.SignalGenerator example has an initial output of opAmp1 which is -15 V in the provided reference results. 

This commit updates the homtopy parameter of opAmp1 to obtain a simulation which starts at the lower saturation limit and is thus consistent with the provided reference.

### DIFF
--- a/Modelica/Electrical/Analog/Examples/OpAmps.mo
+++ b/Modelica/Electrical/Analog/Examples/OpAmps.mo
@@ -783,7 +783,7 @@ package OpAmps "Examples with operational amplifiers"
       Vps=Vps,
       Vns=Vns,
       strict=false,
-      homotopyType=Modelica.Blocks.Types.LimiterHomotopy.UpperLimit)
+      homotopyType=Modelica.Blocks.Types.LimiterHomotopy.LowerLimit)
       annotation (Placement(transformation(extent={{-60,10},{-40,-10}})));
     Modelica.Electrical.Analog.Basic.Resistor r2(R=R2, i(start=Vps/R2))
       annotation (Placement(transformation(


casella:
> This change is not yet in master. The usual workflow is to first fix the master and later back-port to maint/3.2.3 and maint/4.0.x.

@beutlich, the change in master was in the default binding for the homotopy parameter, which we put to NoHomotopy (with good reasons, given that this is a completely general default) after the discussion with Hans. We can't do that in MSL 3.2.3 because that will break backwards compatibility.

It turned out that going from Linear to NoHomotopy in MSL 4.0.0 was enough to get convergence to the correct initial condition also in OpenModelica, so there was no reason to change the setting locally to something different from the default.

Here we have to change the setting locally, because we can't change it to default. At this point, I find it more reasonable to change it to LowerLimit than to NoHomotopy, because it is much safer. But there is no need so far to port this in MSL 4.0.0, which has a different default.

Does this make sense to you?

casella:
@beutlich, see #3590 

BTW, it is not too clear to me what is the difference between 4.0.0 maintenance and 4.0.1. According to semver, shouldn't a patch version exactly be made by backwards compatible bug fixes?

beutlich:
> 4.0.0 maintenance and 4.0.1

4.0.0 maintenance = 4.0.x

casella:
> 
> 
> > 4.0.0 maintenance and 4.0.1
> 
> 4.0.0 maintenance = 4.0.x

OK. So, there is no need for the time being to port bugfixes from master to maint/4.0.x uless we decide that some new features on master will go to 4.1.0. Otherwise, we can simply tag master as 4.0.1 when the time comes in September. Correct?

beutlich:
> Correct?

No. See https://github.com/modelica/ModelicaStandardLibrary/issues/3461#issuecomment-639297099 and https://doc.modelica.org/Modelica%204.0.0/Resources/helpOM/Modelica.UsersGuide.ReleaseNotes.VersionManagement.html for the development workflow.

casella:
> 
> 
> > Correct?
> 
> No. See [#3461 (comment)](https://github.com/modelica/ModelicaStandardLibrary/issues/3461#issuecomment-639297099) and https://doc.modelica.org/Modelica%204.0.0/Resources/helpOM/Modelica.UsersGuide.ReleaseNotes.VersionManagement.html for the development workflow.

@beutlich, thanks for the links. I was not aware of them, I just linked the version management documentation page to the [wiki page](https://github.com/modelica/ModelicaStandardLibrary/wiki) for everyone's reference.

So, if I get it right, any bug fix to 4.0.0 should
- be committed to master, for inclusion in the next major/minor revision
- if accepted, back-ported to maint/4.0.x (in this case 4.0.1) at once, so that tool vendors can pick the latest bugfixes without waiting for the official release

Correct?

PR #3594, opened 2020-07-02
Fix wrong or missing documentation

tobolar:


beutlich:
FYI

* If the PR is targeting master, the milestone should be set to MSL4.1.0.
* If the PR is targeting maint/4.0.x, the milestone should be set to MSL4.0.1.

PR #3595, opened 2020-07-02
Precising the description of u1 and u2

tobolar:


Modelica/Blocks/Math.mo
```diff
@@ -1006,12 +1006,11 @@ y = u1 * u2;
       Documentation(info="<html>
 <p>
 This block computes the output <strong>y</strong>
```
beutlich, line 1008:
```suggestion
This block computes the output (quotient) <strong>y</strong>
```

beutlich:
FYI

- If the PR is targeting master, the milestone should be set to MSL4.1.0.
- If the PR is targeting maint/4.0.x, the milestone should be set to MSL4.0.1.

PR #3590, opened 2020-06-28
Changed homotopy parameter to be consistent

casella:
with provided reference results, where the opAmp1 output voltage starts at lower limit, not upper limit. 

See OpenModelica [test report](https://libraries.openmodelica.org/branches/newInst/Modelica_4.0.0/files/Modelica_4.0.0_Modelica.Electrical.Analog.Examples.OpAmps.SignalGenerator.diff.opAmp1.out.v.html)

PR #3598, opened 2020-07-05
Changed homotopy parameter to be consistent

beutlich:
Back-port #3590 from master to maint/4.0.x.


beutlich:
> LGTM. I have no idea why Jenkins breaks down in this case

I chatted with @sjoelund who seems to have it figured out.

christiankral:
@beutlich is there a reason why I (as a library officer of `Electrical.Analog`) could not merge this PR?

image

beutlich:
There are no specific roles set for library officers. However, in contrast to master branch, we have restricted push access.

image

PR #3600, opened 2020-07-06
Back-port of #3291 for the pump model into MSL 3.2.3 maintenance

casella:
This PR back-ports into MSL 3.2.3 maintenance some of the changes that were made to MSL 4.0.0 in #3291, to make sure that the `PumpingSystem` example can be reliably initialized.

A homotopy expression and a parameter `checkValveClosedInit` was introduced used to manage the `CheckValve=true` case in the pump model. The code of the pump model is now exactly the same as in MSL 4.0.0 (check [here](https://github.com/modelica/ModelicaStandardLibrary/blob/e2983375f27cda0a2ca91d3add1ce8d9bf20e2ee/Modelica/Fluid/Machines.mo#L324) and  [here](https://github.com/modelica/ModelicaStandardLibrary/blob/e2983375f27cda0a2ca91d3add1ce8d9bf20e2ee/Modelica/Fluid/Machines.mo#L443)), which is the combined result of applying 1d9e75ee followed by 94fea03a. 

The new parameter `checkValveClosedInit` defaults to `noHomotopy` as suggested by @HansOlsson in #3291, hence the default behaviour is exactly the same as in released MSL 3.2.3, making the fix fully backwards compatible.

Then, in order to ensure the convergence in OpenModelica of the `PumpingSystem` example, which starts with the pump having the check valve closed, d72f405 was also applied, changing the set up of the homotopy parameter for this specific test case.

Checked with OpenModelica 1.16.0 and Dymola 2020x. The best performance is obtained when also #3599 is included.

### DIFF
--- a/Modelica/Fluid/Machines.mo
+++ b/Modelica/Fluid/Machines.mo
@@ -311,6 +311,8 @@ Then the model can be replaced with a Pump with rotational shaft or with a Presc
     parameter Medium.MassFlowRate m_flow_start = system.m_flow_start
         "Guess value of m_flow = port_a.m_flow"
       annotation(Dialog(tab = "Initialization"));
+    parameter Types.CheckValveHomotopyType checkValveHomotopy = Types.CheckValveHomotopyType.NoHomotopy "= whether the valve is Closed, Open, or unknown at initialization"
+      annotation(Dialog(tab = "Initialization"));
     final parameter SI.VolumeFlowRate V_flow_single_init = m_flow_start/rho_nominal/nParallel
         "Used for simplified initialization model";
     final parameter SI.Position delta_head_init = flowCharacteristic(V_flow_single_init*1.1)-flowCharacteristic(V_flow_single_init)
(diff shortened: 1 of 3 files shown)


Modelica/Fluid/Machines.mo
```diff
@@ -311,6 +311,8 @@ Then the model can be replaced with a Pump with rotational shaft or with a Presc
     parameter Medium.MassFlowRate m_flow_start = system.m_flow_start
         "Guess value of m_flow = port_a.m_flow"
       annotation(Dialog(tab = "Initialization"));
+    parameter Types.CheckValveHomotopyType checkValveHomotopy = Types.CheckValveHomotopyType.NoHomotopy "= whether the valve is Closed, Open, or unknown at initialization"
+      annotation(Dialog(tab = "Initialization"));
```
beutlich, line 315: Adding new components to (base) classes is - strictly seen - not legal for maintenance releases.

casella: @beutlich, I know. I'm inclined to make an exception here, since 3.2.3 is an end-of-line maintenance versions which will have no backwards compatible new releases, i.e. 3.3.0, where I can put in this change. If we want to keep 3.2.3 in the long term and keep fixing it, we have to make some compromise.

Adding homotopy here is essential to get robust intialization into 3.2.3 maintenance - the extra parameter is essential because there are different homotopies, depending on the expected initial condition, and I need to select one.

The default value corresponds to the old behaviour, so the change is 100% backwards compatible.

If you see any other solution to fix this issue, I'm fine with it. Otherwise I think it's better to have a more robust MSL 3.2.3 than strictly respecting formal rules in a borderline case as MSL 3.2.3 maintenance.

What do you think?

beutlich: It is a trade-off which I believe @MartinOtter should decide in the end.

MartinOtter: I think in this very special situation it is fine to make an exception to the formal rules and I agree with Francescos proposal.

beutlich: Usually we add such backward-compatibility breaking changes in the release notes.

casella: @beutlich, I added a new parameter to an existing model, with a default value that keeps the existing behaviour unchanged. I'm not sure I understand how this could break backwards-compatibility. Could you please elaborate?

beutlich: Please see the appropriate [documentation](https://doc.modelica.org/Modelica%204.0.0/Resources/helpOM/Modelica.UsersGuide.ReleaseNotes.VersionManagement.html) (added by bf8b927c82c024c2f3c1ae2261fa4dad84d343ec for MSL v3.1).

> As a recommendation, a valid bug-fix to the maintenance branch may contain one or more of the following changes.
> * [..]
> * Introducing a new name in the public section of a class (model, package, ...) or in any section of a partial class is **not** allowed. Since otherwise, a user might use this new name and when storing its model and loading it with an older bug-fix version, an error would occur.

casella: I think the reference to _class_ is unnecessarily too generic.

I agree this is the case if we add a new _model_ to a _package_ class. But I really can't see how this could be a problem if we add a new _parameter_ to a _model_ class. How could a user use this parameter name? The only possible cause of conflict could be if the user extended the model and added a new parameter to it, which happened to have the same name, but a different meaning. Of course we can't rule this out 100%, but I would say this is really unlikely.

Do I miss something? Should we perhaps update that documentation?

beutlich: The main reason is, that models saved in MSL v3.2.3 Build 5 will no longer be compatible with MSL v3.2.3 Build 4. Usually we claim this kind of compatibility.

casella: Aha. I always thought that being backwards compatible meant that I could use the new library with an old model, not the other way round.

But then this would also apply to minor releases. Aren't we claiming that they are backwards compatible, according to semver rules?

beutlich: The expectation is/was, that no new components are introduced between different builds of MSL v3.2.3 (or according to semver, between different patch releases of MSL v4.0.x).

MartinOtter:
Sorry, I am not good enough in github to understand the current status: Is the request from Thomas (update of release notes) included in the pull request?

"This branch is out-of-date with the base branch": Is it still fine to "merge pull request"?

beutlich:
> 
> 
> Sorry, I am not good enough in github to understand the current status: Is the request from Thomas (update of release notes) included in the pull request?
> 

No, not yet.

> "This branch is out-of-date with the base branch": Is it still fine to "merge pull request"?

I will care about it, i.e., rebase and force-push soon.

arunkumar-narasimhan:
@HansOlsson @casella, this is quite an old PR that is open even after approvals. Did this slip out somehow from earlier maintenance release?

beutlich:
If milestone [MSL4.0.1](https://github.com/modelica/ModelicaStandardLibrary/milestone/62) was closed (w/o releasing/tagging anything), then also [MSL3.2.3+maint](https://github.com/modelica/ModelicaStandardLibrary/milestone/59) should be closed. Basically @casella dropped the maintenance promise we aligned some years ago.

beutlich:
No need to keep it open.

PR #3612, opened 2020-08-03
Avoid plus character in linked image files

beutlich:
The link checker of OpenModelica uses the Python method `urllib.unquote_plus` on modelica URI which leads to false-positives findings if the link target contains plus characters.

See https://build.openmodelica.org/Documentation/FindFiles.log where it currently reads
```
Modelica.Magnetic.QuasiStatic.FundamentalWave.UsersGuide.Concept.html: Not found: Modelica 4.0.0/Resources/Images/Magnetic/QuasiStatic/FundamentalWave/Phi(ref)=Phi,re jPhi,im.png
```

Alternatively, it would be even more convenient if GenerateDoc.mos did not call `urllib.unquote_plus` on physical files in https://github.com/OpenModelica/OpenModelica/blob/81a2b5183a8e363b32fb3340bae12f2e19d8ad6f/OMCompiler/Examples/GenerateDoc.mos#L376


sjoelund:
Yes, it seems GenerateDoc.mos is incorrect.

beutlich:
@sjoelund I have just checked https://build.openmodelica.org/Documentation/FindFiles.log and can still find the false-positive. Did it trigger the rerun after your fix to GenerateDoc.mos?

sjoelund:
No, it usually lags behind 2-3 days since that particular job uses the nightly build installer; the job says on the bottom it's using https://github.com/OpenModelica/OpenModelica/commit/222a00f4b2e9d37a26111a28c9674222f4aae70b which is the commit before my fix.

beutlich:
Fix confirmed.

PR #3616, opened 2020-08-13
Update Run.mos for MultiBody examples

beutlich:
Name of model was changed from `fullRobot` to `FullRobot`.

Back-port of #3614.


beutlich:
Can't we just add a line in the script to delete the old fullRobot.mat result file.

beutlich:
> 
> 
> Or do as DagBruck writes and don't change the result-file name.

Done.

PR #3618, opened 2020-08-17
Fix result file name in Run.mos script

beutlich:
In addition to #3614, which was merged too early.


beutlich:
> 
> 
> I still believe the best solution is to leave the filename unchanged.

For maint/4.0.x, yes. For master we can change the name.

DagBruck (approved):

Issue #3593, opened 2020-06-30
Check documentation of induction machines' examples


tobolar:
In some examples located in  `Magnetic.QuasiStatic.FundamentalWave.Examples.BasicMachines.InductionMachines` there are listed variables in the documentation which can be plotted after simulation. But the variables don't exist in the model, see e.g. in `currentRMSsensor.I` or `currentSensorQS.abs_i` in `IMC_DOL`.

PR #3620, opened 2020-08-24
Refs #3593: Fix documentation of quasi-static machine examples

christiankral:
I checked the quasi static induction machine examples and updated the documentation to match the actual variables.

### DIFF
--- a/Modelica/Magnetic/QuasiStatic/FundamentalWave/Examples/BasicMachines/InductionMachines/IMC_Conveyor.mo
+++ b/Modelica/Magnetic/QuasiStatic/FundamentalWave/Examples/BasicMachines/InductionMachines/IMC_Conveyor.mo
@@ -201,9 +201,9 @@ The mechanical load is a constant torque like a conveyor (with regularization ar
 </p>
 <p>Simulate for 20 seconds and plot (versus time):</p>
 <ul>
-<li>currentQuasiRMSSensor|currentQuasiRMSSensorQS.I: stator current RMS</li>
-<li>imc|imcQS.wMechanical: machine speed</li>
-<li>imc|imcQS.tauElectrical: machine torque</li>
+<li><code>currentQuasiRMSSensor|currentQuasiRMSSensorQS.I</code>: (equivalent) stator current RMS</li>
+<li><code>imc|imcQS.wMechanical</code>: machine speed</li>
+<li><code>imc|imcQS.tauElectrical</code>: machine torque</li>
 </ul>
 <p>Default machine parameters are used.</p>
 </html>"),
--- a/Modelica/Magnetic/QuasiStatic/FundamentalWave/Examples/BasicMachines/InductionMachines/IMC_Characteristics.mo
+++ b/Modelica/Magnetic/QuasiStatic/FundamentalWave/Examples/BasicMachines/InductionMachines/IMC_Characteristics.mo
@@ -121,7 +121,7 @@ Simulate for 1 second and plot (versus <code>imcQS.wMechanical</code> or <code>s
 </p>
 
 <ul>
-<li><code>currentSensorQS.abs_i[1]</code>: (equivalent) RMS stator current</li>
+<li><code>iSensorQS.I</code>: (equivalent) RMS stator current</li>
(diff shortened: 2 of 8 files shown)


christiankral:
The following models use the instance name `iSensorQS` instead of `currentQuasiRMSSensorQS` which makes the example collection a little inconsistent 
- `Modelica.Magnetic.QuasiStatic.FundamentalWave.Examples.BasicMachines.InductionMachines.IMC_Characteristics` 
- `Modelica.Magnetic.QuasiStatic.FundamentalWave.Examples.BasicMachines.InductionMachines.IMS_Characteristics`

@tobolar Are we allowed to rename a sensor in the master branch? If yes, I would unify the appearance of the sensor names.

christiankral:
There is one more issue with a synchronous machine example which I am going to fix.

Modelica/Magnetic/QuasiStatic/FundamentalWave/Examples/BasicMachines/InductionMachines/IMC_Conveyor.mo
```diff
@@ -201,9 +201,9 @@ The mechanical load is a constant torque like a conveyor (with regularization ar
 </p>
 <p>Simulate for 20 seconds and plot (versus time):</p>
 <ul>
-<li>currentQuasiRMSSensor|currentQuasiRMSSensorQS.I: stator current RMS</li>
-<li>imc|imcQS.wMechanical: machine speed</li>
-<li>imc|imcQS.tauElectrical: machine torque</li>
+<li><code>currentQuasiRMSSensor|currentQuasiRMSSensorQS.I</code>: (equivalent) stator current RMS</li>
+<li><code>imc|imcQS.wMechanical</code>: machine speed</li>
+<li><code>imc|imcQS.tauElectrical</code>: machine torque</li>
```
tobolar, line 206: Shouldn't the options be enclosed by parentheses? Thus e.g. `<code>(currentQuasiRMSSensor|currentQuasiRMSSensorQS).I</code>`.

tobolar:
> Are we allowed to rename a sensor in the master branch? If yes, I would unify the appearance of the sensor names.

Do you mean to rename a sensor element in that particular example model?

christiankral:
> > Are we allowed to rename a sensor in the master branch? If yes, I would unify the appearance of the sensor names.
> 
> Do you mean to rename a sensor element in that particular example model?

Yes, this is what I mean

tobolar:
IMO this is no problem when only an example is concerned. I remember there has been some comment on it by (I guess) @dietmarw but I can't find it anymore.

dietmarw:
First of all you need to determine what  is meant by "master branch". The question is what next version the master branch is aiming at. Backward compatible, i.e., 4.1.0, or non-backward compatible, i.e., 5.0.0. As far as I know this has not yet been decided within MAP-LIB. In the latter case there is no restrictions on renames as long as conversion scripts are given. In the former case we tended to allow renames within examples even if they break backward compatibility since the argument is that users should not let their usermodels extend from examples (copying is fine but not extend).

AHaumer (approved):

christiankral:
OK, I propose to separate the issues which became mixed up in this PR. 

- Fix documentation: this is covered by this PR
- Change spelling (currently, e.g., `smpm|smpmQS.tauElectrical`) of [multiple similar variables](https://github.com/modelica/ModelicaStandardLibrary/pull/3620#discussion_r484284023) as suggested by @tobolar: I propose to create a different ticket and PR, respectively, as this issue affects some more models of the MSL
- Unify naming of sensor: Optionally create a new PR as this issue is not directly related with the documentation; it may also require some more discussion if we are fine with renaming instance names of MSL examples in the future

So if there are not objections with this approach, I am asking @tobolar and @dietmarw to re-evaluate this PR so that we can proceed.

dietmarw (approved):
Looks good to me. I don't think parentheses would be strictly necessary in case of the choices but would not harm either. I leave that to the LOs to decide.

dietmarw:
I've rebased the PR on current master in order to get the CI checks working.

PR #3621, opened 2020-08-26
Models should be partial, not the package

### LINKED ISSUE #3615: ModelicaServices classes extends from classes in a partial package

beutlich:
Back-port #3617 from master to fix #3615 in maint/3.2.3.

### DIFF
--- a/Modelica/Utilities/Internal.mo
+++ b/Modelica/Utilities/Internal.mo
@@ -2,7 +2,7 @@ within Modelica.Utilities;
 package Internal
   "Internal components that a user should usually not directly utilize"
   extends Modelica.Icons.InternalPackage;
-partial package PartialModelicaServices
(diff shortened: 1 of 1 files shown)


HansOlsson (approved):
Looks good

tobolar (approved):

PR #3639, opened 2020-10-02
Fix missing include in Common.c (CI)

beutlich:
I got a compile error when using Common.c in MSVC environment otherwise.


sjoelund (approved):

HansOlsson (approved):
Looks good: Since assert is used assert.h should be included.

Issue #2001, opened 2017-01-15
ModelicaExternalC not compatible with hdf5 1.10


modelica-trac-importer:
**Reported by perost on 31 May 2016 14:05 UTC**
ModelicaExternalC doesn't build with the latest release of the hdf5 library, since the arguments to the macro H5Rdereference has changed¹. When trying to build ModelicaExternalC (using autotools) I get the following errors with hdf5 1.10:
```
../../C-Sources/ModelicaMatIO.c:18392:69: error: too few arguments to function call, expected 4, have 3
                ref_id = H5Rdereference(dset_id,H5R_OBJECT,ref_ids+i);
                         ~~~~~~~~~~~~~~                             ^
/usr/include/H5Rpublic.h:63:8: note: 'H5Rdereference2' declared here
H5_DLL hid_t H5Rdereference2(hid_t obj_id, hid_t oapl_id, H5R_type_t ref_type, const void *ref);
       ^
<... several more errors saying the same thing ...>
```

¹ https://www.hdfgroup.org/HDF5/doc/RM/RM_H5R.html#Reference-Dereference

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/2001

modelica-trac-importer:
**Comment by beutlich on 31 May 2016 14:29 UTC**
This is a duplicate of https://github.com/tbeu/matio/issues/37. It will be fixed by the provided patch (upstream) once matio v1.5.7 is released such that I can merge it to the external C code.

modelica-trac-importer:
**Modified by beutlich on 31 May 2016 19:31 UTC**

modelica-trac-importer:
**Comment by beutlich on 21 Jun 2016 19:58 UTC**
Resolved by ab69b8c1c2de2b7406d2a79f9a850c98020aca90. Sorry for the delay.

modelica-trac-importer:
**Changelog modified by beutlich on 21 Jun 2016 19:58 UTC**
Updated matio library to v1.5.7 (ModelicaMatIO.[ch])

modelica-trac-importer:
**Comment by beutlich on 3 Jul 2016 20:10 UTC**
Updated ModelicaMatIO.[ch] to v1.5.8 by 4ae6fed3f1420af0d77c463231d22e23c099da74.

modelica-trac-importer:
**Changelog modified by beutlich on 3 Jul 2016 20:10 UTC**
[...]

PR #3641, opened 2020-10-05
Update matio library to version 1.5.19

beutlich:
This updates ModelicaMatIO to the latest version 1.5.18 of the underlying matio library. The changes in the diff are hardly trackable, thus we probably should revert to use original matio source files one day.

Matio 1.5.8 supports HDF5 1.12.x, which might be relevant for OMC (see #2001 for some previous issue with HDF5 support).

Unfortunately, the new version no longer compiles with gcc 4.6 due an internal compiler error (see https://travis-ci.org/github/tbeu/ModelicaStandardTables/jobs/732885392#L463). This compiler error does not show up with gcc 4.4 or 4.8, thus workarounds are available.

I can provide the new binaries for testing in a separate commit using a dedicated branch.


beutlich:
> I can provide the new binaries for testing in a separate commit using a dedicated branch.

Binaries for early testing have been rebuilt and are available from my branch [update-binaries-matio-1.5.18](https://github.com/beutlich/ModelicaStandardLibrary/tree/update-binaries-matio-1.5.18).

HansOlsson (approved):
Looks good to me.

beutlich:
Will soon update to v1.5.19 (to be released tonight) due to one issue with undefined behaviour reported by @jnjaeschke.

sjoelund (approved):
Seems OK. Also works on gcc-4.6 (I found an unrelated issue with the configure.ac file that I will open a PR for)

beutlich:
Supersedeb by #3785.

PR #3642, opened 2020-10-06
Include provided header files (CI)

beutlich:
There is no need to repeat the declaration if we can include the provided header files from C-Sources.

See also #3453 wehere we set Include and IncludeDirectory annotations to every external function.


sjoelund (approved):

HansOlsson (approved):
Looks good.

PR #3646, opened 2020-10-06
Blocks.Routing.Extractor: add assert for input range

beutlich:
Backport #3627 from master to maint/3.2.3.


StephanZiegler (approved):
Works the same way as on the other branch.

HansOlsson (approved):
Looks good

PR #3648, opened 2020-10-08
Add Version 4.0.0 documentation generated by System Modeler

maltelenz:
* Generated documentation for MSL 4.0.0 using Wolfram System Modeler
* Updated index.html to include the new version
* Changed the display name in selection dropdown from "SystemModeler" to "System Modeler" with a space. This is a branding change from Wolfram.

I see in #3636 that the sitemap needs updating as well, but I don't have the tools or process for doing so.


beutlich (approved):

maltelenz:
Thank you for catching the bug with the `Licenses` directory link. I have reported it internally so we can generate the proper thing directly next time.

PR #3658, opened 2020-11-04
Added start value to ensure convergence to the right flow in suddenExpansion1

### LINKED ISSUE #3656: The reference solution of Modelica.Fluid.Examples.Explanatory.MomentumBalanceFittings is wrong
The `Modelica.Fluid.Examples.Explanatory.MomentumBalanceFittings` model shows some potentially counterintutive issues with momentum balances when dealing with adaptors having different inlet and outlet diameters. As explained in the documentation, the fluid is expected to flow through `suddenExpansion1` from left to right, i.e., to be positive. Unfortunately, the provided reference value is negative, so it is not the intended one.

Some analysis of the involved momentum equation for the upper system shows that in fact both a positive and a negative solution are possible. Indeed, by selecting different start values for the mass flow through `suddenExpansion1` it is possible to obtain two solutions: 53.7 and -30.3 kg/s. The current source code leaves the start value to zero, which explain why Dymola converged to the (wrong) negative solution. I guess in the past it should have converged to the positive one, otherwise this example would not have made sense, probably we missed the regression many years ago.

In any case, since there are two possible solutions and only one is the intended one for the purpose of this example, it is necessary to add a proper start value to ensure convergence to the right one.

casella:
Fixes #3656 in the 3.2.3 maintenance branch

### DIFF
--- a/Modelica/Fluid/Examples/Explanatory.mo
+++ b/Modelica/Fluid/Examples/Explanatory.mo
@@ -197,6 +197,7 @@ present as for T_twoPort.T.
     Modelica.Fluid.Fittings.AbruptAdaptor suddenExpansion1(
       diameter_a=0.1,
       diameter_b=0.2,
+      m_flow_start=60,
       redeclare package Medium = Modelica.Media.Water.StandardWaterOnePhase,
       show_totalPressures=true,
       show_portVelocities=true)


thorade (approved):

mwetter (approved):

casella:
@beutlich, for some reason I do not have the rights to merge on this branch. Can you please give it to me?

casella:
@kadbdelhak, when this is merged the OpenModelica testsuite on 3.2.3 should be automatically updated, we refer to the maintenance branch.

PR #3663, opened 2020-11-11
Add getcwd fallback in ModelicaInternal_fullPathName

### LINKED ISSUE #3660: Modelica.Utilities.Files.createDirectory fails if the directory does not exist
`Modelica.Utilities.Files.createDirectory` calls `fullPathName`, which fails if the path does not exist (on Linux). This means createDirectory does not work. Some possible solutions:

* Change `fullPathName` to never fail (revert to the old code with `getcwd`)
* Change `createDirectory` to not call `fullPathName` (but this would not fix older MSL versions)
* Add an argument to `fullPathName` to not fail (return the input perhaps), and create this additional file in the external sources

beutlich:
Check if realpath fails for non-existing path and add getcwd fallback. Especially, do not call ModelicaError for non-existing path.

Closes #3660 for OM#6204.

### DIFF
--- a/Modelica/Resources/C-Sources/ModelicaInternal.c
+++ b/Modelica/Resources/C-Sources/ModelicaInternal.c
@@ -30,6 +30,10 @@
 */
 
 /* Changelog:
+      Nov. 11, 2020: by Thomas Beutlich
+                     Added getcwd fallback in ModelicaInternal_fullPathName if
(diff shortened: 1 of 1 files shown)


HansOlsson (approved):
Seems ok.

sjoelund (approved):
I didn't realise you already fixed this. It was on my list for this week. I tested the code and it works fine.

PR #3668, opened 2020-11-17
maint/3.2.3: Add getcwd fallback in ModelicaInternal_fullPathName

### LINKED ISSUE #3660: Modelica.Utilities.Files.createDirectory fails if the directory does not exist

beutlich:
Back-port #3663 from master to fix #3660 in maint/3.2.3.

### DIFF
--- a/Modelica/Resources/C-Sources/ModelicaInternal.c
+++ b/Modelica/Resources/C-Sources/ModelicaInternal.c
@@ -30,6 +30,10 @@
 */
 
 /* Release Notes:
+      Nov. 11, 2020: by Thomas Beutlich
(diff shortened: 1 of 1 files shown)


HansOlsson (approved):
Looks good.

sjoelund (approved):

PR #3669, opened 2020-11-17
Fix reading file with EOL=LF on Windows

beutlich:
This PR superseds #3631 and fixes reading of text files with EOL=LF on Windows. It also adds unit tests for these functions.

I decided to duplicate the code of function `readLine` (from ModelicaIO) since we have no concept how to deal with utility functions that are shared between different C libraries. (I also did so once in the past by duplicating function `transpose` between ModelicaIO and ModelicaStandarTables C libraries.)


sjoelund (commented):
I don't have a Windows machine to test this on and as far as I can see, the mingw test only compiles the code and does not run tests. Perhaps we should run this through wine or something in the future.

beutlich:
I tested and debugged locally on my Win machine. I hope that @HansOlsson can also test on Win.

sjoelund (approved):

HansOlsson (approved):
Seems to build on Windows, and the test-case triggering the original issue worked;
so it looks good.

PR #3670, opened 2020-11-18
Do not add -Werror if found flag exists

sjoelund:


beutlich (commented):
Duplicate of #3580.

sjoelund:
So since I keep running into the same bug, perhaps approve the other PR? :)

PR #3679, opened 2020-12-03
Implement choice between sawtooth and triangle reference signal in PowerConverters/DCDC/Control/SignalPWM

AHaumer:
This is a backwards compatible enhancement:
For discrete and quasi-continuous control it is essential to sample the current at the right time instant during the pwm period to catch the mean current. Using triangle reference signal, it is easy to sample exactly in the middle of the pwm period.

### DIFF
--- a/Modelica/Electrical/PowerConverters/DCDC/Control/SignalPWM.mo
+++ b/Modelica/Electrical/PowerConverters/DCDC/Control/SignalPWM.mo
@@ -2,11 +2,14 @@ within Modelica.Electrical.PowerConverters.DCDC.Control;
 model SignalPWM
   "Generates a pulse width modulated (PWM) boolean fire signal"
   extends Icons.Control;
+  import Modelica.Electrical.PowerConverters.Types.SingleReferenceType;
   parameter Boolean useConstantDutyCycle=true
     "Enables constant duty cycle";
   parameter Real constantDutyCycle=0 "Constant duty cycle"
     annotation (Dialog(enable=useConstantDutyCycle));
   parameter SI.Frequency f=1000 "Switching frequency";
+  parameter SingleReferenceType refType=Modelica.Electrical.PowerConverters.Types.SingleReferenceType.Sawtooth
+    "Type of reference signal";
   parameter SI.Time startTime=0 "Start time";
   Modelica.Blocks.Interfaces.RealInput dutyCycle if not
     useConstantDutyCycle "Duty cycle"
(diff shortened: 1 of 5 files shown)


dietmarw (approved):
Looks good.

Modelica/Electrical/PowerConverters/DCDC/Control/SignalPWM.mo
```diff
@@ -2,11 +2,14 @@ within Modelica.Electrical.PowerConverters.DCDC.Control;
 model SignalPWM
   "Generates a pulse width modulated (PWM) boolean fire signal"
   extends Icons.Control;
+  import Modelica.Electrical.PowerConverters.Types.SingleReferenceType;
   parameter Boolean useConstantDutyCycle=true
     "Enables constant duty cycle";
   parameter Real constantDutyCycle=0 "Constant duty cycle"
     annotation (Dialog(enable=useConstantDutyCycle));
   parameter SI.Frequency f=1000 "Switching frequency";
+  parameter SingleReferenceType refType=Modelica.Electrical.PowerConverters.Types.SingleReferenceType.Sawtooth
```
tobolar, line 11: Possibly `` ... refType=SingleReferenceType.Sawtooth`` ?

AHaumer:
Omitting the full path to the option could lead to problems in some tools.
I'd prefer to keep it as it is.

Am Fr., 8. Jan. 2021 um 09:34 Uhr schrieb tobolar <notifications@github.com

Modelica/Electrical/PowerConverters/DCDC/Control/SignalPWM.mo
AHaumer: Omitting the full path to the option could lead to problems in some tools.
If you're ok with keeoing the full path please add your approval.

beutlich: I agree with @tobolar that it does not look that consistent. You import `Modelica.Electrical.PowerConverters.Types.SingleReferenceType` and use it as component type, but not in the assignment. I you want to keep the fully qualified name in the declaration, you should also remove the import to make it consistent.

beutlich:
Please be aware that I rebased and force-pushed most of the open PR branches to update the CI which is now running through GitHub Actions.

christiankral (approved):
Looks good.

Modelica/Electrical/PowerConverters/DCDC/Control/SignalPWM.mo
beutlich: Apparently this conversation was not resolved before the merge. 😕

AHaumer: It was resolved. Once again ... The problem is: 
The use of the imported type is more convenient inside the component, that's right. 
But using the imported type in setting the default value of the parameter causes (or maybe caused?) problems in some simulators in the user interface. 
So why not using the fully qualified path in setting the default parameter? What's the problem?

PR #3688, opened 2020-12-15
Update 3rd-party header uthash to v2.2.0

beutlich:
This updates to uthash v2.2.0 and does some refactoring on inclusion of stdint.h.

### DIFF
--- a/Modelica/Resources/C-Sources/ModelicaInternal.c
+++ b/Modelica/Resources/C-Sources/ModelicaInternal.c
@@ -207,6 +207,32 @@ void ModelicaInternal_setenv(_In_z_ const char* name,
   #define _POSIX_ 1
 #endif
 
+/* Have the <stdint.h> header file */
+#if defined(_WIN32)
+#if defined(_MSC_VER) && _MSC_VER >= 1600
+#define HAVE_STDINT_H 1
+#elif defined(__WATCOMC__) || defined(__MINGW32__) || defined(__CYGWIN__)
+#define HAVE_STDINT_H 1
+#else
+#undef HAVE_STDINT_H
+#endif
+#elif defined(__GNUC__) && !defined(__VXWORKS__)
+#define HAVE_STDINT_H 1
+#else
+#undef HAVE_STDINT_H
+#endif
+
+/* Replacement for integer type header */
+#if !defined(HAVE_STDINT_H)
+#define HASH_NO_STDINT 1
+#if defined(_MSC_VER)
+#include "stdint_msvc.h"
+#else
+#define uint32_t unsigned int
+#define uint8_t unsigned char
+#endif
+#endif
+
 #define HASH_NONFATAL_OOM 1
 #include "uthash.h"
 #include "gconstructor.h"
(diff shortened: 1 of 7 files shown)


Modelica/Resources/C-Sources/ModelicaInternal.c
```diff
@@ -207,6 +207,32 @@ void ModelicaInternal_setenv(_In_z_ const char* name,
...
+/* Replacement for integer type header */
+#if !defined(HAVE_STDINT_H)
+#define HASH_NO_STDINT 1
+#if defined(_MSC_VER)
+#include "stdint_msvc.h"
+#else
+#define uint32_t unsigned int
+#define uint8_t unsigned char
+#endif
+#endif
```
Quuxplusone, line 234: FWIW, I recommend using `typedef` instead of `#define`.
And since you're already providing [your own "stdint_msvc.h"](https://github.com/modelica/ModelicaStandardLibrary/blob/master/Modelica/Resources/C-Sources/stdint_msvc.h), have you considered just always including it? It could simplify lines 225–234 to just

```
#if !defined(HAVE_STDINT_H)
#define HASH_NO_STDINT 1
#include "stdint_msvc.h"
#endif
```

Or, if you rewrite "stdint_msvc.h" to do all these `#if`s and then conditionally include the real `<stdint.h>`, you could replace all this with simply

    #include "stdint_msvc.h"  // either include <stdint.h> or polyfill what we need from it
    #define HASH_NO_STDINT 1  // because we've already included it
    #include "uthash.h"

beutlich: > 
> 
> FWIW, I recommend using `typedef` instead of `#define`.

There was some issue with some legacy (Borland ?) compiler not supporting `typedef`. I could not reproduce locally when I have just tested it, but it in no harm to stick with `#define` here.

> And since you're already providing [your own "stdint_msvc.h"](https://github.com/modelica/ModelicaStandardLibrary/blob/master/Modelica/Resources/C-Sources/stdint_msvc.h), have you considered just always including it?

That will break with Borland compiler on Win since stdint_msvc.h errors for other compilers than MSVC. Therefore I want to keep the fallback.

Quuxplusone: > since stdint_msvc.h errors for other compilers than MSVC

I was proposing that you should change stdint_msvc.h so that _it alone_ does all the `#if`s you're currently duplicating here — i.e., after those changes, it _shouldn't_ error for non-MSVC compilers, and anytime you need anything from <stdint.h>, you can just unconditionally `#include "stdint_msvc.h"` (or maybe rename to `#include "stdint_polyfill.h"`). This isn't just a uthash issue for you, right? it applies to anyplace you use <stdint.h>? So it would make sense to factor out as much of that logic to a single place as possible.

beutlich: Right, Good point. Going to refactor to avoid code duplication.

Quuxplusone:
Yes, that looks like what I had in mind. :)

HansOlsson (approved):
Looks good

sjoelund (approved):

PR #3692, opened 2020-12-22
Add tableOnFile parameter to icon of CombiTable blocks

beutlich:
This PR adds tableOnFile as the primary parameter to the icons of the table blocks.


HansOlsson (approved):
Looks good, even though most of the changes are just removing redundant ".0" - which possibly should be a separate PR.

beutlich:
> 
> 
> Looks good, even though most of the changes are just removing redundant ".0" - which possibly should be a separate PR.

Right. See #3720.

PR #3696, opened 2020-12-23
Fix some compiler warnings

beutlich:


HansOlsson (approved):
Ok, makes sense.
I was first confused, but then I realized that the compiler warning must be that the initializer is unconditionally over-written by the assignment and thus unnecessary. I almost find such compiler warnings excessive, but on the other hand I don't like unnecessary initialization either.

sjoelund (approved):
It's fine, but some other tools complain that you don't initialize the variable. Which is a problem if you write new code that uses it. It's impossible to appease all tool warnings.

PR #3700, opened 2021-01-01
Make array inputs of external C functions const (introduced in Modelica 3.5)

### LINKED ISSUE #3839: External ModelicaStandardTables lack const qualifiers for array inputs (Modelica Language Specification version 3.5)
The Modelica spec 3.5 finally added const qualifiers for the mapping of array inputs to external C functions -- see [3.5](https://specification.modelica.org/maint/3.5/functions.html#arrays-1) vs [3.4](https://specification.modelica.org/maint/3.4/Ch12.html#arrays-1)
ModelicaStandardTables should follow the standard to not break the compilation of generated code.

For instance `ModelicaStandardTables_CombiTable2D_init2` is declared without const qualifier for the table argument in [ModelicaStandardTables.h](https://github.com/modelica/ModelicaStandardLibrary/blob/master/Modelica/Resources/C-Sources/ModelicaStandardTables.h):
```c
MODELICA_EXPORT void* ModelicaStandardTables_CombiTable2D_init2(_In_z_ const char* fileName,
                                                _In_z_ const char* tableName,
                                                _In_ double* table, size_t nRow,
                                                size_t nColumn, int smoothness,
                                                int extrapolation,
                                                int verbose) MODELICA_NONNULLATTR;
```

`Modelica.Blocks.Types.ExternalCombiTable2D` calls this function with the array input table though:
```
class ExternalCombiTable2D
  "External object of 2-dim. table defined by matrix"
  extends ExternalObject;

  function constructor "Initialize 2-dim. table defined by matrix"
    input String tableName "Table name";
    input String fileName "File name";
    input Real table[:, :];
    input Modelica.Blocks.Types.Smoothness smoothness;
    input Modelica.Blocks.Types.Extrapolation extrapolation=Modelica.Blocks.Types.Extrapolation.LastTwoPoints;
    input Boolean verboseRead=true "= true: Print info message; = false: No info message";
    output ExternalCombiTable2D externalCombiTable2D;
  external "C" externalCombiTable2D = ModelicaStandardTables_CombiTable2D_init2(
          fileName,
          tableName,
          table,
          size(table, 1),
          size(table, 2),
          smoothness,
          extrapolation,
          verboseRead) annotation (IncludeDirectory="modelica://Modelica/Resources/C-Sources", Include="#include \"ModelicaStandardTables.h\"", Library={"ModelicaStandardTables", "ModelicaIO", "ModelicaMatIO", "zlib"});
  end constructor;

  ...
end ExternalCombiTable2D;
```

beutlich:
Can only be merged once MSL is based on MLS 3.5.

Resolves #3839.

### DIFF
--- a/Modelica/Blocks/Tables.mo
+++ b/Modelica/Blocks/Tables.mo
@@ -1330,7 +1330,7 @@ protected
     external \"C\" dummy_y = mydummyfunc(dummy_u);
     annotation(IncludeDirectory=\"modelica://Modelica/Resources/Data/Tables\",
            Include = \"#include \"usertab.c\"
-double mydummyfunc(double* dummy_in) {
+double mydummyfunc(const double* dummy_in) {
    return 0;
 }
 \");
--- a/Modelica/Resources/C-Sources/ModelicaFFT.c
+++ b/Modelica/Resources/C-Sources/ModelicaFFT.c
@@ -518,7 +518,7 @@ static void mrkiss_fftr(mrkiss_fftr_cfg st, const mrkiss_fft_scalar *timedata, m
     }
 }
 
-int ModelicaFFT_kiss_fftr(_In_ double* u, size_t nu, _In_ double* work, size_t nwork,
+int ModelicaFFT_kiss_fftr(_In_ const double* u, size_t nu, _In_ double* work, size_t nwork,
                           _Out_ double *amplitudes, _Out_ double *phases) {
 
     /* Compute real FFT with mrkiss_fftr
--- a/Modelica/Resources/C-Sources/ModelicaFFT.h
+++ b/Modelica/Resources/C-Sources/ModelicaFFT.h
@@ -69,7 +69,7 @@
 #define _Out_
 #endif
 
-MODELICA_EXPORT int ModelicaFFT_kiss_fftr(_In_ double* u, size_t nu, _In_ double* work, size_t nwork,
+MODELICA_EXPORT int ModelicaFFT_kiss_fftr(_In_ const double* u, size_t nu, _In_ double* work, size_t nwork,
     _Out_ double *amplitudes, _Out_ double *phases) MODELICA_NONNULLATTR;
 
 #endif
--- a/Modelica/Resources/C-Sources/ModelicaIO.c
+++ b/Modelica/Resources/C-Sources/ModelicaIO.c
@@ -112,7 +112,7 @@ void ModelicaIO_readRealMatrix(_In_z_ const char* fileName,
     int verbose) {
     ModelicaNotExistError("ModelicaIO_readRealMatrix"); }
 int ModelicaIO_writeRealMatrix(_In_z_ const char* fileName,
-    _In_z_ const char* matrixName, _In_ double* matrix, size_t m, size_t n,
+    _In_z_ const char* matrixName, _In_ const double* matrix, size_t m, size_t n,
     int append, _In_z_ const char* version) {
     ModelicaNotExistError("ModelicaIO_writeRealMatrix"); return 0; }
 double* ModelicaIO_readRealTable(_In_z_ const char* fileName,
@@ -290,7 +290,7 @@ void ModelicaIO_readRealMatrix(_In_z_ const char* fileName,
 
 int ModelicaIO_writeRealMatrix(_In_z_ const char* fileName,
                                _In_z_ const char* matrixName,
-                               _In_ double* matrix, size_t m, size_t n,
+                               _In_ const double* matrix, size_t m, size_t n,
                                int append,
                                _In_z_ const char* version) {
     int status;
--- a/Modelica/Resources/C-Sources/ModelicaIO.h
+++ b/Modelica/Resources/C-Sources/ModelicaIO.h
@@ -112,7 +112,7 @@ MODELICA_EXPORT void ModelicaIO_readRealMatrix(_In_z_ const char* fileName,
 
 MODELICA_EXPORT int ModelicaIO_writeRealMatrix(_In_z_ const char* fileName,
                                _In_z_ const char* matrixName,
-                               _In_ double* matrix, size_t m, size_t n,
+                               _In_ const double* matrix, size_t m, size_t n,
                                int append,
                                _In_z_ const char* version) MODELICA_NONNULLATTR;
   /* Write matrix to file
--- a/Modelica/Resources/C-Sources/ModelicaRandom.c
+++ b/Modelica/Resources/C-Sources/ModelicaRandom.c
@@ -122,7 +122,7 @@ static void ModelicaRandom_deleteCS(void) {
 #define ModelicaRandom_INVM64 5.42101086242752217004e-20 /* = 2^(-64) */
 #define ModelicaRandom_RAND(INT64) ( (int64_t)(INT64) * ModelicaRandom_INVM64 + 0.5 )
 
-void ModelicaRandom_xorshift64star(_In_ int* state_in,
+void ModelicaRandom_xorshift64star(_In_ const int* state_in,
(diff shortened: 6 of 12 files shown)


sjoelund (approved):
It should be possible to merge this even if the MSL was based on MLS 3.4 since the code should still be backwards compatible. The code as it is today breaks tools that pass const data.

beutlich:
> 
> 
> It should be possible to merge this even if the MSL was based on MLS 3.4 since the code should still be backwards compatible. The code as it is today breaks tools that pass const data.

I do not think this should be done as some tools generate the function prototypes according to MLS 3.4.

sjoelund:
Then the best solution would be to not include the prototypes at all (like it was in MSL 3.2.3). Because it is really annoying to try to use libraries that assume MLS 3.5 in their prototypes at the same time as using models from the MSL...

beutlich:
It is a breaking change, which means a tool/library cannot be compliant to MLS 3.4 and 3.5 at the same time.

sjoelund:
A library can be compliant simply by not using the include annotation

beutlich:
If a tool generates the function prototypes they need to match the actual function implementation. I believe w/o a proper fix of Mo#1726 the dilemma cannot be resolved,

sjoelund:
> If a tool generates the function prototypes they need to match the actual function implementation.

No, it doesn't. As long as the implementation does not modify the inputs, it will work fine in both cases.

> I believe w/o a proper fix of Mo#1726 the dilemma cannot be resolved,

There could be other solutions for this as well. One would be that the tool tells the C-code which language standard it is using. Like `-DMODELICA_LANGUAGE_VERSION=0x030500`

beutlich:
grafik

Some compilers are strict and require consistent types for declararion and definition.

Anyway, this PR must not be merged right now.

sjoelund:
> Some compilers are strict and require consistent types for declararion and definition.
> 
> Anyway, this PR must not be merged right now.

If there is no Include annotation, there is no conflict. And the actual interface is the same and will work fine in both Modelica 3.4 and 3.5 compilers.

henrikt-ma:
> If a tool generates the function prototypes they need to match the actual function implementation. I believe w/o a proper fix of [Mo#1726](https://github.com/modelica/ModelicaSpecification/issues/1726) the dilemma cannot be resolved,

But tools shall not generate prototypes if an `Include`-annotation is given, see last paragraph before https://specification.modelica.org/master/functions.html#argument-type-mapping.  Hence, one just has to make sure that the all the MSL external functions have `Include`-annotations.

sjoelund:
> But tools shall not generate prototypes if an `Include`-annotation is given, see last paragraph before https://specification.modelica.org/master/functions.html#argument-type-mapping. Hence, one just has to make sure that the all the MSL external functions have `Include`-annotations.

Yes, but then the prototype would need to be according to MLS 3.5. As it is, one of our code generators tries to use the prototype and calls it with a constant vector... But it's declared as modifying the vector.

henrikt-ma:
> Yes, but then the prototype would need to be according to MLS 3.5. As it is, one of our code generators tries to use the prototype and calls it with a constant vector... But it's declared as modifying the vector.

OK, then I understand, and agree that this can only be merged once we're using MLS 3.5.

sjoelund:
> OK, then I understand, and agree that this can only be merged once we're using MLS 3.5.

I think it could be merged before then because you can call the MLS 3.5 prototype with inputs corresponding to MLS 3.4. MLS 3.5 is more permissive in what inputs are allowed (no MLS 3.4 tool can pass constant vectors as input, so the opposite direction is no problem). This PR should work fine in C/C++ with -Wall -Werror when passing non-const inputs.

There are some alternatives.
1. Just merge this. It's fine.
2. Modify this to also remove `Include` annotation. This is also fine, and works in tools that both generate prototype and use the Include annotation at the same time (really not recommended)
3. Just change to MLS 3.5 in MSL master as soon as possible.

henrikt-ma:
> * Just merge this. It's fine.

I'd call it incorrect, but expected to work.

> * Modify this to also remove `Include` annotation. This is also fine, and works in tools that both generate prototype and use the Include annotation at the same time (really not recommended)

Clang doesn't like conflicting prototypes (the non-const corresponding to the automatically generated MLS 3.4 prototype):
```
const.c:2:5: error: conflicting types for 'foo'
int foo(const int* x);
    ^
const.c:1:5: note: previous declaration is here
int foo(int* x);
    ^
```

sjoelund:
> Clang doesn't like conflicting prototypes (the non-const corresponding to the automatically generated MLS 3.4 prototype):
> 
> ```
> const.c:2:5: error: conflicting types for 'foo'
> int foo(const int* x);
>     ^
> const.c:1:5: note: previous declaration is here
> int foo(int* x);
>     ^
> ```

That's why I said it's an option to remove the Include annotation. Then there will be no conflict as only the automatically generated would be present.

sjoelund:
Oh yeah, we could also have

4. Introduce some magic preprocessor directive available in MLS 3.6+ so the code could add const to the prototype if it knows that the tool expects more recent MLS...

henrikt-ma:
> That's why I said it's an option to remove the Include annotation. Then there will be no conflict as only the automatically generated would be present.

As @tbeu pointed out, a nice compiler will reject an implementation that doesn't match the prototype.  My point was that a correct prototype cannot coexist at all with the incorrect prototype generated by the tool, which could cause problems if the generated prototype would end up in a compilation unit where some other external function's `Include`-annotation would pull in the correct prototype.

sjoelund:
Yes, it would indeed be unfortunate to have a mix of Include or not Include annotations being used. And thinking about it, there are libraries in the wild that use libraries from the MSL in their own external "C" functions...

rfranke (commented):
This looks like long overdue as the functions must not change their inputs. This has also been clarified with Modelica spec 3.5. As discussed in issue #3839, the const qualifiers don't affect the use of MSL. They might affect tools that generate conflicting function prototypes without const -- but this must not happen since the introduction of Include annotations in MSL.

rfranke (approved):

dietmarw (changes requested):
Blocking accidental merge. Need to wait until master is based on MLS 3.~~5~~6.

HansOlsson:
> Blocking accidental merge. Need to wait until master is based on MLS 3.5.

This has now been changed (to 3.6 to be specific).
@dietmarw can you update?

beutlich:
PR should be rebased first and missing functions (e.g., ModelicaStandardTables_CombiTimeTable_init3) need to be taken into account.

dietmarw:
It would help if there was a formal decision what MSL version would now allow MLS 3.6 features so the milestone of this PR can be set accordingly.

maltelenz:
> It would help if there was a formal decision what MSL version would now allow MLS 3.6 features so the milestone of this PR can be set accordingly.

Same comment as https://github.com/modelica/ModelicaStandardLibrary/pull/3686#issuecomment-1853434631

beutlich:
> PR should be rebased first and missing functions (e.g., ModelicaStandardTables_CombiTimeTable_init3) need to be taken into account.

Rebased and applied to more occurrences, see https://github.com/modelica/ModelicaStandardLibrary/pull/3700/commits/e66d5aad0459004d18f1f1cda7c8ca15a023bb28.

casella:
@sjoelund, @rfranke, @HansOlsson I need two last positive reviews to merge this into 4.1.0

Deadline: 02 Feb 2024.

Thanks!

HansOlsson (approved):
Looks good.

sjoelund (approved):
Looks fine to me (but I have not tested the compilation of it on OpenModelica since I'm on parental leave and don't have the time for it)

PR #3703, opened 2021-01-05
Avoid deprecated "builtin" external language for common mathematical functions

beutlich:
Modelica Language Specification 3.5 deprecates the usage of the "builtin" external language.

### DIFF
--- a/Modelica/Math/package.mo
+++ b/Modelica/Math/package.mo
@@ -11030,7 +11030,7 @@ function sin "Sine"
   output Real y "Dependent variable y=sin(u)";
 algorithm
   y := .sin(u);
-  annotation (
+  annotation (Inline=true,
     Icon(coordinateSystem(
         preserveAspectRatio=true,
         extent={{-100,-100},{100,100}}), graphics={
@@ -11066,7 +11066,7 @@ function cos "Cosine"
   output Real y "Dependent variable y=cos(u)";
 algorithm
   y := .cos(u);
-  annotation (
+  annotation (Inline=true,
     Icon(coordinateSystem(
         preserveAspectRatio=true,
         extent={{-100,-100},{100,100}}), graphics={
@@ -11102,7 +11102,7 @@ function tan "Tangent (u shall not be -pi/2, pi/2, 3*pi/2, ...)"
   output Real y "Dependent variable y=tan(u)";
 algorithm
   y := .tan(u);
-  annotation (
+  annotation (Inline=true,
     Icon(coordinateSystem(
         preserveAspectRatio=true,
         extent={{-100,-100},{100,100}}), graphics={
@@ -11138,7 +11138,7 @@ function asin "Inverse sine (-1 <= u <= 1)"
   output Modelica.Units.SI.Angle y "Dependent variable y=asin(u)";
 algorithm
   y := .asin(u);
-  annotation (
+  annotation (Inline=true,
     Icon(coordinateSystem(
         preserveAspectRatio=true,
         extent={{-100,-100},{100,100}}), graphics={
@@ -11172,7 +11172,7 @@ function acos "Inverse cosine (-1 <= u <= 1)"
   output Modelica.Units.SI.Angle y "Dependent variable y=acos(u)";
 algorithm
   y := .acos(u);
-  annotation (
+  annotation (Inline=true,
     Icon(coordinateSystem(
         preserveAspectRatio=true,
         extent={{-100,-100},{100,100}}), graphics={
@@ -11206,7 +11206,7 @@ function atan "Inverse tangent"
   output Modelica.Units.SI.Angle y "Dependent variable y=atan(u)";
 algorithm
   y := .atan(u);
-  annotation (
+  annotation (Inline=true,
     Icon(coordinateSystem(
         preserveAspectRatio=true,
         extent={{-100,-100},{100,100}}), graphics={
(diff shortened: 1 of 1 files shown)


henrikt-ma (changes requested):
These shouldn't be external functions at all.  Thereby, we don't assume mathematical understanding about all the functions you might find in external libraries.  Instead, these should just be wrappers around the elementary mathematical functions of the core language.

beutlich:
@henrikt-ma @HansOlsson There is now an infinite lookup recursion. Am I doing the right thing?

henrikt-ma (changes requested):
As you noticed yourself, you need to use fully qualified references to the built-in functions:
```
y := .sin(u);
```

henrikt-ma (approved):
Looks good to me.

(I have a personal preference for outputs defined by binding equation since functions written in this form tend to be easier to reason about mathematically in general.)

henrikt-ma:
Actually, I'll withdraw my approval, since I'd like us to first discuss `Inline = true`.  Any reason this annotation shouldn't be used here?

henrikt-ma (commented):
Actually, I'll withdraw my approval, since I'd like us to first discuss `Inline = true`.  Any reason this annotation shouldn't be used here?

beutlich:
> 
> 
> Actually, I'll withdraw my approval, since I'd like us to first discuss `Inline = true`. Any reason this annotation shouldn't be used here?

For one-line assignments the default inlining should be true, at least in Dymola and SimulationX.

henrikt-ma:
> For one-line assignments the default inlining should be true, at least in Dymola and SimulationX.

As far as I can tell, this isn't prescribed by the specification, so I don't really see why one shouldn't set `Inline = true`.  I'm requesting additional reviews to collect more opinions.  For me, a good resolution would be to leave out `Inline` here, combined with a specification issue about describing situations when `Inline = true` is default.

eshmoylova (approved):
Looks good to me. We would inline them the way they are written. But if other tools would prefer having `Inline = true` explicit I have to objections to that.

sjoelund (changes requested):
I think I would prefer `Inline=true` explicitly added since it is possible to generate a diagnostic message if the tool fails to inline the call and in these cases if it fails, a tool that does not inline it also likely doesn't add an automatic derivative function. These functions are used quite a lot and adding the annotation makes it explicit that this must be inlined or results may be bad or performance poor.

henrikt-ma:
> I think I would prefer `Inline=true` explicitly added since it is possible to generate a diagnostic message if the tool fails to inline the call and in these cases if it fails, a tool that does not inline it also likely doesn't add an automatic derivative function. These functions are used quite a lot and adding the annotation makes it explicit that this must be inlined or results may be bad or performance poor.

So what did you think of instead making it a specification issue to allow simple functions like this to have `Inline = true` by default?  For example, we could state that all functions written in binding equation form would have `Inline = true` by default.

sjoelund:
> So what did you think of instead making it a specification issue to allow simple functions like this to have Inline = true by default? For example, we could state that all functions written in binding equation form would have Inline = true by default.

You'd need to make it with a binding equation and each input is used no more than once to make it sane to Inline in some tools. By making it explicit, Inline should be performed. The implicit one that it has a binding equation is just a hint that it might be good to inline this call.

henrikt-ma:
> You'd need to make it with a binding equation and each input is used no more than once to make it sane to Inline in some tools. By making it explicit, Inline should be performed. The implicit one that it has a binding equation is just a hint that it might be good to inline this call.

What I meant was to really really make `Inline = true` for functions expressed in some easily recognized ways.  Then, it would be just as wrong to not inline as when `Inline = true` is set explicitly.  I don't expect everybody to like this idea, I'm only trying to bring clarity to what people think of this approach.  I prefer this approach, since I think that there are many functions expressed in simple ways where the author intuitively assumes that it will be inlined, and therefore doesn't bother to add the `Inline` annotation, causing unnecessary room for unexpected differences across tools.

henrikt-ma:
> > You'd need to make it with a binding equation and each input is used no more than once to make it sane to Inline in some tools. By making it explicit, Inline should be performed. The implicit one that it has a binding equation is just a hint that it might be good to inline this call.
> 
> What I meant was to really really make `Inline = true` for functions expressed in some easily recognized ways. Then, it would be just as wrong to not inline as when `Inline = true` is set explicitly. I don't expect everybody to like this idea, I'm only trying to bring clarity to what people think of this approach. I prefer this approach, since I think that there are many functions expressed in simple ways where the author intuitively assumes that it will be inlined, and therefore doesn't bother to add the `Inline` annotation, causing unnecessary room for unexpected differences across tools.

The functions in this PR would all be possible to express in an extremely simple form:
```
  input Modelica.SIunits.Angle u;
  output Real y = sin(u);
```

These are some of the requirements they would fulfill:
- In binding equation form; no algorithm section.
- Every input only used once.
- Only a single output.

beutlich:
> I think I would prefer `Inline=true` explicitly added

Can do. I have no strong objection.

henrikt-ma:
Several more complicated functions in `Modelica.Math.Distributions` already have `Inline = true`, so the approach of not doing it here and expecting the specification to make it the default doesn't seem like a good option.  It would be better to set `Inline = true` now, then think about making it the default in the specification, and if it does, then adapt the MSL accordingly.

If the specification would end up only making `Inline = true` for functions in "binding equation form", it won't be a big deal to rewrite the functions in this form, so the current form with an algorithm containing a single assignment statement seems totally fine for now.

sjoelund (approved):

henrikt-ma (approved):

HansOlsson (approved):
Looks good. I realized that we in some rare cases will have to slightly modify Dymola to support these as efficiently as the previous variant, but that shouldn't be a blocker - and likely has other benefits.

PR #3704, opened 2021-01-06
Changed dofpT issue #3695

### LINKED ISSUE #3695: Trajectory of 'k' in R134a_setState_pTX doesn't look suitable for comparison
Can someone please verify that `k` is a meaningful comparison signal in `R134a_setState_pTX`:
- https://github.com/modelica/MAP-LIB_ReferenceResults/blob/v4.0.0/ModelicaTest/Media/TestOnly/R134a_setState_pTX/comparisonSignals.txt

It doesn't look like the typical comparison signal:
k

wischhusen:
Changed code for function dofpT for testing to resolve #3695.

### DIFF
--- a/Modelica/Media/R134a.mo
+++ b/Modelica/Media/R134a.mo
@@ -2471,17 +2471,17 @@ This function integrates the derivative of temperature w.r.t. time in order to a
       output SI.Density d "Density";
 
     protected
-      constant Real p_breaks[:]=R134aData.pbreaks
-        "Grid points of reduced pressure";
-      constant Real dl_coef[:, 4]=R134aData.dlcoef
-        "Coefficients of cubic spline for rho_liq(p)";
-      constant Real dv_coef[:, 4]=R134aData.dvcoef
-        "Coefficients of cubic spline for rho_vap(p)";
+      constant Real T_breaks[:]=R134aData.Tbreaks
+        "Grid points of reduced temperature";
+      constant Real dlt_coef[:, 4]=R134aData.dltcoef
+        "Coefficients of cubic spline for rho_liq(T)";
+      constant Real dvt_coef[:, 4]=R134aData.dvtcoef
+        "Coefficients of cubic spline for rho_vap(T)";
 
       Boolean liquid "Is liquid";
       Boolean supercritical "Is supercritcal";
       Integer int "Interval number";
-      Real pred "Reduced pressure";
+      Real Tred "Reduced temperature";
       Real localx "Abscissa of local spline";
       Integer i "Newton iteration number";
       Real dp "Pressure difference";
@@ -2498,9 +2498,9 @@ This function integrates the derivative of temperature w.r.t. time in order to a
       i := 0;
       error := 0;
       found := false;
-      pred := p/R134aData.data.FPCRIT;
-      (int,error) := Common.FindInterval(pred, p_breaks);
-      localx := pred - p_breaks[int];
+      Tred := T/R134aData.data.FTCRIT;
+      (int,error) := Common.FindInterval(Tred, T_breaks);
+      localx := Tred - T_breaks[int];
       // set decent initial guesses for d and T
       supercritical := p > R134aData.data.FPCRIT;
       if supercritical then
@@ -2509,10 +2509,10 @@ This function integrates the derivative of temperature w.r.t. time in order to a
       else
         liquid := T <= Modelica.Media.R134a.R134a_ph.saturationTemperature(p);
         if liquid then
-          d := R134aData.data.FDCRIT*Common.CubicSplineEval(localx, dl_coef[int,
+          d := R134aData.data.FDCRIT*Common.CubicSplineEval(localx, dlt_coef[int,
             1:4])*1.02;
         else
-          d := R134aData.data.FDCRIT*Common.CubicSplineEval(localx, dv_coef[int,
+          d := R134aData.data.FDCRIT*Common.CubicSplineEval(localx, dvt_coef[int,
             1:4])*0.95;
         end if;
       end if;


thorade (approved):
As stated by Stefan, density guess values are now calculated from temperature instead of pressure, leading to more stable iterations.

henrikt-ma (commented):
Since I'm not a model domain expert, I can't say much until I see the updated reference result.

I can still mention, though, that the model seems to simulate fine, and all result trajectories look extremely nice – quite different in quality from the current reference results.

Regarding the general problem of not being able to say much until there is an updated reference result:
- The natural solution would be to have the updated reference in the very same PR.
- All tool vendors would be able to do CI testing of the MSL _master_.  This would mean early feedback to MAP-Lib, as well as giving tool vendors much more freedom to plan when to test and report issues regarding an upcoming MSL release. 
- This doesn't have to be prohibitively expensive storage wise: A suitably compressed result that still occupies more than 10–100 KB probably contains signals that are not suitable comparison variables anyway.  (At Wolfram we have very good experience from working with correctness checking against compressed results that we keep in the same repository as the Modelica library sources.)
- With the figure annotations expected in Modelica 3.5, having reference trajectories for selected variables stored with the model would take us very close to a situation where tools could actually be rendering figures in documentation instead of just showing screen shots:  Never more screen shots that could be out of date (only in case out-of-date reference result still allows correctness checks to pass), users can make a plot together with a simulated result to judge for themselves whether the match is close enough, no more figures produced by one tool looking out of place in another tool.

beutlich:
> Since I'm not a model domain expert, I can't say much until I see the updated reference result.

Do you can confirm that the issue is fixed in WSM?

> The natural solution would be to have the updated reference in the very same PR

Nope. We do not have continuous reference results, but only generate them with new tags of MSL. If there is a new tag (= release) it is no big deal to regenerate them.

henrikt-ma:
> > Since I'm not a model domain expert, I can't say much until I see the updated reference result.
> 
> Do you can confirm that the issue is fixed in WSM?

No.  Since I'm not a domain expert the only things I would be able to confirm would be either the removal of the problematic signal from the reference result (no need to involve System Modeler to verify this), or check the System Modeler simulation against an updated reference result.

> > The natural solution would be to have the updated reference in the very same PR
> 
> Nope. We do not have continuous reference results, but only generate them with new tags of MSL. If there is a new tag (= release or pre-release) it is no big deal to regenerate them.

Than that would mean that a new pre-release tag is needed in this case, so that tool vendors get access to the updated reference result.

beutlich:
> 
> 
> > > Since I'm not a model domain expert, I can't say much until I see the updated reference result.
> > 
> > 
> > Do you can confirm that the issue is fixed in WSM?
> 
> No. Since I'm not a domain expert the only things I would be able to confirm would be either the removal of the problematic signal from the reference result (no need to involve System Modeler to verify this), or check the System Modeler simulation against an updated reference result.

OK, got it.

> > > The natural solution would be to have the updated reference in the very same PR
> > 
> > 
> > Nope. We do not have continuous reference results, but only generate them with new tags of MSL. If there is a new tag (= release or pre-release) it is no big deal to regenerate them.
> 
> Than that would mean that a new pre-release tag is needed in this case, so that tool vendors get access to the updated reference result.

Which would imply that there is some master plan about upcoming maintenance and main releases.

henrikt-ma:
> Which would imply that there is some master plan about upcoming maintenance and main releases.

Right. I should probably have said that a new post-release tag is needed.

beutlich:
> 
> 
> > Which would imply that there is some master plan about upcoming maintenance and main releases.
> 
> Right. I should probably have said that a new post-release tag is needed.

Which simply is a pre-release of the next maintenance release.

henrikt-ma:
> > > Which would imply that there is some master plan about upcoming maintenance and main releases.
> > 
> > 
> > Right. I should probably have said that a new post-release tag is needed.
> 
> Which simply is a pre-release of the next maintenance release.

In that case, this PR should target a maintenance branch rather than _master_.  For something targeting _master_, I suppose it could be an extremely early pre-release of a new version whose version hasn't even been decided yet.

I really can't see how to do this in a sensible way without model updates going hand in hand with updated reference results.

beutlich:
> 
> 
> > > > Which would imply that there is some master plan about upcoming maintenance and main releases.
> > > 
> > > 
> > > Right. I should probably have said that a new post-release tag is needed.
> > 
> > 
> > Which simply is a pre-release of the next maintenance release.
> 
> In that case, this PR should target a maintenance branch rather than _master_. For something targeting _master_, I suppose it could be an extremely early pre-release of a new version whose version hasn't even been decided yet.

Nope, all fixes go to master first. If it is labeled as bug it will be back-ported to all supported maintenance branches. That is the workflow we used to follow.

> I really can't see how to do this in a sensible way without model updates going hand in hand with updated reference results.

This would only work if reference results are stored within the same repository as the library and are automatically updated. IIRC the Buildings library is following that workflow. For what we have now, given the resources we need to face, we can only provide updated reference results with new tags.

henrikt-ma:
> > I really can't see how to do this in a sensible way without model updates going hand in hand with updated reference results.
> 
> This would only work if reference results are stored within the same repository as the library and are automatically updated. IIRC the Buildings library is following that workflow. For what we have now, given the resources we need to face, we can only provide updated reference results with new tags.

(Way to go, Buildings, that makes it two of us!)

Of course, scarce resources must be taken into consideration, and I can only imagine this working if the workload was distributed upon all contributors by making it a required part of any pull request with impact on simulation results.  A key resource here – in addition to the library contributions themselves – would be continuous integration testing, which I think should be a high priority for quality assurance, don't you agree, @casella?  Ideally, this should even prevent PRs from being merged if they break correctness compared to the reference results.

I understand that there are other challenges here as well, like how to deal with tool-specific issues causing CI test failures, and how to make it sufficiently easy for library developers to update the reference results.  These are challenges I believe are necessary to address to reach the next level of quality assurance.

MartinOtter:
Sorry. I have not enough knowledge about this topic to judge. Therefore, removed me as reviewer.

hubertus65 (approved):

PR #3701, opened 2021-01-01
Docker pull at Travis CI

beutlich:
refs #3699: Try workaround from https://travis-ci.community/t/docker-hub-you-have-reached-your-pull-rate-limit/10517/6

PR #3682, opened 2020-12-08
Update link in documentation to no longer rely on Freescience.info

beutlich:
Freescience.info does not seem to be reliable when it comes to availability, see [here](https://travis-ci.org/github/modelica/ModelicaStandardLibrary/builds/747784592#L343).

PR #3709, opened 2021-01-11
Utilize GitHub Actions for CI

beutlich:
This replaces Travis CI by GitHub Actions and has the advantage that the CI jobs run instantaneously and are visible as Checks directly in the PR.

This supersedes also #3701. Note, that no CI checks are removed, instead the OM checks are only run by Jenkins and no longer by Travis CI or GitHub Actions.

The one failing link check already is addressed by #3682. The link checks are run as weekly cron job and not with the CI for PRs.

I also set the Jenkins and GitHub Actions checks as required to pass.


dietmarw (approved):
Nice work.

sjoelund (approved):
Seems OK. Although I'd rather have it required to also pass compiling the C-sources.

beutlich:
> 
> 
> Seems OK. Although I'd rather have it required to also pass compiling the C-sources.

jenkins/pr-merge is required. I will also see if I cann add it to GH Actions, too.

sjoelund:
> jenkins/pr-merge is required. I will also see if I cann add it to GH Actions, too.

Oh, I must have been tired. Your first entry said it was so :D

PR #3712, opened 2021-01-12
Initialize motion in x direction (r_rel_a_1) instead of in z (r_rel_a_3)

### LINKED ISSUE #3706: Wrongly fixed initial conditions in example Modelica.Mechanics.MultiBody.Examples.Constraints.PrismaticConstraint
This ticket concerns the MSL sample model `Modelica.Mechanics.MultiBody.Examples.Constraints.PrismaticConstraint` (master branch as of 2021-01-11, Tag [4.0.0](https://github.com/modelica/ModelicaStandardLibrary/blob/6626538a236a34a390655b55d5e8c906bfe31772/Modelica/Mechanics/MultiBody/Examples/Constraints/PrismaticConstraint.mo#L88).)

The fixations in `freeMotionScalarInit` are wrong. The initial values `r_rel_a_1.start` and `v_rel_a_1.start` must be fixed instead of `r_rel_a_3.start` and `v_rel_a_3.start`, respectively.

With the wrong fixation of `r_rel_a_3.start` the initial equation system is singular since it imposes an initial local coordinate in the direction of the constraint.
But this local coordinate is alredy determined by the constraint equation `r_rel_a[3]=0;` in element `constraint`.

tobolar:
close #3706

### DIFF
--- a/Modelica/Mechanics/MultiBody/Examples/Constraints/PrismaticConstraint.mo
+++ b/Modelica/Mechanics/MultiBody/Examples/Constraints/PrismaticConstraint.mo
@@ -67,7 +67,7 @@ model PrismaticConstraint
     c=springOfJoint.c,
     s_unstretched=springOfJoint.s_unstretched,
     numberOfWindings=springOfJoint.numberOfWindings)
-                        annotation (Placement(transformation(
+    annotation (Placement(transformation(
         origin={-50,20},
         extent={{-10,-10},{10,10}})));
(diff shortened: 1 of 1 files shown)


maltelenz (approved):
In System Modeler, the model passes correctness against the reference result from 4.0.0 (which it also did before these changes). I did not look deeper to try to understand the model and the changes.

HansOlsson (approved):
Looks good to me, but I don't understand why we don't detect the error yet.
However, we detect another issue - will open issue.

PR #3736, opened 2021-01-27
More clearly specify damper cage in synchronous machine models

christiankral:
In the Electrical.Machines package the rotor leakage inductances and rotor resistances clearly refer to the damper cage. The damper cage shall also be clearly indicated in the cage parameters of the fundamental wave synchronous machine models.


dietmarw (approved):
Valuable clarification.

AHaumer (approved):
Good clarification.
Shouldn't we check Electrical.Machines and Magentic.QuasiStatic.FundamentalWave, too?

Issue #3737, opened 2021-01-28
PartialElementaryJoint should be extended from PartialTwoFrames


hyumo:
`Modelica.Mechanics.MultiBody.Interfaces.PartialElementaryJoint` is inherently a child class of `Modelica.Mechanics.MultiBody.Interfaces.PartialTwoFrames` except for one extra line `Connections.branch(frame_a.R, frame_b.R);`

This would help make replaceable joints easier.

tobolar:
This is a known issue in Mechanics, see e.g. https://github.com/modelica/ModelicaStandardLibrary/issues/2171#issuecomment-280661422 ("Overall the inheritance seems odd in the MultiBody-library, ...")
I prefer to solve this at once instead of having single problem-motivated solutions.

tobolar:
See #3739

PR #3738, opened 2021-01-28
Make PartialElementaryJoint extend from PartialTwoFrames 

hyumo:
see #3737


tobolar:
This is just a partial solution of larger inconsistency in MultiBody. I suggest to close without merging but considering it for the final solution.

hyumo:
ok, thx. I'll close it.

PR #3743, opened 2021-02-06
Resort variable declarations of Modelica.Fluid.Dissipation.HeatTransfer.HeatExchanger.{kc_flatTube, kc_flatTube_KC, kc_roundTube_KC} such that they are declared before used

beutlich:
I resorted the variable declarations in Modelica.Fluid.Dissipation.HeatTransfer.HeatExchanger.{kc_flatTube, kc_flatTube_KC, kc_roundTube_KC} such that they are declared before used.

As reported by OM#6366. Not sure if this is actually a tool issue since that ticket was wontfixed w/o further analysis.


HansOlsson (approved):
I can't see that it's needed, but should work as well as current variant.

MartinOtter (approved):

Issue #1164, opened 2017-01-14
assert(true, "message") will never be evaluated


modelica-trac-importer:
**Reported by thorade on 10 Jun 2013 11:30 UTC**
Section 8.3.7 of the Modelica Language Specification describes the assert statement:
 `assert(condition, message, level = AssertionLevel.error);`
 If the condition is `true`, message is **not** evaluated.
 If the condition is `false` different actions are taken depending on the level input.

Searching the MSL for `assert(true` yields 5 results, 4 of them in `Modelica.Fluid`. These will never be evaluated. Some are in the `else` branch of an ` if...elseif...else` construct, so if I understood the authors intention correctly, they should be evaluated.

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/1164

modelica-trac-importer:
**Comment by rfranke on 21 Jun 2013 20:29 UTC**
SVN 7c6963a2c40ba9b3778ba313961e59e9bee69300 fixes 4 occurences in Modelica.Fluid.Pipes and 2 occurences in Modelica.Fluid.Dissipation. The changes in Fluid do not effect simulation results, because the assertions are raised if an unknown modelStructure has been selected. If such a selection is possible at all, then equations are missing in the respective else branch and the translation should fail anyway.

Nevertheless, thank you for the bug report!

modelica-trac-importer:
**Comment by dietmarw on 21 Jun 2013 22:40 UTC**
The remaining one is found in:
```
Modelica/Media/Water/package.mo
193:            assert(true, "No events for pT-model");
```

modelica-trac-importer:
**Modified by otter on 27 Jun 2013 11:39 UTC**

modelica-trac-importer:
**Comment by hubertus on 27 Jun 2013 16:52 UTC**
Fixed in 31c12768cec12bca631d2ebd52a50fb3ce2c10f9

PR #3744, opened 2021-02-06
Add assertion in else branches of Modelica.Fluid.Dissipation.HeatTransfer.HeatExchanger.{kc_flatTube, kc_flatTube_KC, kc_roundTube, kc_roundTube_KC}

### LINKED ISSUE #1429: Conditionally uninitialized output variable Re in Modelica.Fluid.Dissipation.HeatTransfer.HeatExchanger.kc_flatTube used
**Reported by jfrenkel on 17 Feb 2014 14:34 UTC**
In the function Modelica.Fluid.Dissipation.HeatTransfer.HeatExchanger.kc_flatTube the output Re is only specified if IN_con.geometry == TYP.LouverFin or IN_con.geometry == TYP.RectangularFin is. True the enumeration does not yet comprise other possibilities. But if someone include on more possibility this is a good point for an undiscovered error. Initialise Re with a default value would prevent this easily. 

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/1429

beutlich:
Fix #1429 by adding assertions to else branches of Modelica.Fluid.Dissipation.HeatTransfer.HeatExchanger.{kc_flatTube, kc_flatTube_KC, kc_roundTube, kc_roundTube_KC}.

### DIFF
--- a/Modelica/Fluid/Dissipation.mo
+++ b/Modelica/Fluid/Dissipation.mo
@@ -1046,6 +1046,9 @@ This record is used as <strong>input record</strong> for the heat transfer funct
         elseif IN_con.geometry == TYP.RectangularFin then
           Re := abs(IN_var.m_flow)*IN_con.D_h/(IN_var.eta*A_c);
           Nu := max(1e-3, kc*IN_con.D_h/IN_var.lambda);
+        else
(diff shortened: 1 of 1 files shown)


HansOlsson (approved):
Looks good.

thorade (approved):
In the last branch it would also suffice to just use 
`assert(false, "this should not happen");`
See #1164

PR #3755, opened 2021-02-18
Move the heatport on the diagram to match the icon

dietmarw:
Previously the heatport was covering the electrical connector


AHaumer:
@dietmarw could we include #3748 in this PR? Otherwise two PRs depend on each other.
I'd committ to your PR a backwards compatible solution for #3748 with an optional internal ground and optional ground connector.

christiankral:
So this PR is going to be replaced by #3756

dietmarw:
OK that's fine with me.

PR #3784, opened 2021-03-31
Fix LCC build

beutlich:
Testing with LCC 3.8 on Win revealed some issues.


HansOlsson (approved):
I trust that it is correct, although I don't have LCC.

beutlich:
@sjoelund OK for you, too?

sjoelund (approved):

PR #3805, opened 2021-05-02
Update CI to check for deprecated Text.lineColor annotation

beutlich:
It turned out, that the deprecated Text.lineColor annotation gets introduced by accident. This is the CI check to prevent it for future contributions.


HansOlsson (approved):
Seems ok.
As far as I can tell it shouldn't happen with Dymola 2021X and later.

dietmarw (approved):

PR #3817, opened 2021-05-24
Implemented split of stray inductance in common and individual part

AHaumer:
Up to now, the stray field of (symmetric) poylphase windings is modeled as a common inductance or permeance. During some master thesis, I came upon problems with field orineted control of polyphase machines (m>3) fed by swicthing power electronics. The solution was to split the stray inductance in a common part and an individual part (individually for everey phase) to limit current rise in certain cases. Since I assume that other users might encounter similar problems, I've cleaned up my solution to create that PR. 
The ratio of common stray inductance with respect to total stray inductance is defined by a parameter `ratioCommonLeakage` with a default of 1, which is perrfectly backwards compatible. Note that for machines with three phases (Electrical.Machines) and for quasistatic machines (Magnetic.QuasiStatic.FundamentalWave) this parameter has no influence, and for performance reasons implementation in these two libaries is skipped. The parameter is essential only for Magnetic.FundamentalWave. 
Note that the situation is described in detail in Magnetic.FundamentalWave.UsersGuide.WindingModel.
The example FundamentalWave.Examples.BasicMachines.InductionMachines.ComparisonPolyphase.IMC_DOL_CommonLeakage compares three identical 5-phase machines but `ratioCommonLeakage`={0, 0.5, 1},


AHaumer:
**This is still a draft!**
I've corrected the implementation of individual tray permeances.
@christiankral : Shouldn't the comparison of two identical machine, one with stray permeance as common leakage and the other as individual leakages, give the same result? I'm afraid the parameter calculation is not correct ...

Modelica/Magnetic/FundamentalWave/UsersGuide/WindingModel.mo
```diff
@@ -57,5 +57,12 @@ Therefore this splitted stray inductance is only modeled in the <a href=\"Modeli
 The parameter <code>ratioCommonLeakage</code> has a default of 1, which is perfectly backwards compatible.
 </p>
 
+<h4>Note:</h4>
+<p>
+For a model with <code>ratioCommonLeakage &lt; 1 - eps</code> (including indvidual stray inductances)
```
beutlich, line 62:
```suggestion
For a model with <code>ratioCommonLeakage &lt; 1 - eps</code> (including individual stray inductances) 
```

Modelica/Magnetic/FundamentalWave/UsersGuide/WindingModel.mo
```diff
@@ -57,5 +57,12 @@ Therefore this splitted stray inductance is only modeled in the <a href=\"Modeli
 The parameter <code>ratioCommonLeakage</code> has a default of 1, which is perfectly backwards compatible.
 </p>
 
+<h4>Note:</h4>
+<p>
+For a model with <code>ratioCommonLeakage &lt; 1 - eps</code> (including indvidual stray inductances) 
+typically more states have to be inialized than
+for a model with <code>ratioCommonLeakage = 1</code> (without indvidual stray inductances).
```
beutlich, line 64:
```suggestion
for a model with <code>ratioCommonLeakage = 1</code> (without individual stray inductances).
```

AHaumer:
Now I've moved the individual leakage from a magnetic implementation to an electrical inductor, and updated the documentation. 
The test case evaluates correct.

AHaumer:
We even have the choice between electrical implementation of the individual leakage as electrical polyphase inductor or magentic permeances (the second one see SymmetricPolyphaseWinding_M_).
@christiankral and @dietmarw which implementation would you prefer?
Both are working, both are clear, both are backwards compatible.

AHaumer:
Ok after discussions with @christiankral we decided for a magnetic implementation of the indivudual leakage.
From my side this PR is ready for review and merging.

dietmarw:
Just to clarify, since this is an extension and not a bug-fix it should go into the next minor version which allows extensions.

AHaumer:
> 
> 
> Just to clarify, since this is an extension and not a bug-fix it should go into the next minor version which allows extensions.

Somehow it's a bug fix. The library promises polyphase machine models. If you design FOC for polyphase drives, the models won't work with switching power electronics (only with averaging PE models). Shall I add the "bug" label?

dietmarw:
It would not work since you are adding several new components and not just fixing the existing things. I think it is better and less work to simply include it in the next minor version. Nothing says that the bug fix version is going to happen earlier than the new minor version. In the past it used to be at the same time anyway.

AHaumer:
> 
> 
> It would not work since you are adding several new components and not just fixing the existing things. I think it is better and less work to simply include it in the next minor version. Nothing says that the bug fix version is going to happen earlier than the new minor version. In the past it used to be at the same time anyway.

I'm modifyng one component, and adding one new example that proofs that the modification works properly.

christiankral:
- This PR is not 100% backwards compatible, as -- for example -- the plugs `electroMagneticConverter.plug_p` and `electroMagneticConverter.plug_n` are not available any more in `FundamentalWave.BasicMachines.Components.SymmetricMultiPhaseWinding`
- So if we indicate this PR as a bug fix, the proposed non-backwards compatible change may be accepted
- If we name it a feature there are two fully backwards compatible options from my point of view:
    1. Add the individiual leakage permeance as a conditional component to `FundamentalWave.Components.MultiPhaseElectroMagneticConverter`; this component then already includes the feature of the individual leakages
    2. Create a duplicate of `FundamentalWave.Components.MultiPhaseElectroMagneticConverter`, name it `FundamentalWave.Components.MultiPhaseElectroMagneticConverterWithLeakage` (similar to `FluxTubes.Basic.ElectroMagneticConverterWithLeakageInductance`), and add the (conditional) individual; permeances there; then model `FundamentalWave.BasicMachines.Components.SymmetricMultiPhaseWinding` has to rely on `MultiPhaseElectroMagneticConverterWithLeakage` instead of `MultiPhaseElectroMagneticConverter`

AHaumer:
OK whether its a bug fix or an enhancement, backwards compatibility is important.
@christiankral I've choosen your option 1 (already committed);
>       
>       1. Add the individiual leakage permeance as a conditional component to `FundamentalWave.Components.MultiPhaseElectroMagneticConverter`; this component then already includes the feature of the individual leakages

dietmarw (approved):
I added some minor cosmetic and typo fixes. With the backward compatible solution it now looks fine to merge.

christiankral (approved):
Looks good.

AHaumer:
What's the problem with the license/cla - seems that the check got stuck ,,,

dietmarw:
I'll rebase the branch and that should trigger the check once more.

PR #3818, opened 2021-05-26
Fix Inconsistent M_OLine as reported in #3748

dietmarw:
This applies the bugfix from #3815 and #3827 to maint/4.0.x


Modelica/Electrical/Analog/Lines/M_OLine.mo
```diff
@@ -26,6 +26,8 @@ public
     each final min=Modelica.Constants.small,
     each unit="F/m") = {2.38e-11,1.01e-10,8.56e-11,5.09e-12,2.71e-11,2.09e-11,
     7.16e-11,1.83e-11,1.23e-10,2.07e-11} "Capacitance per meter";
+  parameter Boolean useInternalGround=true "= true if internal ground is used, otherwise use reference pin"
```
beutlich, line 29: We have a dilemma here, since our guideline states:

> Introducing a new name in the public section of a class (model, package, ...) or in any section of a partial class is **not** allowed

dietmarw: I don't think it is a big issue here. And as you say, it is only a recommendation. The LO's should decide. In the worst case it will be not part of 4.0.1 but 4.1.0 (whenever that is ;-) )

dietmarw: Actually the project leader needs to decide if this can simply get merged, or need a mention in the release notes in addition or simply should not get merged.

AHaumer (approved):
In my humble opinion it's no big deal introducing a new parameter in a bugfix as long as the default of the parameter leads to the same behaviour as before.
For me it's ok.

AHaumer:
Should be based on #3827

Modelica/Electrical/Analog/Lines/M_OLine.mo
AHaumer: > Actually the project leader needs to decide if this can simply get merged, or need a mention in the release notes in addition or simply should not get merged.

I doubt that @MartinOtter will comment on that. So if we (@dietmarw, @christiankral and @AHaumer) agree on the solution, it should be ok.

dietmarw: That might be the pragmatic solution but not the correct one especially since this is regarding a maintenance release. It is unfortunate that we don't have an active project leader but that is not reason enough to bypass this kind of approval.

dietmarw:
@christiankral  Thanks for the improvements. The problem is that since this is only meant as bugfix it should be ideally a copy of the bugfix commit applied to master. So basically whatever is in this PR should already have been applied to master. That way we make sure that a bugfix release has not suddenly "features" that the next minor release is missing again.
So I'm going to remove your visual improvements again. These can be put into a separate enhancement PR (which will **not** get backported to 4.0.1).

dietmarw:
So now the two bugfixes from #3815 and #3827 are included in this backport. Please **only** check for validity and do not apply any enhancement commits. If you find something you like to improve, put them into a new PR targeted at master and the next MINOR version. 
@christiankral The changes you have done and which I have removed can be found in the new draft PR #3833

christiankral (approved):
I suppose everything is good now.

AHaumer (approved):
Fine with me

AHaumer:
Can please anybody tell me why merging is blocked?

sjoelund:
> Can please anybody tell me why merging is blocked?

GitHub tells me this can be merged, squashed, or rebased. Nothing blocking

beutlich:
> 
> 
> Can please anybody tell me why merging is blocked?

As configured, only repo admins can merge to the maintenance branches main/4.0.x and maint/3.2.3. In the case of this PR, @MartinOtter is assigned and as such should take the responsibility for the merge.

PR #3819, opened 2021-05-26
Lines examples

max-privato:
Let's try this way.
This PR is commented in https://github.com/modelica/ModelicaStandardLibrary/pull/3815#issuecomment-848663677, 
comment by myself on 26/05/21, around 12.30 A.M.


Modelica/Electrical/Analog/Examples/Lines.mo
```diff
@@ -0,0 +1,376 @@
...
+      smooth=Smooth.None));
+  annotation (
+    Diagram(coordinateSystem(preserveAspectRatio=false, extent={{-100,-80},{
+            100,60}}), graphics),
+    Icon(coordinateSystem(extent={{-100,-80},{100,60}})),
+    experiment(
+      StopTime=0.003,
+      Interval=1e-06,
+      Tolerance=1e-06,
+    uses(Modelica(version="4.0.0"))),
```
beutlich, line 159: Uses-annotation should be removed.

dietmarw:
Some comments after checking the example briefly:
 - In the MSL we try to apply the policy of one file per model. So new examples should follow that.
 - I would probably not create a package for the two examples but simply add them directly but use a meaningful name. E.g., `Lines_SmoothStep` and `Lines_CompareTrunks`.
 - variables and parameters that are referred to inside the documentation should be enclosed in `<code>...</code>` 
 - we also need a `comparisonSignals.txt` under `Resources/Reference/Modelica/Electrical/Analog/Examples/...`. This simply contains a selection of signals that should be used for future regression tests.

Modelica/Electrical/Analog/Examples/Lines.mo
```diff
@@ -0,0 +1,376 @@
...
+  connect(ground4.p, tline3.n1) annotation (
+      Line(points = {{20, -46}, {20, -44}}, color = {0, 0, 255}));
+  connect(tline4.n1, tline3.n2) annotation (
+      Line(points = {{50, -44}, {40, -44}}, color = {0, 0, 255}));
+  connect(tline4.p1, tline3.p2) annotation (
+      Line(points = {{50, -24}, {40, -24}}, color = {0, 0, 255}));
+  connect(ground5.p, tline4.n1) annotation (
+      Line(points = {{50, -46}, {50, -44}}, color = {0, 0, 255}));
+    annotation (
+      uses(Modelica(version="4.0.0")),
```
beutlich, line 343: Uses-annotation should be removed.

Modelica/Electrical/Analog/Examples/Lines.mo
```diff
@@ -0,0 +1,376 @@
...
+        extent={{-100,-100},{100,100}},
+        preserveAspectRatio=true,
+        initialScale=0.1,
+        grid={1,1})),
+    Diagram(coordinateSystem(
+        extent={{-100,-100},{100,100}},
+        preserveAspectRatio=true,
+        initialScale=0.1,
+        grid={2,2})),
+    uses(Modelica(version="4.0.0")));
```
beutlich, line 375: Uses-annotation should be removed.

beutlich (dismissed):
Make sure to add the corresponding comparisonSignals.txt files for each example.

AHaumer:
@max-privato Nice ecamples! I hope it's ok that I've cleaned up them.
@dietmarw The examples reveal another problem: It's impossible to initialize the OLine model!
The componentes C[N] and L[N+1] are protected - so no chance to set initial values or use initial equations.
This requires changes in the OLine model - I'll take care of that a bit later.

AHaumer:
I've made the respective components public, and added initialization to the examples.
@max-privato please check whether I've picked appropriate comparisonSignals for regression tests in 
Modelica\Resources\Reference\Modelica\Electrical\Analog\Examples\Lines\{SmoothStep, CmpLineTrunks}\comparisonSignals.txt

max-privato:
> Some comments after checking the example briefly:
> 
> * In the MSL we try to apply the policy of one file per model. So new examples should follow that.
> * I would probably not create a package for the two examples but simply add them directly but use a meaningful name. E.g., `Lines_SmoothStep` and `Lines_CompareTrunks`.

I understand. However, let me try o explain the reason beyond my choice of a new folder:
1. it seems to me that the list of examples in Electric analog is getting too long. 
2.  OpAmps has already a folder of its own. 
3.  the plan was to add a third Lines example related to M_Oline. 

Therefore, if I were in your shoes I would consider the possibility to keep examples grouped by topic, at least in the case of Lines.
Please let me know if you stand by the position of leaving the Lines models in the generic Examples folder (no subfolder) or not after the comments above.

> * variables and parameters that are referred to inside the documentation should be enclosed in `<code>...</code>`

Ok I'll change this
> * we also need a `comparisonSignals.txt` under `Resources/Reference/Modelica/Electrical/Analog/Examples/...`. This simply contains a selection of signals that should be used for future regression tests.

Ok. I'll do this.

AHaumer:
@max-privato do you have a hint about the meaning of the protected variables in TLine? They have to get a description string. For initialization, we have to make them public.

max-privato:
> I've made the respective components public, and added initialization to the examples.
> @max-privato please check whether I've picked appropriate comparisonSignals for regression tests in
> Modelica\Resources\Reference\Modelica\Electrical\Analog\Examples\Lines{SmoothStep, CmpLineTrunks}\comparisonSignals.txt

I checked, they are ok for me.

max-privato:
> @max-privato do you have a hint about the meaning of the protected variables in TLine? They have to get a description string. For initialization, we have to make them public.

There must be something related with incident and reflected waves. I'll check on my electromagnetic book today.

AHaumer:
I have unified the TLine models, and marked the old ones {TLine1, TLine2, Tline3} as obsolete. The parameter defaults of the new model are set such way that exchanging one of the old models {TLine1, TLine2, Tline3} to the new TLine we get the same results.

AHaumer:
@max-privato thanks a lot for checking the meaning of TLine.es and TLine.er - just add a description string in TLine.mo.
If you manage to implement an example for M_OLine and for ULine, we have examples for all line models - perfect!

AHaumer:
I have access to Branin1967, and I've added description strings. Citing Branin:
"We regard [es] as a voltage source corresponding to the incident voltage wave arriving now (time=t) at  the output terminals (x=d) after having left the input terminals (x=0) at time t-Td.  Similarly, we regard [er] as a voltage source corresponding to the reflected voltage wave now arriving at the input terminals after having left the output Td seconds ago."
Additionally, he takes an inner impedance Z0 into account. 
Anyway, intializing a de-energized line in the example seems to be ok.

max-privato:
> I have access to Branin1967, and I've added description strings. Citing Branin:
lucky you!

I must add a few words to what I said this morning regarding initialization of TLine (I cannot find where within gitHub).
Initialization is theoretically impossible, and in practice very difficult with TLine, because it implements a behavior mathematically described by partial-derivative equations. These have infinite states, which correspond to the (theoretically) infinite delay() input past values, within the delay interval.

max-privato:
> I have access to Branin1967, and I've added description strings. Citing Branin:

Can you share it with me? I'm very curious about it. 
I should be entitled to receive a copy of it on my University e-mail, since it is has a subscription to IEEE Explorer.

sjoelund:
> > I have access to Branin1967, and I've added description strings. Citing Branin:
> 
> Can you share it with me? I'm very curious about it.
> I should be entitled to receive a copy of it on my University e-mail, since it is has a subscription to IEEE Explorer.

I believe it is https://ieeexplore.ieee.org/document/1447963 (I can access it with my university subscription)

AHaumer:
@sjoelund you're right. I'm just a little bit hesitant regarding copyright infringement ...
If @max-privato sends me an email: 
https://www.oth-regensburg.de/fakultaeten/elektro-und-informationstechnik/personen-kontakt/contact/anton-haa-haumer.html
I'll be happy to answer in a private discussion.

sjoelund:
Well, I think the DOI link could be added somewhere so you could find the source from the documentation @max-privato should be able to access it from work at any rate

AHaumer:
> I must add a few words to what I said this morning regarding initialization of TLine (I cannot find where within gitHub).
> Initialization is theoretically impossible, and in practice very difficult with TLine, because it implements a behavior mathematically described by partial-derivative equations. These have infinite states, which correspond to the (theoretically) infinite delay() input past values, within the delay interval.

Well the model applies voltage sources at output and input (which are a combination of forward travelling and reflected wave) and inner resistance Z0. The forward travelling wave arrives delayed at the output, and the reflected wave arrives delayed at the input again. So in Modelica, only start values for the iteration of the output voltage v2 and the voltage source er of the reflected wave are required.

max-privato:
> I believe it is https://ieeexplore.ieee.org/document/1447963 = https://doi.org/10.1109/PROC.1967.6033 (I can access it with my university subscription)
Thank you. I could get the document.

max-privato:
Yesterday I uploaded the CLA.
I would like the following sentence to be added to the info of the two examples:
"Example originally developed by Massimo Ceraolo from the University of Pisa, on May 2021"
I think this is compliant with the CLA I've signed and similar to what I can read on other examples. Unfortunately, I cannot do this change here at this stage. 
Can someone do this for me?

beutlich:
You can fill the revisions annotation for the history. See for example: https://github.com/modelica/ModelicaStandardLibrary/blob/0dc38e43f673e29962593057bfb60c03a43ccd5a/Modelica/Electrical/Analog/Examples/CauerLowPassAnalog.mo#L116-L125

max-privato:
> If you manage to implement an example for M_OLine and for ULine, we have examples for all line models - perfect!

I'm working on M_OLine. It was more complicated than expected, but the results are satisfying. You'll see.
I will include functions to generate line parameters from geometry, to make M_Oline more user-friendly.

dietmarw:
Just be careful to not create any non-backward compatible changes to the `M_OLine` and `OLine` models whilst working on this.

max-privato:
> Just be careful to not create any non-backward compatible changes to the `M_OLine` and `OLine` models whilst working on this.

Sure!

AHaumer:
Added an example of a lightning hitting a transmission line.
The block Utilities.Lightning prescribes the current of the lightning. If we keep it in utilities, we can skip #3821. 
I'll try to look up the parameters for a threephase model of the line, using M:OLine taking coupling between the phases into account.

christiankral (dismissed):
The examples look good. Some comments, however:

- I prefer the solution of having a lightning model in blocks as proposed in #3821: so I did not comment on model `Modelica/Electrical/Analog/Examples/Utilities/Lightning.mo` and wait for a decision
- The term *PIs* may be clear if one is used it; however I propose to explain what is meant (e.g. in `Modelica.Electrical.Analog.Examples.Lines`, so that each example using this term may refer to this definition as `<a href="modelica://Modelica.Electrical.Analog.Examples.Lines">PIs</a>`

AHaumer:
I had to correct the implementation of the block Lightening according to Rock2012 as doubel-exponential function.
I'd prefer to place this block in Modelica.Electrical.Analog.Examples.Utilities.

AHaumer:
What do you mean by _PIs_, where is this term used?
>     * The term _PIs_ may be clear if one is used it; however I propose to explain what is meant (e.g. in `Modelica.Electrical.Analog.Examples.Lines`, so that each example using this term may refer to this definition as `<a href="modelica://Modelica.Electrical.Analog.Examples.Lines">PIs</a>`

AHaumer:
> 
> What do you mean by _PIs_, where is this term used?
> 
I see ... in fact, the term "PI" is not correct, I replaced it by "segment". Looking into OLine, we use a series connection of T-shaped segments (although they are organized a little bit different in OLine): R/2 + L/2 + (C || G) + R/2 + L/2
(C || G) are connected to refPin or ground. 
R/2 + L/2 at the output are concatenated with R/2 + L/2 at the input of the next segment.

AHaumer:
It seems to me that @max-privato opened a "big barrel":

- For all the line models, we had no examples and no possibility for proper initialization.
Now we have examples and proper initialization for OLine and TLine (summarizing TLine1, TLine2, TLine3).
- An example for ULine ist still missing, but initialization is possible.
- For M_OLine, initialization is possible but not very clear; additionally, I would have an example (again the lightening) if we manage to find proper parameters for M_OLine (i.e. couplings between the phases in terms of M, C and G).

Modelica/Electrical/Analog/Examples/Lines/CompareLineTrunks.mo
```diff
@@ -0,0 +1,197 @@
...
+<p>This wave can be seen looking at intermediate points of the line, looking at both OLine (several segments per line trunk) and TLine models.</p>
+<p>OLine in the lossless case is less accurate, since shows oscillations that are not present in a distributed parameter line but, 
+since it allows the addition of resistances and conductances, allows evaluation of losses. </p>
+<p>Suggested tests:</p>
+<h4>7 segments per trunk, lossless</h4>
+<p>Run the model as it is. Since resistances and conductances are very small in OLine, the simulation is substantially lossless.</p>
+<p>Show in the same plot <code>ramp.y, oLine1.p2.v, oLine2.p2.v, oLine3.p2.v, oLine4.p2.v.</code> </p>
+<p>Show in another plot, simultanelusly, <code>ramp.y, tLine1.p2.v, tLine2.p2.v, tLine3.p2.v, tLine4.p2.v.</code> Here oLine shows unreal oscillations, which, however, reduce if the number of segments per trunc increases. </p>
+<h4>50 segments per trunk, lossless</h4>
+<p>Try increasing this number from 7 to 50 using parameter <span style=\"font-family: Courier New;\">segsPerTrunk</span>, and look at the same plots as per test 1</p>
```
beutlich, line 186: span style tags should not be used. Use code tag instead.

AHaumer:
@max-privato I suppose your comment at closed PR #3815 should go here:
I was trying to create a good example for M_OLine. I thought of a high voltage transmission line. I have a nice example showing the voltage a fence near the line reaches, and the current the body receives when touches the fence.

When doing this I suddenly realized that there is a big weakness of M_Oline model.
It is that it does not consider resistances on off-diagonal elements of the impedance matrix. When we model a line, which can have possibly ground current return,  off-diagonal elements have resistance, which is due to the ground. 
I think of modeling it according to with the Carson's theory, which is still widely used by Power System Engineers, even though it was formulated in 1926 and even though more complex theories exist.
But Carson formulas have already some complexity, and some simplification are usually proposed, to limit the infinite series of its expansion to the most significant terms.
I  already implemented Carson's formulas (stopping the series rather early so that to be able to use the obtained parameter values just for power frequency analysis) in Modelica when I wrote the paper "Modeling and Simulation of AC Railway Electric Supply Lines Including Ground Return" on IEEE Transactions on Transportation Electrification (check Appendix).

So, I could either:
1. propose for MSL an example in which the currents on the three wires have zero-sum (using the existing O_MLine, and writing caution note that the same model is not adequate when we have some ground current)
2. try to expand first, through an ad hoc PR, M_OLine so that it accepts non-zero resistances also on the off-diagonal matrix elements
3. wait for someone doing the change on O_MLine this and then concentrate on the example (I say this because in the past I proposed some ideas, and other implemented them faster than I could do).

What do library officials think?

I suppose that if point 2. above is chosen I have first to expect this ticket #3815 to be closed?
Anyway if 2 or 3 is chosen I should wait for this ticket, or the ticket related to the M_Oline enhancement with the inclusion of off-diagonal resistances be closed?

max-privato:
> @max-privato I suppose your comment at closed PR #3815 should go here:
Thanks.

I've had a look at the M_OLine code and it seems that implementing solutions 2 and 3 require a total restructuring of M_Oline, which is totally out of scope of just trying to offer an example of OM_Oline usage.
Maybe I'll go to on with option.

AHaumer:
> I was trying to create a good example for M_OLine. I thought of a high voltage transmission line. I have a nice example showing the voltage a fence near the line reaches, and the current the body receives when touches the fence.

We would appreciate that!
 
> When doing this I suddenly realized that there is a big weakness of M_Oline model.
> It is that it does not consider resistances on off-diagonal elements of the impedance matrix. When we model a line, which can have possibly ground current return, off-diagonal elements have resistance, which is due to the ground.
> I think of modeling it according to with the Carson's theory, which is still widely used by Power System Engineers, even though it was formulated in 1926 and even though more complex theories exist.
> But Carson formulas have already some complexity, and some simplification are usually proposed, to limit the infinite series of its expansion to the most significant terms.
> I already implemented Carson's formulas (stopping the series rather early so that to be able to use the obtained parameter values just for power frequency analysis) in Modelica when I wrote the paper "Modeling and Simulation of AC Railway Electric Supply Lines Including Ground Return" on IEEE Transactions on Transportation Electrification (check Appendix).
> 
> So, I could either:
> 
>     1. propose for MSL an example in which the currents on the three wires have zero-sum (using the existing O_MLine, and writing caution note that the same model is not adequate when we have some ground current)
> 
>     2. try to expand first, through an ad hoc PR, M_OLine so that it accepts non-zero resistances also on the off-diagonal matrix elements
> 
>     3. wait for someone doing the change on O_MLine this and then concentrate on the example (I say this because in the past I proposed some ideas, and other implemented them faster than I could do).
> 
> 
> What do library officials think?

I have to think about that (and look up Carsons' theory), I don't understand it fully at the first reading.
Maybe I'm wrong, but you could model 3 phases and as the fourth phase the neutral where the sum of phase currents is flowing back, independent of overall ground.
At first sight, I would tend to spend not too much time on M_OLine. This model would better fit into Electrical.Polyphase. 
Bear in mind; Initialization of all the inductances and capacitances is quite a bit of work (unless we initialize with start=0, fixed=true).

> I suppose that if point 2. above is chosen I have first to expect this ticket #3815 to be closed?
> Anyway if 2 or 3 is chosen I should wait for this ticket, or the ticket related to the M_Oline enhancement with the inclusion of off-diagonal resistances be closed?

#3815 is closed. I'd tend to not spend too much work on back-port changes to maintenance (#3818).

christiankral:
I agree with @AHaumer to not put too much effort into the "Analog" line models. So in case we "need" to have an `M_OLine` example, I'm in favor of option 1.

I prefer to develop polyphase transient and quasi stastic line models and examples in the long run. Then we shall deprecate the original models and examples of the Analog package, as the line models use a connector concept which is not used anywhere else in the MSL.

max-privato:
> I agree with @AHaumer to not put too much effort into the "Analog" line models. So in case we "need" to have an `M_OLine` example, I'm in favor of option 1.

Good. I'll do this.

> I prefer to develop polyphase transient and quasi stastic line models and examples in the long run. Then we shall deprecate the original models and examples of the Analog package, as the line models use a connector concept which is not used anywhere else in the MSL.

Probably having a power line model in MSL would help Modelica to compete with dedicated tools such as ATP, EPRI's EMTP, EMTP-DC, Matlab SimscapeElectrical, etc.

Once I'm done with the example, if I find time I'll work on a paper on modelling power lines taking into account ground return, such as the one Simscape electrical "Frequency-dependent Overhead Line (Three-phase)" or "Frequency-Dependent Transmission Line".

I'll prepare the paper using Modelica (I already have some building blocks). Once this is done I'll see if something general and easy-to-use enough can be proposed for MSL.

dietmarw:
@max-privato I've just force pushed to your PR to remove your latest merge from master commit. Please don't merge the master branch into the PR branches. If needed we can rebase a PR branch onto master before finally merging (as in accepting it).

max-privato:
> @max-privato I've just force pushed to your PR to remove your latest merge from master commit. Please don't merge the master branch into the PR branches. If needed we can rebase a PR branch onto master before finally merging (as in accepting it).

bleah.. I didn't even realised I did this!

max-privato:
Well,
I've prepared an M_OLine example.
It was much trickier and required much more time than expected. I've explained everything in the PR comment.
Maybe these difficulties will become opportunities at a later time.

**Is it better that I wait for this PR #3819 to be merged, before issuing a new PR, I suppose?** 

Indeed this new PR will involve also including two functions as separate files, which will be useful not only for the example but per se. If this is judged to be overkill, I can easily drop the PR.

AHaumer:
Just wondering waht's the status of this PR ...
@max-privato, you're preparing a new PR with the MO_Line example after this one is closed?
If this is correct (and this PR is ready for review) I'm going to review but I'd also ask @dietmarw and @christiankral for review.

max-privato:
> Just wondering what's the status of this PR ...
> @max-privato, you're preparing a new PR with the MO_Line example after this one is closed?
> If this is correct (and this PR is ready for review) I'm going to review but I'd also ask @dietmarw and @christiankral for review.

max-privato:
> > Just wondering what's the status of this PR ...
> > @max-privato, you're preparing a new PR with the MO_Line example after this one is closed?

Yes, this is the idea, I have that PR already, just waiting for this PR to be closed. 
I decided to separate that one from this to avoid cross effects.

> > If this is correct (and this PR is ready for review) I'm going to review but I'd also ask @dietmarw and @christiankral for review.

AHaumer:
I have packed too much in this PR (fix of `M_OLine` and initialization of lines) to ensure initialization of the examples - 
therefore I saved the examples, set this PR to draft and create a new PR based on master with these examples after #3827 and #3828 are merged. @max-privato thanks again for your examples, and I hope this is ok for you.

dietmarw:
@AHaumer Please show a bit of patience. There is probably no need for yet another PR on this. Let me first try to rebase it and should this fail **then** we could scrap it and create a new PR.

max-privato:
> I have packed too much in this PR (fix of `M_OLine` and initialization of lines) to ensure initialization of the examples -
> therefore I saved the examples, set this PR to draft and create a new PR based on master with these examples after #3827 and #3828 are merged. @max-privato thanks again for your examples, and I hope this is ok for you.


Yes, they are ok.
I saw the many developments that have undergone on the topic of M_OLine and examples. You did a lot of work!

But I did not stay inactive, however. I've spent much time on a new example for M_OLine (had to solve several nasty issues).
I made no changes on M_Oline, but had to add to the library a couple of functions having the purpose to create from line geometry the Y and Z matrices.
I'll wait for this PR to be finished, before creating a new one with this example.

AHaumer:
> But I did not stay inactive, however. I've spent much time on a new example for M_OLine (had to solve several nasty issues).
> I made no changes on M_Oline, but had to add to the library a couple of functions having the purpose to create from line geometry the Y and Z matrices.
> I'll wait for this PR to be finished, before creating a new one with this example.

Thanks that's fine @max-privato, but you'll have to rebase your branch on master as soon as the other PRs are merged.

AHaumer:
> 
> 
> @AHaumer Please show a bit of patience. There is probably no need for yet another PR on this. Let me first try to rebase it and should this fail **then** we could scrap it and create a new PR.

Well my problem is that some days (as today) I have time to take care of MSL, many other days I'm busy with teaching.
After merging #3827 and #3828, maybe this description makes it easier:

- The examples are collected in `Analog.Examples.Lines`
- They additionally utilize `Analog.Examples.Utilities.{LightningDoubleExp, LightningHeidler}`
- In the examples, the usage of `TLine1` should be changed to the unified `TLine`

AHaumer:
@dietmarw and @christiankral , what's your opinion:
Before we review and merge this PR, should we move `Utlities.{LightningDoubleExp, LightningHeidler}` to `Blocks.Sources` and implement a current source and a voltage source in `Analog`?
According to my investigations, the same impulse trajectories are used to describe currents (like lightning) and voltages. The blocks ae well documented and tested, and the're compatible with other signal sources.

AHaumer:
#3821 has to be merged first, then I should inspect and clean up this PR again.

christiankral:
> #3821 has to be merged first, then I should inspect and clean up this PR again.

#3821 is already closed and thus not considered any more. I suppose #3831 is the correct reference here.

dietmarw:
After the #3831 has been merged I have now rebased this PR onto master which now **only** adds the line examples. 
Things that I noticed:
 1. Some of the examples are missing the comparison files and should be added as well.
 2. Should the `DemonstrateLightning` example also be moved under the `Lines` package. This was the case in the original version of this PR but was implemented outside of `Lines` subpackage in #3831. One could argue that `DemonstrateLightning` is not necessarily a lines only example (doesn't even contain any of the lines models). 
 3. We can now start and look a bit closer into cleaning up the examples and their documentation.

beutlich:
>     1. Some of the examples are missing the comparison files and should be added as well.

Unfortunately we do not have a CI check that each example model also requires a corresponding comparisonSignals.txt file, since it is not trivial to extract all example model names. I already tried to motivate @eshmoylova and @svorkoetter to improve moparser to optionally return/print the names of the example models, but without progress.

AHaumer:
> 
> 
> After the #3831 has been merged I have now rebased this PR onto master which now **only** adds the line examples.
> Things that I noticed:
> 
>     1. Some of the examples are missing the comparison files and should be added as well.
> 
>     2. Should the `DemonstrateLightning` example also be moved under the `Lines` package. This was the case in the original version of this PR but was implemented outside of `Lines` subpackage in #3831. One could argue that `DemonstrateLightning` is not necessarily a lines only example (doesn't even contain any of the lines models).
> 
>     3. We can now start and look a bit closer into cleaning up the examples and their documentation.

1. I'll take care of the comparison files after everything is cleaned up.
2. No - Lightning is not only of interest for Lines.
3. I have a problem:
I tried to sync/pull my local copy with max-privato's repo, but "git didn't exit cleanly" and I got stuck, git keeps complaining, no matter what I try to do, even when I try to switch/checkout master, I do not know how to proceed - @dietmarw , do you have an advice?

dietmarw:
>        I tried to sync/pull my local copy with max-privato's repo, but "git didn't exit cleanly" and I got stuck, git keeps complaining, no matter what I try to do, even when I try to switch/checkout master, I do not know how to proceed - @dietmarw , do you have an advice?

Since I had to force-push you will not be able to pull your changes. You need to reset your branch again to be at the same state as the PR-branch. Given that your the branch on your machine is called "Lines-Examples" and the remote is called "max-privato" the commands would be: 
 1. make sure to get rid of any pending changes that you have already done on your currently checked out branch but not yet committed by either: 
     a. throw them away: `git reset --hard HEAD `
     b. keep them in a stash for later `git stash` 
 2. 
```
git checkout  Lines-Examples
git reset --hard max-privato/Lines-Examples
```

now your Lines-Example should be at the same state as the PR branch on GitHub.

AHaumer:
> 
> 
> > ```
> >    I tried to sync/pull my local copy with max-privato's repo, but "git didn't exit cleanly" and I got stuck, git keeps complaining, no matter what I try to do, even when I try to switch/checkout master, I do not know how to proceed - @dietmarw , do you have an advice?
> > ```
> 
> Since I had to force-push you will not be able to pull your changes. You need to reset your branch again to be at the same state as the PR-branch. Given that your the branch on your machine is called "Lines-Examples" and the remote is called "max-privato" the commands would be:
> 
>     1. make sure to get rid of any pending changes that you have already done on your currently checked out branch but not yet committed by either:
>        a. throw them away: `git reset --hard HEAD `
>        b. keep them in a stash for later `git stash`
> 
> 
> ```
> git checkout  Lines-Examples
> git reset --hard max-privato/Lines-Examples
> ```
> 
> now your Lines-Example should be at the same state as the PR branch on GitHub.

Thanks @dietmarw . I'm not familiar with such actions (reset, rebase, ...). 
I'll take care of the examples after a meeting later this evening.

dietmarw:
@AHaumer  Here is a nice resource you can get familiar with over the summer holidays ;-) https://www.gitkraken.com/learn/git/tutorials

Modelica/Electrical/Analog/Examples/Lines/LightningLosslessTransmissionLine.mo
```diff
@@ -36,7 +36,7 @@ model LightningLosslessTransmissionLine
     annotation (Placement(transformation(extent={{-10,-30},{10,-10}})));
   Sources.LightningImpulseCurrent
                         lightningImpulseCurrent(
-    approximation=Modelica.Electrical.Analog.Types.ImpulseApproximation.DoubleExp,
+    approximation=Modelica.Electrical.Analog.Types.ImpulseApproximation.Heidler,
     I=10e3,
```
dietmarw, line 40: The original examples from @max-privato did use DoubleExp. But probably not an issue?

AHaumer: I'd prefer to set Heidler as default, since this function is even used in the respective standard.

AHaumer (approved):
Thanks @max-privato for the examples and @dietmarw for your help!
It was a piece of work but (with the linked PRs) a big step forward in improving quality:

- improvements of the lines models, including possibility for initialization
- nice examples - hope @max-privato shares his M_OLine example with us

This PR only includes the examples, I'm fine with them.

Modelica/Electrical/Analog/Examples/Lines/LightningLosslessTransmissionLine.mo
dietmarw: I see. Fine with me.

AHaumer:
@dietmarw do you know what's the problem with the continuous-integration/jenkins?

sjoelund:
> @dietmarw do you know what's the problem with the continuous-integration/jenkins?

I'll restart it. It looks like some connection error.

AHaumer:
> 
> 
> > @dietmarw do you know what's the problem with the continuous-integration/jenkins?
> 
> I'll restart it. It looks like some connection error.

Thanks @sjoelund! So we are waiting for the review of @dietmarw

max-privato:
> * improvements of the lines models, including possibility for initialization
> * nice examples - hope @max-privato shares his M_OLine example with us
> 

As I said, I have already prepared the example and two unavoidable (and I  think very useful) support functions. But I prefer to wait this PR to be merged first, to avoid putting too much stuff in a single PR.

As soon as this PR is merged I'll create the new one

dietmarw:
The review here should really be done by @christiankral  since I'm not a library officer of Electrical.

AHaumer:
> 
> 
> The review here should really be done by @christiankral since I'm not a library officer of Electrical.

Come on, but you're one of the rare electrical engineers ...

christiankral:
@max-privato Please consider https://github.com/max-privato/ModelicaStandardLibrary/pull/1

max-privato:
> @max-privato Please consider [max-privato#1](https://github.com/max-privato/ModelicaStandardLibrary/pull/1)

Nice correction, thank You 
I merged.

christiankral:
I will complete my review as soon as I find time to do so.

max-privato:
> I will complete my review as soon as I find time to do so.

If I get the language well, this means it may be in a day, a month, a year, never ;-)

christiankral:
Some comments before completing my review:
 
### OscilloscopeProbe

- The waveform of the oscilloscope voltage depends on the `interval` setting of the experiment annotation. I wonder if this what we want to have in an example when comparing reference signals, i.e. 1E-6 s = 5.01E-4 s - 5.00E-4 s
  image
- It would possibly make sense to change the parameters such way that a de-tuned setup is investigated instead

@AHaumer @max-privato Any suggestion or opinion here?

AHaumer:
Sigh ... I thought this is a nice example, with parameters from a practical application. 
You would have to use a much smaller intervall since voltage change is settled within ns. 
Clearly enough: Choosing an interval of µs you'll miss the details, but choosing a ns or less you simulate for ages.
OscilloscopeProbe

AHaumer:
Shall we skip the example? @christiankral what's your opinion?

christiankral:
@AHaumer Yes, due to these numerical issues I am in favor of removing this example.

@max-privato What's your opinion here?

max-privato:
> @AHaumer Yes, due to these numerical issues I am in favor of removing this example.
> 
> @max-privato What's your opinion here?

Sorry, I read this question only today.  I downloaded the PR and had a look.

As a power engineer frankly I don't understand enough of the details of the example to give knowingly opinion. 
I don't even understand why for the cable the capacity is mentioned while the inductance is not!

So, for this example, @AHaumer @christiankral  I totally trust your opinions,

AHaumer:
Dor a BNC normally inductance is neglectible, but capacitance is not.
I see we have a basic problem here, independent of the line model.
I'll remove the example.

max-privato:
> Dor a BNC normally inductance is neglectible, but capacitance is not.
I see. 
It's the opposite of what happens in short power lines. This makes sense since in  power lines the conductors are much father apart than in an oscilloscope line. And distance increases inductance and decreases capacitances

max-privato:
Any news on this?

christiankral (approved):
I think it looks all good now.

max-privato:
Thanks. 
Thank you also for having a look at #3846 that logically follows this (and fixing icon and units).

christiankral:
Please have a look at #3846 and apply the requested changes.

max-privato:
I suppose I'm really bad with gitHub.
If I go there I can look at the proposed changes, which for me are ok.
Unfortunately, I don't know what the system expects me to do! Please,
Christian, in case you can, do it in my place, or, explain to me what and
how to do to apply the requested changes.
Thanks!



Il giorno gio 23 set 2021 alle ore 17:07 Christian Kral <
***@***.***> ha scritto:

PR #3824, opened 2021-05-31
M_OLine example

max-privato:
I know there is some opinion that M_OLine should be considered obsolete.
However, since I spent some time finding an interesting example of its usage, I created this PR, just in case it may help.
I did it as a Power System Engineer, using M_OLine as a Power Line model.
Unfortunately, M_OLine is not able to consider ground effects correctly, when currents flow into the ground.

Therefore, the example I created is a bit tricky, even though it has some effectiveness.
This is just a "discussion PR": I don't really know whether accepting this is the right thing for MSL. 
Naturally, all limitations of the example proposed are listed and explained in the example documentation.

To create the example I had also to introduce functions that produce Matrices for the lines. These are useful per-se, , i.e. independently on M_OLine, and could be used in the future in case some evolution of M_Line, able to accept resistances also off-diagonal the Z matrix is created (I have a version that does this on my PC already).

### DIFF
--- a/Modelica/Electrical/Analog/Examples/Lines/PowerLineWithFence.mo
+++ b/Modelica/Electrical/Analog/Examples/Lines/PowerLineWithFence.mo
@@ -1,0 +1,129 @@
+within Modelica.Electrical.Analog.Examples.Lines;
+model PowerLineWithFence
+  extends Modelica.Icons.Example;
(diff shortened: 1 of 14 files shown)


max-privato:
This is just a draft that I thought stayed personal. 
I'm closing this and reopening a new one when everything is ready.

PR #3826, opened 2021-06-04
Fix Lines

AHaumer:
As discussed in #3815, I added the `Short` with instance name `M` to the submodel `segment` for backwards compatibility.
I've removed `segment_last.M` since it is not connected.
Furthermore, it was impossible to intialize all lines models since the varables or components we have to access for initialization were protected - I've made them public, so at least initialization is possible.

After reviewing and merging this PR, I kindly ask @dietmarw to help to:

- back-port these changes to 3.2.3 in #3818 if desired.
- rebase #3819 so we can proceed with the examples.


dietmarw:
This PR will not work. Could you please simply apply the discussed changes in to branch that is based on the **current** master. Then this PR would only contain the one change of re-establishing backward compatibility which was missed in #3815

dietmarw:
Basically this PR should only contain the *one* commit e554b54cceabc01eef6d2caf0893f74ddbbc3fd6 that you have applied last. All the other commits are already present in master (bar all the merge-conflict and fix commits you had to deal with in this branch because of the intermediate merging of master into this branch). 

If you don't manage then I can fix this PR for you. But it makes more sense if you try it in order to learn the ropes.

PR #3828, opened 2021-06-04
Fix initialization of Lines as discussed in #3826

AHaumer:
The models `Analog.Lines.{OLine, ULine, TLine1, TLine2, TLine3}` lack the possibility for proper initialization, since the components that have to be accessed for initialization are protected. I've made them public.
`Analog.Lines.M_OLine` is bugfixed in #3827.
@dietmarw can you back-port this to maint/4.0.x?
I suppose after merging this PR and #3827 we should rebase #3819 or create a new PR.

PR #3829, opened 2021-06-04
Fix initialization of lines for maintenance

AHaumer:
This is the same fix as #3828 for master but without the unification of TLines for maintenance/4.0.x.


dietmarw:
@AHaumer Sorry this is wrong. To not waste time I will fix both PR such that they will be correct. So please sit back and let me do this for now.

AHaumer:
> 
> 
> @AHaumer Sorry this is wrong. To not waste time I will fix both PR such that they will be correct. So please sit back and let me do this for now.

Sorry, I would appreciate if you explain to me how I should have done.

dietmarw:
You did not base your commit on the latest master is the simple answer. You used some other branch with lots of merge commits. 
1. checkout master and make sure to do a git pull in order to are in sync with upstearm master
2. create a branch
3. apply the commit(s)
4. push the branch and create the PR

dietmarw:
Mind, for this you do not need to do it now since I've already have taken care of it.

PR #3835, opened 2021-06-10
Remove Evaluate=true for fixed=false parameter 'n_b'

henrikt-ma:
From the brief discussion in https://github.com/modelica/ModelicaSpecification/issues/2951, it seems that `Evaluate = true` on a parameter with `fixed = false` is not expected to have any effect.  Thus I'd like to propose that the `Evaluate = true` is removed in order to get better correspondence between presence of `Evaluate = true` and parameters actually becoming evaluated.


beutlich:
@henrikt-ma Thanks for the PR. Please do not forget to set library officers as reviewers and add the appropriate labesl. Thanks again.

henrikt-ma:
> @henrikt-ma Thanks for the PR. Please do not forget to set library officers as reviewers and add the appropriate labels. Thanks again.

Thanks doing it for me this time, I'll try to figure out such details myself next time.  Is there a page somewhere describing how to do this properly?

tobolar (approved):

MartinOtter (approved):

PR #3842, opened 2021-06-13
Fix permanent link redirect

beutlich:
As reported by https://github.com/modelica/ModelicaStandardLibrary/runs/2812636856?check_suite_focus=true


AHaumer (approved):
Fine with me

christiankral (approved):
The new link works fine and thus I approve this PR.

@beutlich Would it be more useful to use a DOI link instead it the "regular" links of the publisher change every once in a while?

PR #3845, opened 2021-06-21
QuasiStatic example induction machine with inverter

AHaumer:
Just realized that an example of an induction machine with frequency inverter is missing. 
This example allows to compare rotor frequency and slip in base region and field weakening.


christiankral (approved):
Looks good

PR #3848, opened 2021-06-28
Remove `align=bottom` from guidelines and code

### LINKED ISSUE #3844: Conflicting guidelines for HTML documentation conventions for tables
In `Modelica.UsersGuide.Conventions.Documentation.Format.Tables` the guidelines say that the caption should be defined as:

```
<caption align="bottom"><strong>Tab 2:</strong> Caption starts with a capital letter</caption>
```
But this would place the caption at the **bottom** and not the top. The problem is that Dymola renders the position incorrectly and always displays it on the top. Now the question, should the caption appear at the top of the table or the bottom. If at the top then it needs to be updated and tools need to make sure to get the rendering right.

One additional problem is that `align` is no longer supported in HTML5, so the guideline should most likely updated  anyway.

dietmarw:
This fixes #3844 by removing the recommendation and application of `align=\"bottom\"` since tools tend to interpret this differently (e.g., ignore) and the `align` parameter is also deprecated in HTML5. Not using and recommending it  makes it more future proof.

### DIFF
--- a/Modelica/Electrical/Analog/Basic/SaturatingInductor.mo
+++ b/Modelica/Electrical/Analog/Basic/SaturatingInductor.mo
@@ -76,7 +76,7 @@ This approximation is with good performance and easy to adjust to a given charac
 </p>
 
 <table border=\"1\" cellspacing=\"0\" cellpadding=\"2\">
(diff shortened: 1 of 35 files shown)


christiankral (approved):
Looks good

HansOlsson (approved):
Seems ok

PR #3868, opened 2021-09-21
startBackward and startForward should not be reference results (#3866)

uschna:
remove startBackward and startForward from the reference results, because the transition from these states to mode = Backward or Forward depends on the random value of the velocity difference in the state Stuck


HansOlsson (approved):
Looks good.

henrikt-ma (commented):
As explained in https://github.com/modelica/ModelicaStandardLibrary/issues/3866#issuecomment-923767409, I don't have much to say regarding this except that these variables have always been causing problems when comparing our simulation results to the reference results.  I can't tell whether this should be solved by more robust modeling or by removing the variables from comparison.

PR #3871, opened 2021-09-23
Remove ModelicaUtilities.h from MSL distribution

### LINKED ISSUE #3867: Is ModelicaUtilities.h tool-specific?
ModelicaUtilities.h is mentioned in the language specification, [Sect. 12.9.6](https://specification.modelica.org/maint/3.5/functions.html#utility-functions), and available in [Modelica/Resources /C-Sources](https://github.com/modelica/ModelicaStandardLibrary/tree/master/Modelica/Resources/C-Sources).

Is it expected to be customized by each tool vendor, or should it be left as on the MA repo?

What is the recommended way to #include it in a .c source code file within a certain library's Resources directory?

beutlich:
ModelicaUtilities.h is ought to be distributed by Modelica tools and no longer by the MSL.

Closes #3867.


sjoelund (approved):

sjoelund:
Just need to fix the CI jobs

beutlich:
> 
> 
> Just need to fix the CI jobs

Can you please tell me how to configure the include directory in https://github.com/modelica/ModelicaStandardLibrary/blob/364a7f372da8f566e15a76f6b014ba342b75c8cb/.github/workflows/checkCI.yml#L19-L20

Thanks.

sjoelund:
> > Just need to fix the CI jobs
> 
> Can you please tell me how to configure the include directory in
> 
> https://github.com/modelica/ModelicaStandardLibrary/blob/364a7f372da8f566e15a76f6b014ba342b75c8cb/.github/workflows/checkCI.yml#L19-L20
> 
> Thanks.

`./configure ... CPPFLAGS="..."` (a bit busy in the conference)

henrikt-ma (commented):
Looks like this is mostly affecting CI, which I don't know much about.  I'm just a bit surprised that the file is kept at all – isn't the intention that also CI should have access to a Modelica tool environment when building the libs?

beutlich:
Current CI is tool-invariant. It only is some unit tests without the requirement to have a Modelica tool available.

sjoelund:
> Looks like this is mostly affecting CI, which I don't know much about. I'm just a bit surprised that the file is kept at all – isn't the intention that also CI should have access to a Modelica tool environment when building the libs?

It could be done that way, but I think it's easier to just keep a small header in the CI.

beutlich:
> 
> 
> Looks like this is mostly affecting CI, which I don't know much about. I'm just a bit surprised that the file is kept at all – 

Only in the repository (for CI reasons), but no longer in the MSL distribution.

> isn't the intention that also CI should have access to a Modelica tool environment when building the libs?

Current CI is tool-invariant. It only is some unit tests without the requirement to have a Modelica tool available.

beutlich:
@sjoelund Maybe you can fix teh Jenkinsfile since I do not know how you like the fix, in `allAutotools`, `autotools` or each caller?

beutlich:
Reminder for me to fix the conficts of #3686 afterwards.

henrikt-ma (approved):
> > isn't the intention that also CI should have access to a Modelica tool environment when building the libs?
> 
> Current CI is tool-invariant. It only is some unit tests without the requirement to have a Modelica tool available.

OK, it seems like one could consider the CI its own Modelica tool environment in this case, so having a non-distributed copy of the file for this purpose makes sense.

beutlich:
Will change the VS project again to take the ModelicaUtilities.h from an environment variable `MODELICA_UTILITIES_INC`.

beutlich:
@sjoelund I introduced environment variable `MODELICA_UTILITIES_INCLUDE` in the build scripts. Can you please review again, especially configure.ac. Thanks.

sjoelund (changes requested):
I don't see the value of using your own environment variable when there is an established standard for CPPFLAGS (for make, cmake and configure scripts). I could see the point for vcproj though. It could perhaps be used as a fallback if configure or cmake do not find the header (less easy to detect for makefiles)

beutlich:
OK, only for VS then. Will do

beutlich:
The issue I see with BuildProjects/gcc/Makefile is that you now need to set ``make CPPFLAGS="-DNDEBUG -DHAVE_UNISTD_H -DHAVE_STDARG_H -DHAVE_HIDDEN -DHAVE_MEMCPY -Ipath/to/ModelicaUtilitiesHeader"` on the command line as there is no append on the CPPFLAGS.

beutlich:
> 
> 
> Reminder for me to fix the conficts of #3686 afterwards.

Done.

PR #3886, opened 2021-10-27
Generate MSL Release Notes by GitHub Actions

beutlich:
See https://github.com/beutlich/ModelicaStandardLibrary/actions/runs/1391688197 for an example.


thorade (approved):
The script and the result look good to me

PR #3898, opened 2021-11-16
Fix broken link to current release

tobolar:


dietmarw (changes requested):
I'm afraid none of the relative links will work. Also not your new version. If you think this is important then you should replace the all with absolute links (just like it is already done in the README in the master branch).

tobolar:
> If you think this is important  ...

I don't know the idea behind the README in maint branches. IMO if the text is changed to `Modelica Standard Library v3.2.3+build.4 (2020-06-04)` compared to master, then the link shall either be working, or shall be deleted.
But your are more familiar to this and my PR ist just a suggestion. So simply decide it on your own.

dietmarw:
The problem is that the PR as it is now will not work.

tobolar:
No problem, I will close this PR.

PR #3902, opened 2021-11-25
Electrical.Batteries.Utilities.PulseSeries

### LINKED ISSUE #3901: Possible bug in Modelica.Electrical.Batteries.Utilities.PulseSeries
In the model `Modelica.Electrical.Batteries.Utilities.PulseSeries` there seems to be a bug. Specifically in the declaration:

```
parameter SI.Time Tstart2[n1] = array(... for k in 1:n2);
```
In that case `Tstart2` has a size of `n1` but the iterator goes up to size `n2`. Same issue with the equation
```
on2 = oneTrue(array(time >= Tstart2[k] and time < Tstart2[k] + T2 for k in 1:n1));
```
In order for the model to work `n1` and `n2` must be equal.

In the examples where the model is used, `n1` and `n2` happen to be equal.

AHaumer:
fixes #3901 - thanks for pointing out.


dietmarw (dismissed):

christiankral:
In the actual implementation the pause time between the two series is `Tp1 + Tp`: Consequently the documentation is updated in f6b3e1105f92bda7116b2d3c3afa438d9b262e23. 

image

As the many parameters may be confusing I am adding this image above to the documentation.

@ahaumer I wonder if this is intended, or if the pause time between the two pulse series shall be simply `Tp`

@ahaumer The parameter `offset` is actually obsolete, as the two ampitudes can be set independently, however.

dietmarw:
@christiankral From the image it is not quite clear to me what Tp2 is referring to exactly.

christiankral:
> @christiankral From the image it is not quite clear to me what Tp2 is referring to exactly.

OK then:

image

dietmarw (approved):

MartinOtter (approved):

PR #3905, opened 2021-11-27
ModelicaIO.c: Fix error handling for invalid format specifier in text file

### LINKED ISSUE #3903: Modelica.Blocks.Tables: Wrong format specifier is ignored
If for a Modelica.Blocks.Tables.CombiTable1D (tableOnFile=true) the format of the dimensions within the file is wrongly given then this is ignored. ModelicaStandardTables_CombiTable1D_getValue() returns 0. for this table.

Example file:

```
------------------------
#1
double	TestTable(2.2)
0	1
1	2
------------------------
```

Here "(2.2)" is used instead of the correct "(2,2)".

Within the constructor of the CombiTable1D the function 
	void* ModelicaStandardTables_CombiTable1D_init(...)
is called.

Within this function the function
	static double* readTxtTable(_In_z_ const char* fileName, _In_z_ const char* tableName,
								_Out_ size_t* m, _Out_ size_t* n)
is called. Here
	nRow = (unsigned long)_strtol_l(token, &endptr, 10, loc);
is executed where token is "2.2". As result, nRow is 2 and *endptr is '.'. So because of
        if (*endptr != 0) {
            continue;
the next lines of the file are read without further actions.

At the end nRow and thus *m is 2 and nCol and *n is 0 and readTxtTable() and finally ModelicaStandardTables_CombiTable1D_init() return NULL, i.e. tableID is NULL.

What happens afterwards: The Modelica code says

initial algorithm
if tableOnFile then
	tableOnFileRead := readTableData(tableID, false, verboseRead);

The function readTableData calls 
	double ModelicaStandardTables_CombiTable1D_read(void* _tableID, int force, int verbose)
Here _tableID is NULL and only
   return 1.; /* Success */
is executed.

Thus the return value of readTableData() is 1. But even if a 0. would be returned -- the value of tableOnFileRead is never checked in the model.

Now the simulation continues and when
	double ModelicaStandardTables_CombiTable1D_getValue(void* _tableID, int iCol, double u)
is called then _tableID is NULL and therefore only
    double y = 0.;
    return y;
is executed, i.e. 0. is returned.

So the user thinks everything is fine but instead wrong results are given: 0. instead of the real table values.

beutlich:
Fix #3903.

@gkurzbach FYI

### DIFF
--- a/Modelica/Resources/C-Sources/ModelicaIO.c
+++ b/Modelica/Resources/C-Sources/ModelicaIO.c
@@ -37,6 +37,10 @@
       Modelica.Utilities.Streams.writeRealMatrix
 
    Changelog:
+      Nov. 27, 2021: by Thomas Beutlich
+                     Fixed error handling for invalid format specifier in
+                     text file (ticket #3903)
+
       Dec. 22, 2020: by Thomas Beutlich
                      Added reading of CSV files (ticket #1153)
 
@@ -510,7 +514,7 @@ static void readMatIO(_In_z_ const char* fileName,
         if (NULL == token) {
             free(matrixNameCopy);
             ModelicaFormatError(
-                "Variable \"%s\" not found on file \"%s\".\n",
+                "Variable \"%s\" not found in file \"%s\".\n",
                 matrixName, fileName);
         }
         else {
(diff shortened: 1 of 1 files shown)


sjoelund (approved):
I'm a bit uncertain about returning after ModelicaError, but that was there before so not a new issue.

HansOlsson (approved):
Looks good.

PR #3910, opened 2021-12-06
Fix parameter propagation of IntersectivePWM

christiankral:
The parameter `3` is propagated to the `fromSpacePhasor` block used in the model `IntersectivePWM`. ~~This is not correct, as the model is designed for a variable number of phases, `m`.~~ So instead the parameter `m` shall be propagated.

### DIFF
--- a/Modelica/Electrical/PowerConverters/DCAC/Control/IntersectivePWM.mo
+++ b/Modelica/Electrical/PowerConverters/DCAC/Control/IntersectivePWM.mo
@@ -15,34 +15,34 @@ block IntersectivePWM "Intersective PWM"
   Modelica.Blocks.Interfaces.BooleanOutput fire_n[m] "Negative fire signal"
     annotation (Placement(transformation(extent={{100,-70},{120,-50}})));
   Modelica.Electrical.Machines.SpacePhasors.Blocks.FromSpacePhasor
-    fromSpacePhasor(m=m)
+    fromSpacePhasor(final m=m)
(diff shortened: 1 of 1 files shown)


christiankral:
I also noticed, that the propagated parameters of the blocks `SawTooth` and `Trapezoid` (also) are not `final`.

christiankral:
> The parameter `3` is propagated to the `fromSpacePhasor` block used in the model `IntersectivePWM`. ~This is not correct, as the model is designed for a variable number of phases, `m`.~ So instead the parameter `m` shall be propagated.

As the parameter `m` is a `constant`, not a `parameter`, the model is valid in the original version. For the sake of consistency the parameter `m` is used instead `3. In case of a model extension the parameter propagation has not to be changed.

henrikt-ma:
> As the variable `m` is a `constant`, not a `parameter`, the model is valid in the original version. For the sake of consistency the parameter `m` is used instead `3`.

But as long as `m` isn't a final constant, there really is a difference between using `m` and using `3`; The `m` can be modified, for instance, when instantiating the class whereas `3` never changes.  That is, to me this change looks good and like more than something just for the sake of consistency.  On the other hand, the comment _For the sake of consistency the parameter m is used instead 3_ makes me wonder whether the intention is to also make the constant `m` final?

AHaumer (approved):
fine with me

dietmarw:
@christiankral The PR still has draft status. Is it ready for review?

dietmarw (approved):
Sorry for the late reaction, I missed it getting ready for review.

PR #3913, opened 2021-12-13
Implementation and demo of IntervalTest

AHaumer:
In Simulink a block IntervalTest (i.e. checking whether the input lies within or without the specified interval) exists, in Modelica you have to implement it again and again using Modelica.Blocks.Logical.{And, Or, GreaterThreshold, GreaterEqualThreshold, LessTreshold, LessEqualThreshold, Greater, GreaterEqual, Less, LessEqual}. 
I'm pretty sure that this is a block used in many applications, here is a convenient implementation of this functionality.

### DIFF
--- a/Modelica/Blocks/Logical.mo
+++ b/Modelica/Blocks/Logical.mo
@@ -358,6 +358,83 @@ Real input u2, otherwise the output is <strong>false</strong>.
 </html>"));
   end LessEqual;
 
+  block IntervalTest
+    "Output y is true, if input u is within the specified interval"
+     extends Modelica.Blocks.Interfaces.partialBooleanSO;
+    parameter Boolean useConstantLimits=true "Otherwise dynamic inputs" annotation(Evaluate=true);
+    parameter Real constantLowerLimit "Lower limit of interval" annotation(Dialog(enable=useConstantLimits));
+    parameter Boolean ClosedOnLeft=false "Include lower limit?" annotation(Evaluate=true);
+    parameter Real constantUpperLimit "Upper limit of interval" annotation(Dialog(enable=useConstantLimits));
+    parameter Boolean ClosedOnRight=false "Include upper limit?" annotation(Evaluate=true);
+    parameter Boolean InsideInterval=true "u inside interval?" annotation(Evaluate=true);
+    Modelica.Blocks.Interfaces.RealInput u annotation (Placement(transformation(
+            extent={{-140,-20},{-100,20}}), iconTransformation(extent={{-140,-20},{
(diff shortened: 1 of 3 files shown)


christiankral (approved):
Looks OK from my side.

Modelica/Resources/Reference/Modelica/Blocks/Examples/DemoIntervalTest/comparisonSignals.txt
```diff
@@ -0,0 +1,3 @@
+time
+and1.y
+intervallTest.y
```
beutlich, line 3:
```suggestion
intervalTest.y
```

Modelica/Blocks/Logical.mo
```diff
@@ -358,6 +358,84 @@ Real input u2, otherwise the output is <strong>false</strong>.
...
+              {-86,84}})));
+  equation
+    if useConstantLimits then
+      upperLimitInternal = constantUpperLimit;
+      lowerLimitInternal = constantLowerLimit;
+    else
+      connect(upperLimitInternal, upperLimit);
+      connect(lowerLimitInternal, lowerLimit);
+    end if;
+    assert(upperLimitInternal>lowerLimitInternal, "Erroneous interval limits");
```
tobolar, line 393: If this assertion is really needed, I would set this as a warning only - insted of hard termination of the simulation. Optionally, a parameter of type "AssertionLevel" could be add to give the user an opportunity to decide - but this approach is not used elsewhere in MSL so maybe not that good idea.

Modelica/Blocks/Logical.mo
```diff
@@ -358,6 +358,84 @@ Real input u2, otherwise the output is <strong>false</strong>.
...
+    Modelica.Blocks.Interfaces.RealInput upperLimit if not useConstantLimits
+      annotation (Placement(transformation(extent={{-140,60},{-100,100}})));
+
+  protected
+    Modelica.Blocks.Interfaces.RealInput lowerLimitInternal annotation (Placement(
+          transformation(extent={{-94,-84},{-86,-76}}), iconTransformation(extent={
+              {-94,56},{-86,64}})));
+    Modelica.Blocks.Interfaces.RealInput upperLimitInternal annotation (Placement(
+          transformation(extent={{-94,76},{-86,84}}), iconTransformation(extent={{-94,76},
+              {-86,84}})));
```
henrikt-ma, line 384: Maybe it could be considered good style to at least not have an `iconTransformation` on a protected component?  Maybe remove the entire `Placement` annotation?

PR #3914, opened 2021-12-13
Implementation and test of UnwrapAngle (angle tracking observer)

AHaumer:
In Simulink a block UnwrapAngle exists, in Modelica it's missing. I'm pretty sure that this is a block used in many applications, here is a robust implementation (angle tracking observer) of this functionality.


tobolar (commented):
Disregarding the implementation itself, my comments here:
1. It is great to have such a block.
2. IMO, the documentation is leaned too close to electrics. Maybe better to write it in a general way, just like e.g. WrapAngle do.
3. The output `w`is useful, and shall probably be renamed to "dy", "der_y", "y_der", or something like that. (Also delete that string from the icon completely.)
4. A bit odd to use an element (Rotator) from Electrical within Blocks. But I understand it is needed here.
5. Example `DemoATO` shall be renamed to not include an abbreviation. ATO is not explained. (And yes, I have no clue what it is.)

AHaumer:
ATO = Angle Tracking Observer
And yes, the second output is w = d phi / dt.

dietmarw:
I'm a bit torn here. I really dislike the "shopping" for the `Rotator` of a base component from deep within another domain library. So ideally I would see the `Rotator` (and also the other phasor blocks) into `Blocks.Math`. But this would automatically mean this extension being pushed towards the next MAJOR version. And who knows when that is going to happen. 

Alternatively one could accept the "dirty" implementation for the 4.x series but create a 5.0.0 ticket at the same time where the move of those blocks will be executed. 

@AHaumer One minor thing, the documentation of `Rotator` says: "Rotates a space phasor (voltage or current) input u by the angle in **negative** mathematical direction." but the images shows the rotation happening in positive (ant-clockwise) direction. So which one is it?

dietmarw (commented):
I'm a bit torn here. I really dislike the "shopping" for the `Rotator` of a base component from deep within another domain library. So ideally I would see the `Rotator` (and also the other phasor blocks) into `Blocks.Math`. But this would automatically mean this extension being pushed towards the next MAJOR version. And who knows when that is going to happen.

Alternatively one could accept the "dirty" implementation for the 4.x series but create a 5.0.0 ticket at the same time where the move of those blocks will be executed.

@AHaumer One minor thing, the documentation of `Rotator` says: "Rotates a space phasor (voltage or current) input u by the angle in negative mathematical direction." but the images shows the rotation happening in positive (anti-clockwise) direction. So which one is it?

AHaumer:
> @AHaumer One minor thing, the documentation of `Rotator` says: "Rotates a space phasor (voltage or current) input u by the angle in negative mathematical direction." but the images shows the rotation happening in positive (anti-clockwise) direction. So which one is it?

I hope I don't mix it up now but it seems to be consistent:
angle is measured from (alfa, beta) - coordinate system to (d, q) - coordinate system, and the image explains that angle has to be subtracted from the input's argument (i.e. rotate the input to obtain it's representation in the new coordinate system).

dietmarw:
@AHaumer I still think you mix things up. The negative mathematical direction is clockwise. The arrow that is shown in the documentation is pointing anti-clockwise (i.e., positive direction) whilst the icon show clockwise (i.e., negative direction). Now the main confusion here is that what is rotated is not the actual signal but the signal with regard to its reference coordinate system. 

I suggest changing the formulation so it says:
"Rotates a space phasor (voltage or current) input `u` by the `angle` in **positive** mathematical direction. **This is achieved by rotating the rotating reference by the negative value of angle `angle`.** This block represents the transformation of one space phasor u from one rotating reference (coordinate) frame into another where the space phasor is `y`. The output reference frame leads the input reference frame by angle `angle`.

(changes in bold)

Now this should be a different ticket really.

christiankral:
> I suggest changing the formulation so it says: "Rotates a space phasor (voltage or current) input `u` by the `angle` in **positive** mathematical direction. **This is achieved by rotating the rotating reference by the negative value of angle `angle`.** This block represents the transformation of one space phasor u from one rotating reference (coordinate) frame into another where the space phasor is `y`. The output reference frame leads the input reference frame by angle `angle`.
> 
> (changes in bold)

I think it's a good idea to improve the documentation of the `Rotator` model. If we do so, the wording *(voltage or current)* shall be changed to *(e. g., voltage or current)*, as the transformation works also for flux phasors, etc. We possibly perform these changes in a different ticket.

christiankral:
> 3\. The output `w`is useful, and shall probably be renamed to "dy", "der_y", "y_der", or something like that. (Also delete that string from the icon completely.)

I agree that `der_y` (my preference) might be a more meaningful name for the angular velocity.

tobolar:
@AHaumer  Regarding example `Modelica.Blocks.Examples.DemoUnwrapAngle`, why is it necessary to have components from the `Mechanics.Rotational` package? It works all fine also without it.

tobolar:
> I agree that der_y (my preference) might be a more meaningful name for the angular velocity.

Me too.

casella:
I had a problem with atan2() angle jumps when dealing with drone models with @arun3688 and had to come up with an ugly non-OO hack that picks the rigid body angle states to get an unwrapped angle. So this is also interesting for me. I'll get back to this once MSL 4.1.0 is out.

PR #3915, opened 2021-12-13
Implemented and tested MultiAnd / MultiOr

AHaumer:
Comparing more than 2 Boolean signals with logical blocks {and, or} quickly gets tedious. 
For these cases I've implemented Modelica.Blocks.Logical.{MultiAnd, MulitOr}, 
using Modelica.Math.BooleanVectors.allTrue and Modelica.Math.BooleanVectors.anyTrue.


Modelica/Blocks/Logical.mo
```diff
@@ -21,7 +21,10 @@ the output is <strong>false</strong>.
   end And;
 
   block MultiAnd "Logical and of Boolean vector elements"
-    extends Modelica.Blocks.Interfaces.PartialBooleanMISO;
```
HansOlsson: Why are you switching from connector with automatic sizing  to vector input without that feature?
To me that just seems like worse usability.

If you want another name for the input that's fine - but it can still have automatic sizing.

(Oh, and how do you comment on an entire commit - not on specific lines?)

henrikt-ma:
Related to this, but not to be handled in this PR, isn't it a mistake that `allTrue` and `anyTrue` are implemented using an algorithm with iteration rather than using `min` and `max` array reductions?

AHaumer:
@henrikt-ma: To be honest, I don't know. Feel free to propose a PR.
Also strange: allTrue and andTrue are (nearly) the same.

Modelica/Blocks/Logical.mo
```diff
@@ -20,6 +20,23 @@ the output is <strong>false</strong>.
 </html>"));
   end And;
 
+  block MultiAnd "Logical and of Boolean vector elements"
+    extends Modelica.Blocks.Interfaces.partialBooleanSO;
+    parameter Integer nin=2 "Number of inputs";
+    Modelica.Blocks.Interfaces.BooleanInput u[nin]
+      annotation (Placement(transformation(extent={{-140,-20},{-100,20}})));
+  equation
+    y=Modelica.Math.BooleanVectors.allTrue(u);
```
henrikt-ma, line 29: Don't use the weird `allTrue`, use the strangely named but so much naturally defined `andTrue` instead:
```suggestion
    y = Modelica.Math.BooleanVectors.andTrue(u);
```

henrikt-ma:
> @henrikt-ma: To be honest, I don't know. Feel free to propose a PR.

OK, I'll do it for the sake of demonstrating the power of the language.

> Also strange: allTrue and andTrue are (nearly) the same.

Good point, I turned that into a suggestion.

henrikt-ma:
When introducing `MultiAnd` and `MultiOr`, I would also have expected a `MultiNand`, a `MultiNor` and a `MultiXor`. The latter should be based on `oneTrue`, and yes, the name `MultiXor` might be a bit controversial, but then `OneTrue` could be a less controversial name.  (For completeness, one could also imagine having an `OddTrue` corresponding to the alternative definition of variadic exclusive disjunction.)

henrikt-ma (commented):
I think others are better suited than me for approving or requesting changes on MSL PRs – I'll try to stick to just leaving comments.  Right now, I am satisfied with the use of the weirdly named `andTrue` instead of `allTrue`, and my only remaining remark is the concern about only introducing variadic variants of the `And` and `Or` blocks, while leaving `Xor`, `Nand` and `Nor` without variadic counterparts.

sjoelund:
> Also strange: allTrue and andTrue are (nearly) the same.

The normal would be to have `allTrue(arr)` be `min(arr)`. But for some reason it wasn't. It's actually `allTrueAndArrayNotEmpty(arr)`... So there is a second function there. It's really confusing :)

> Maybe because OM has troubles with vectorized function calls: see Modelica.Electrical.PolyPkase.Basic.{Star, Delta} and some more.

I don't see any vectorized calls in those. Also, reductions are not vectorized calls. `min(arr)` is not treated as a reduction in OpenModelica (it simply has its own implementation). Vectorized calls in OpenModelica are transformed into reduction expressions: `f(arr) = {f(i) for i in arr}`.

HansOlsson (changes requested):
To me the lack of connectorSizing is problematic.
If it were the only And-block I could sort of understand it as you normally would connect two inputs and it would be good that it was pre-configured, but we have another variant for that so the Multi-variants don't need to be focused on that.
Note that there are two relevant use-cases covered by connector sizing:

- Just connecting a number of scalar signals
- Connecting an entire vector input to the model to this input on a sub-component

In both cases both adjusting `nin` and connecting seems unnecessary, and would be handled automatically by connectorSizing. If there is some reason not to that should at least be documented.
Note that the vector to vector-case indicate that `nin=0 `and `nin=1` are relevant.

Modelica/Blocks/Logical.mo
tobolar: It seems there are two possibilities how to implement this and I agree with @HansOlsson that it is indeed confusing to combine them both. Esp. if the classes' naming suggests the same handling.

HansOlsson: To me this suggests that we could improve the GUI for connectorSizing (both in tools and in the non-normative part of the specification).
Note that the second proposal written in textual form should still work with connectorSizing.

AHaumer: No it doesn't (I tried it).

AHaumer:
@HansOlsson  don't forget the case when you want to connect more than one vector.

AHaumer:
@henrikt-ma I have to admit I was too busy to consider Nand, Nor, and Xor. And yes you are right, MultiXor is somehow confusing. How would you define it?

HansOlsson:
> @henrikt-ma I have to admit I was too busy to consider Nand, Nor, and Xor. And yes you are right, MultiXor is somehow confusing. How would you define it? And we shouldn't forget MultiNot.

I think we shouldn't add `MultiXor `yet because there are multiple possibilities (parity seems more common) and we don't have users demanding it yet.
I don't see a need for `MultiNot `- the normal `Not` work, and if you want to negate a vector you just make a vector of `Not`s

henrikt-ma:
> @henrikt-ma I have to admit I was too busy to consider Nand, Nor, and Xor. And yes you are right, MultiXor is somehow confusing. How would you define it?

For me, defining it in terms of `oneTrue` would be the natural choice, but because of the risk of confusion, it might be better to instead make two variants (neither of them called `MultiXor`): `OneTrue` and `OddTrue`

> And we shouldn't forget MultiNot.

Yes, let's forget `MultiNot`.

HansOlsson:
Regarding multi-xor: https://electronics.stackexchange.com/questions/93713/how-is-an-xor-with-more-than-2-inputs-supposed-to-work
But as stated before I don't think we should add it until needed.

AHaumer:
@HansOlsson you're right, MultiNot is not necessary.
And I'd avoid MultiXor - for me it's extermely confusing ;-)
Ok we should add MultiNand and MulitNor after clarifying the input (vector of automaticSizing).

tobolar:
> don't forget the case when you want to connect more than one vector.

@AHaumer Would in that case `Modelica.Blocks.Routing.Multiplex2` work?

AHaumer:
@tobolar I'll try that - I think that'll work, you're right!

HansOlsson:
> @HansOlsson don't forget the case when you want to connect more than one vector. I feel somehow uncomfortabel with "solve that with the GUI". Any sound solution?

To me there are two cases, and I see neither as a blocker: either you write textually and then connectorSizing is a non-issue, or you use the GUI and then it's a matter of improving the GUIs to handle this.

AHaumer:
@HansOlsson sigh ... ok let's do it that way you suggest. 
Although it's not very intuitive for newbies until the GUI is improved.

henrikt-ma:
> Regarding multi-xor: https://electronics.stackexchange.com/questions/93713/how-is-an-xor-with-more-than-2-inputs-supposed-to-work But as stated before I don't think we should add it until needed.

Right, all the evidence suggests that at least the name `MultiXor` should be avoided.  Better to have both `OneTrue` and `OddTrue`, and in my opinion it would make sense to define them along with `MultiAnd` and `MultiOr` so that the MSL provides one canonical block for performing each of the operations.

Modelica/Resources/Reference/Modelica/Blocks/Examples/DemoMultiAndOr/comparisonSignals.txt
```diff
@@ -0,0 +1,5 @@
+time
+and2.y
+multiAnd.y
+or2.y
+multiOr.y
```
HansOlsson, line 5: Aren't the multiNor and multiNand variables missing here?

HansOlsson (approved):
seems good, except for the comment about missing reference signals that aren't that critical.

Modelica/Resources/Reference/Modelica/Blocks/Examples/DemoMultiAndOr/comparisonSignals.txt
AHaumer: You're right. done.

Modelica/Blocks/package.mo
```diff
@@ -927,6 +927,88 @@ Note, that
...
+<p>
+Compare and / or / nand / nor for (only) 3 Boolean input signals, 
+logical operations performed pairwise and as vector (or arbitrary length). 
+Compare:
+</p>
+<ul>
+<li>and2.y vs. multiAnd.y</li>
+<li>notA.y vs. multiNand.y</li>
+<li>or2.y  vs. multiOr.y</li>
+<li>notB.y vs. multiNor.y</li>
```
christiankral, line 1007: I am not sure of compare `a` vs `b` is the correct terminology. This is how I understand it:

1. Compare `a` vs `b` means to me that `a` is on the vertical axis, `b` is on the horizontal axis
2. Compare `a` with `b` means to me that `a` and `b` are plotted as a function of time (in the same graph)
3. Compare `a` and `b` means to me the same as 2.

I am kindly asking for a native speaker to comment on this

MartinOtter (approved):

Modelica/Blocks/Logical.mo
```diff
@@ -38,20 +52,50 @@ the output is <strong>false</strong>.
...
-    y = not ((u1 and u2) or (not u1 and not u2));
+    y = not (u1 and u2);
     annotation (Icon(coordinateSystem(preserveAspectRatio=true, extent={{-100,-100},
               {100,100}}), graphics={Text(
             extent={{-90,40},{90,-40}},
-            textString="xor")}), Documentation(info="<html>
+            textString="nand")}), Documentation(info="<html>
 <p>
-The output is <strong>true</strong> if exactly one input is <strong>true</strong>, otherwise
+The output is <strong>true</strong> if at least one input is <strong>false</strong>, otherwise
```
henrikt-ma, line 78: This could be formulated more consistently with `MultiOr` (since there is always a non-zero number of inputs):
```suggestion
The output is <strong>true</strong> if any of the inputs is <strong>false</strong>, otherwise
```

Modelica/Blocks/Logical.mo
```diff
@@ -20,6 +20,20 @@ the output is <strong>false</strong>.
...
 
+  block MultiAnd "Logical and of Boolean vector elements"
+    extends Modelica.Blocks.Interfaces.PartialBooleanMISO;
+  equation
+    y = Modelica.Math.BooleanVectors.andTrue(u);
+    annotation (Icon(coordinateSystem(initialScale=0.1),
+      graphics={ Text(extent={{-90,40},{90,-40}},textString="and")}),
+      Documentation(info="<html>
+<p>
+The output is <strong>true</strong> if all elements of the input vector are <strong>true</strong>, otherwise
```
henrikt-ma, line 31: Must also mention that the input vector must be non-empty:
```suggestion
The output is <strong>true</strong> if the input vector is not empty and all elements of the input vector are <strong>true</strong>, otherwise
```

Modelica/Blocks/Logical.mo
```diff
@@ -20,6 +20,20 @@ the output is <strong>false</strong>.
...
+the output is <strong>false</strong>.
+</p>
```
henrikt-ma, line 33: There should be a warning about not being the negation of `MultiNand`:
```suggestion
<em>Note that the result is not the negation of <strong>MultiAnd</strong> in the case of an empty input vector.</em>
</p>
```

Modelica/Blocks/Logical.mo
```diff
@@ -38,20 +52,50 @@ the output is <strong>false</strong>.
...
+  equation
+    y=not Modelica.Math.BooleanVectors.allTrue(u);
+    annotation (Icon(coordinateSystem(initialScale=0.1), graphics={
+                 Text(extent={{-80,30},{60,-30}},
+            textString="nand",
+            textColor={0,0,0})}),
+      Documentation(info="<html>
+<p>
+The output is <strong>true</strong> if at least one element of the input vector is <strong>false</strong>, otherwise
+the output is <strong>false</strong>.
```
henrikt-ma, line 95: This is wrong for the case of zero inputs.  Better to use formulations with _all_ or _any_ as for the other blocks:
```suggestion
The output is <strong>true</strong> if any element of the input vector is <strong>false</strong>, otherwise
the output is <strong>false</strong>.
```

Modelica/Blocks/Logical.mo
```diff
@@ -38,20 +52,50 @@ the output is <strong>false</strong>.
...
+</p>
```
henrikt-ma, line 96: There should be a warning about not being the negation of `MultiAnd`:
```suggestion
<em>Note that the result is not the negation of <strong>MultiNand</strong> in the case of an empty input vector.</em>
</p>
```

Modelica/Blocks/Logical.mo
HansOlsson: Wouldn't it be a lot better to handle the case where the input vector is empty (in the usual consistent mathematical way) instead of documenting the exception?

Modelica/Blocks/Logical.mo
henrikt-ma: Alternatively, it could be formulated analogous to `MultiNor`.

Modelica/Blocks/Logical.mo
```diff
@@ -68,20 +112,36 @@ the output is <strong>false</strong>.
...
+</html>"));
+  end MultiNor;
+
+  block Xor "Logical 'xor': y = u1 xor u2"
     extends Blocks.Interfaces.partialBooleanSI2SO;
   equation
-    y = not (u1 and u2);
+    y = not ((u1 and u2) or (not u1 and not u2));
```
henrikt-ma, line 134: Wouldn't it be easier expressed like this?
```suggestion
    y = (u1 and not u2) or (not u1 and u2);
```

henrikt-ma (changes requested):
The documentation needs to be written more carefully, specially due to the weird semantics of `allTrue`.

Modelica/Blocks/Logical.mo
HansOlsson: Since `andTrue` (which is used) is the normal implementation of `and` for an array and has the description: 

> Returns true if all elements of the Boolean input vector b are true.

I don't see a problem with the previous formulation.

Modelica/Blocks/Logical.mo
HansOlsson: I don't see that there has to be at non-zero number of inputs.

HansOlsson:
> The documentation needs to be written more carefully, specially due to the weird semantics of `allTrue`.

But it has been changed to use andTrue instead as far as I can tell.

Modelica/Blocks/Logical.mo
henrikt-ma: No, the whole problem is that this freaking block uses `allTrue` instead of `andTrue`.

Modelica/Blocks/Logical.mo
henrikt-ma: We have `u1` and `u2`, which adds up to two inputs, no more, no less.

henrikt-ma:
> > The documentation needs to be written more carefully, specially due to the weird semantics of `allTrue`.
> 
> But it has been changed to use andTrue instead as far as I can tell.

No, `allTrue` is used in `MultiAnd`.

Modelica/Blocks/Logical.mo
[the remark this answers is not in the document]
henrikt-ma: I'm sorry, my bad.  It is `MultiNand` that still uses `allTrue`:
```
  block MultiNand "Logical nand of Boolean vector elements"
    extends Modelica.Blocks.Interfaces.PartialBooleanMISO;
  equation
    y=not Modelica.Math.BooleanVectors.allTrue(u);
```

henrikt-ma: Sorry about the confusion.  I'll close this and move the suggestion to `MultiNand` where it belongs.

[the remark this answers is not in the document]
HansOlsson: So, shouldn't we fix that so that Nand is the same as Not And?

Modelica/Blocks/Logical.mo
```diff
@@ -38,20 +52,50 @@ the output is <strong>false</strong>.
...
+    extends Modelica.Blocks.Interfaces.PartialBooleanMISO;
+  equation
+    y=not Modelica.Math.BooleanVectors.allTrue(u);
+    annotation (Icon(coordinateSystem(initialScale=0.1), graphics={
+                 Text(extent={{-80,30},{60,-30}},
+            textString="nand",
+            textColor={0,0,0})}),
+      Documentation(info="<html>
+<p>
+The output is <strong>true</strong> if at least one element of the input vector is <strong>false</strong>, otherwise
```
henrikt-ma, line 94: This is where we need to deal with the use of `allTrue`:
```suggestion
The output is <strong>true</strong> if the input vector is empty or at least one element of the input vector is <strong>false</strong>, otherwise
```

Modelica/Blocks/Logical.mo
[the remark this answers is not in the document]
henrikt-ma: I'd love to see that kind of change, but can we do it this late in the release cycle considering that it is a bugfix on the border to a plain backwards incompatible change?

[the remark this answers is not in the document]
HansOlsson: > I'd love to see that kind of change, but can we do it this late in the release cycle considering that it is a bugfix on the border to a plain backwards incompatible change?

The models are new so we don't have to care about backwards compatibility, and to me it looks like MultiAnd was corrected, but not MultiNand.

Modelica/Blocks/Logical.mo
```diff
@@ -38,20 +52,50 @@ the output is <strong>false</strong>.
...
 the output is <strong>false</strong>.
 </p>
 </html>"));
-  end Xor;
+  end Nand;
+
+  block MultiNand "Logical nand of Boolean vector elements"
+    extends Modelica.Blocks.Interfaces.PartialBooleanMISO;
+  equation
+    y=not Modelica.Math.BooleanVectors.allTrue(u);
```
HansOlsson, line 87:
```suggestion
    y=not Modelica.Math.BooleanVectors.andTrue(u);
```
Change this as well to use `andTrue` - so that Nand=Not-And.

Modelica/Blocks/Logical.mo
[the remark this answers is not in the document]
henrikt-ma: @AHaumer, could you please consider fixing `MultiNand` as part of this PR, so that we don't have to handle the annoying difference between `allTrue` and `andTrue` in the documentation?

Modelica/Blocks/Logical.mo
henrikt-ma: @AHaumer, could you please consider including this change?

PR #3916, opened 2021-12-13
Implemented and tested Triggered{Mean, RectifiedMean, RootMeanSquare}

AHaumer:
I noticed that the blocks {Mean, RectifiedMean, RootMeanSquare} are extremely usefull to inspect the results of power electronics simulations, but they depend on constant frequency. Therefore I created a triggered version of these blocks. 
Mathematically it is not correct to talk about these characteristc values, if the period of the input signal is not constant.
So the values during the periods of time with varying frequency are more or less an estimaton, but during peridos of time with constant frequency they are 100% correct. 
The blocks react on the positive edge of the Boolean trigger signal. For instance, in applications with synchronous machines the trigger could be deduced from the angle of rotation - once per revolution or once per polepair.


Modelica/Blocks/Math.mo
```diff
@@ -2264,37 +2264,64 @@ explicitly set to 0.0, if the mean value results in a negative value.
...
-Note: The output is updated after each period defined by 1/f.
+Note: The output is updated at each trigger instance.
+</p>
+
+<p>
+If parameter <strong>yGreaterOrEqualZero</strong> in the Advanced tab is <strong>true</strong> (default = <strong>false</strong>),
+then the modeller provides the information that the mean of the input signal is guaranteed
+to be &ge; 0 for the exact solution. However, due to inaccuracies in the numerical integration scheme,
+the output might be slightly negative. If this parameter is set to true, then the output is
+explicitly set to 0.0, if the mean value results in a negative value.
```
christiankral, line 2319: I am afraid I do not understand the reason for the implementation of `yGreaterOrEqualZero`:
- Is it a side effect that has to be suppressed or is this just a "normal" integration phenomenon?
- It possibly shall be explained under which circumstances a slightly negative output occurs
- What happens if -- in the above case -- the negative input is fed to the block in a simulation model: 
    - Is then the output slightly positive (instead of negative)? 
    - Shall we also implement countermeasure for this case?  

### Simulation results

- Possibly some analytical reference signals shall be provided and compared to the triggered block calculations.
- We shall then also explain in the documentation what signals shall be compared to each other.

AHaumer: This was Introduced some time ago (I'm not sure who did that). The reason is simple:
The result of mean could be slightly negative due to numerical reasons, although the analytical result is zero. 
This is annoying since Mean is used in RootMeanSquare, and RMS can only be greater or euqal to zero.

christiankral: > ### Simulation results
> * Possibly some analytical reference signals shall be provided and compared to the triggered block calculations.
> * We shall then also explain in the documentation what signals shall be compared to each other.

@AHaumer this shall be provided to the users to better understand the purpose of the example(s)

PR #3920, opened 2021-12-16
Liberate machine parameter m

AHaumer:
I recently noticed that setting the parameters 
`Modelica.Electrical.Machines.Utilities.ParameterRecords.InductionMachineData.m` and 
`Modelica.Electrical.Machines.Utilities.ParameterRecords.InductionMachineData.statorCoreParameters.m` 
final is too restrictive. This prevents proper usage for polyphase machine models from `Modelica.Magnetic.FundamentalWave`.
This does'nt change the behaviour or the result of examples and user models, it just enables user to implement parameter records for real machines with arbitrary number of phases.

### DIFF
--- a/Modelica/Electrical/Machines/Utilities/ParameterRecords/InductionMachineData.mo
+++ b/Modelica/Electrical/Machines/Utilities/ParameterRecords/InductionMachineData.mo
@@ -30,7 +30,7 @@ record InductionMachineData "Common parameters for induction machines"
         2*pi*fsNominal/p) "Friction loss parameter record"
     annotation (Dialog(tab="Losses"));
   parameter Machines.Losses.CoreParameters statorCoreParameters(
-    m=m,
+    final m=m,
     PRef=0,
     VRef=100,
     wRef=2*pi*fsNominal)
--- a/Modelica/Electrical/Machines/Utilities/ParameterRecords/InductionMachineData.mo
+++ b/Modelica/Electrical/Machines/Utilities/ParameterRecords/InductionMachineData.mo
@@ -2,7 +2,7 @@ within Modelica.Electrical.Machines.Utilities.ParameterRecords;
 record InductionMachineData "Common parameters for induction machines"
   extends Modelica.Icons.Record;
   import Modelica.Constants.pi;
-  final parameter Integer m=3 "Number of phases" annotation(Evaluate=true);
+  parameter Integer m=3 "Number of phases" annotation(Evaluate=true);
   parameter SI.Inertia Jr=0.29 "Rotor's moment of inertia";
   parameter SI.Inertia Js=Jr "Stator's moment of inertia";
   parameter Integer p(min=1) = 2 "Number of pole pairs (Integer)";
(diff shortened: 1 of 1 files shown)


Modelica/Electrical/Machines/Utilities/ParameterRecords/InductionMachineData.mo
```diff
@@ -30,7 +30,7 @@ record InductionMachineData "Common parameters for induction machines"
         2*pi*fsNominal/p) "Friction loss parameter record"
     annotation (Dialog(tab="Losses"));
   parameter Machines.Losses.CoreParameters statorCoreParameters(
-    final m=m,
+    m=m,
```
henrikt-ma, line 33: I am not very familiar with these electrical machine models, but what I can see from my myopic perspective looks a bit surprising.  Could you please just explain why it would make sense to have a different `m` in the `statorCoreParameters` member than in the rest of the `InductionMachineData` record?

AHaumer: Well we want to keep the parameters for core losses together, and the CoreParameter record is used for 
- DC machines (m=1)
- 3phase machines (Modelica.Electrical.Machines m=3)
- PolyPhase machines (Modelica.Magnetic.FundamentalWave, m>=3)
So we are instantiating the CoreParameter record in different machine parameter records, propagating the desired number of phases m. How should the CoreParameter record know which number of phases is used in the machine parameter record (one layer higher in hierarchy)? Or do I miss something?

henrikt-ma: OK, I'll rephrase the question:  Why was `final` removed here?  Does any of your examples involve a `InductionMachineData` where the `m` differs from the `m` inside its own `statorCoreParameters`?

AHaumer: Ok I understand - and you're right:
As long as I propagate final m=m from within the enclosing parameter record there's no need to give it free to change (and it is even better to restrict it). Reverted that.

christiankral (approved):
Good. This change makes sense.

henrikt-ma (commented):
I don't see any obvious problem with making the `m` in `InductionMachineData` non-final, but on the other hand I have no idea what other parts of the MSL that could rely on this `m` being 3 (which was a sensible assumption to make in the past when it was final).

christiankral:
The change should be good as 
- `Electrical.Machines` is limited to `m = 3` and 
- `Magnetic.FundamentalWave` can handle machine models with phase numbers other than three.

The parameter `m` defined in the record `InductionMachineData` is used within the instance `statorCoreParameters` of the parameter record `Machines.Losses.CoreParameters`: There, in the core parameter record, `m` is used in

```modelica
  final parameter SI.Conductance GcRef=if (PRef <= 0) then 0
       else PRef/VRef^2/m
 ```

This implementation shall be OK for any symmetric machine.

dietmarw (approved):
Looks good to me.

PR #3924, opened 2021-12-22
Fix the bad Evaluate annotation

adrpo:
Fix annoatation: 'Evaluate' -> 'Evaluate=true'


casella (approved):
LGTM. Who know how we got that.

beutlich:
> LGTM. Who know how we got that.

It has been ever like this since Fluid was added in 2009.

beutlich (approved):

PR #3939, opened 2022-01-21
Removes commented Modelica URI link in documentation with list

### LINKED ISSUE #3937: Dead Modelica URI link in documentation
A dae link like this causes problems for those that try to keep their documentation free of dead links:
https://github.com/modelica/ModelicaStandardLibrary/blob/8345852d3b7d8af752cd4960e4dff5a935ba8f3b/Modelica/Electrical/Machines/Thermal/package.mo#L140

This is the only case I'm aware of.  Any chance we could avoid pointing to non-existing classes like this?

dietmarw:
Fixes #3937
Documentation should state what **is** implemented and not point to something that *might* get implemented. Here this resulted also in a list without heading.


henrikt-ma (commented):
I have no objections.

christiankral (approved):
This makes sense

hubertus65 (approved):
Looks good to me.

PR #3948, opened 2022-02-10
Change y into x in spatialDistribution() definitions

beutlich:
Consider upstream fix from MO#3113 by @casella .


HansOlsson (approved):
Ok

Issue #3963, opened 2022-03-25
Parameter T not considered in Polyphase Resistors and Conductors


m-kessler:
In the following models the parameter `T` is of no use, as it is not propagated to the resistor and conductor instances:

```
Modelica.Electrical.Polyphase.Basic.Resistor
Modelica.Electrical.Polyphase.Basic.Conductor
Modelica.Electrical.Polyphase.Basic.VariableResistor
Modelica.Electrical.Polyphase.Basic.VariableConductor
```

To reproduce, simulate this:

```
model PolyphaseResistor
  Modelica.Electrical.Polyphase.Basic.Resistor resistor(
    R=fill(1, 3), T_ref=fill(300, 3), alpha=fill(1, 3), T=fill(400, 3));
  Modelica.Electrical.Polyphase.Basic.Star star;
  Modelica.Electrical.Analog.Basic.Ground ground;
equation 
  connect(resistor.plug_n, star.plug_p);
  connect(star.pin_n, ground.p);
end PolyphaseResistor;
```
Then plot `resistor.resistor[1].R_actual`. It should be 101Ω, but it is 1Ω.

AHaumer:
You are right, I'll create a PR within short

PR #3964, opened 2022-03-25
Propagate parameter T in Polyphase Resistors and Conductors

AHaumer:
fix issue #3963
Thanks to @m-kessler for pointing that out.
Don't how this coudl happen, and why nobody stumbled into that up to now.
Do we have to back-port this bugfix to previous versions? 
@dietmarw could you please review and comment? Thanks in advance!


dietmarw (approved):
I've cleaned up the white spaces and dropped the one dummy commit "???".
@AHaumer As for the reason why this was not found earlier, I guess the models are simply missing some test models for exactly that temperature behaviour. Having those would have caught this.

PR #3965, opened 2022-03-28
Fix parameter propagation of IntersectivePWM

dietmarw:
* Fix not propagated parameter

* Use final attribute for parameter propagation

Backport #3910


christiankral (approved):
Looks good.

PR #3972, opened 2022-04-08
Add problem matchers for GitHub Actions (CI)

beutlich:


dietmarw (commented):
Could you explain a bit more about the background of this change?
Also the unrelated change in ModeliaReference (though trivial) should probably not be part of this PR, right?

beutlich:
> Could you explain a bit more about the background of this change?

Problem Matchers are a way to scan the output of actions for a specified regex pattern and surface that information prominently in the UI. Currently, failing CI tasks are only reported as part of the Actions tab in the PR. With this change, the failures are directly reported on the files. For example, see `https://github.com/conan-io/conan-center-index/pull/10227/files`

image

> Also the unrelated change in ModeliaReference (though trivial) should probably not be part of this PR, right?

OK.

beutlich:
@sjoelund Can you please review! Thanks!

beutlich:
> Also the unrelated change in ModeliaReference (though trivial) should probably not be part of this PR, right?

Done. There now is #3980 for that trivial correction.

dietmarw (approved):

PR #3976, opened 2022-04-19
Eddy current sources

AHaumer:
Just a small backwards compatible improvement of already existing components: 
Modelica.Mechanics.{Rotational, Translational}.Sources.EddyCurrent{Torque, Force} would be much more versatile if they would enable (optional) influence of excitation.


dietmarw (changes requested):
Looks good but the unrelated (dummy) change should be removed from `DemonstrateLightning.mo` for this PR.

dietmarw (changes requested):
Still contains changes to `DemonstrateLightning.mo` unfortunately.

AHaumer:
> Still contains changes to `DemonstrateLightning.mo` unfortunately.

Could you please point me to the changes, or just revert them? I donÄt see more than two empty lines that I removed.

dietmarw (approved):
Looks good now.

beutlich:
Please squash commit to a single one before merge!

tobolar (commented):
@AHaumer Very nice implementation! 
Minor comments:
1. To my understanding, `excitation` shall be between 0 and 1. If true, this shall be documented as well. Shall also ´excitation` be limited to these values in the eddy current brakes?
2. Maybe the equations for `tau` and `f` can be given in the documentation as well - for better understanding.

AHaumer:
> @AHaumer Very nice implementation! Minor comments:
> 
>     1. To my understanding, `excitation` shall be between 0 and 1. If true, this shall be documented as well. Shall also ´excitation` be limited to these values in the eddy current brakes?
> 
>     2. Maybe the equations for `tau` and `f` can be given in the documentation as well - for better understanding.
> 
> 
> Note: Changes in examples will trigger discrepances in regression tests.

You are right, negative excitation doen't make sense.
Excitation greater 1 does make sense, at least for a short period of time we could over-excite.
I'll consider to include the equations fpr tau and f in the documentation.
Bear in mind that the components were already implemented in MSL, this is just am enhancement with variable excitation.

AHaumer:
One could even provide an auxilliary model consisting of resistance and inductance of the excitation winding, providing the excitaton signal (i.e. the magnetic flux of the inductance). This way one could model control of the excitation current as wel as the thermal analysis of the excitaton winding in a realistic way.

beutlich:
> Please squash commits to a single one before merge!

I enabled auto-merge with squash to merge the changes as a single commit.

christiankral (approved):
On my opinion negative currents shall be allowed, as the sign of the current does not affect the generated torque. Additionally, overloading the eddy current source shall be possible. So I am in favor of not limiting the range of the current input.

PR #3979, opened 2022-05-01
Link checker: Fix 403 error (CI)

beutlich:
1. This fixes the HTTPError 403 reported by https://github.com/modelica/ModelicaStandardLibrary/runs/6244261298?check_suite_focus=true
2. It also fixes an UnicodeDecodeError when run locally via Python 3.7.3. I used the codecs.open function to not break compatibility with Python 2.7.


beutlich:
That 403 error was gone by latest scheduled run this morning.

Issue #2108, opened 2017-01-15
MSL uses ModelicaReference


modelica-trac-importer:
**Reported by beutlich on 11 Nov 2016 13:30 UTC**
Should the MSL have an uses annotation on ModelicaReference since there are modelica URIs inside MSL referring to it?

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/2108

dietmarw:
Question is if the URI dependency in the documentation is enough to have a `uses` clause. So you could argue that technically the MSL does not depend on `Modelica Reference` to function the documentation might be targeting some dead links. This is currently not an issue since `Modelica Reference` is shipped side by side with `Modelica`. So I like to postpone this until we have a modular version of `Modelica Reference`.

beutlich:
Not a relevant issue.

PR #3985, opened 2022-05-11
Added assert condition block

mestinso:
- Block is useful for applying asserts via graphical layer (rather than via equations)

### DIFF
--- a/Modelica/Blocks/Logical.mo
+++ b/Modelica/Blocks/Logical.mo
@@ -986,7 +986,7 @@ parameter \"terminationText\".
             extent={{-90,40},{90,-40}},
             textColor={0,0,0},
             textString="assert")}), Documentation(info="<html>
-<p>This block is used by connecting a (possibly) <strong>time varying</strong> input condition for the Boolean input variable <strong>u</strong>. If this input variable is <strong>false</strong>, the simulation will either abort with a message indicating the cause of the error (if the assertion level is set to error) or not abort with a message indicating the cause of the warning (if the assertion level is set to warning).</p>
+<p>This block is used by connecting a (possibly) <strong>time varying</strong> input condition for the Boolean input variable <strong>u</strong>. If this input variable is <strong>false</strong>, the simulation will either abort with an error (if <strong>assertionLevel</strong> is set to <strong>error</strong>), or issue a warning and continue (if <strong>assertionLevel</strong> is set to <strong>warning</strong>). The <strong>message</strong> will be included in the message text of the error or warning.</p>
 </html>"));
   end AssertCondition;
   annotation (Documentation(info="<html>
--- a/Modelica/Blocks/Logical.mo
+++ b/Modelica/Blocks/Logical.mo
@@ -953,7 +953,7 @@ either abort with an error (if <strong>assertionLevel</strong> is set to
 The <strong>message</strong> will be reported in either case.
 Typically, this block is used by connecting a&nbsp;(possibly) <strong>time
 varying</strong> outside condition and the input&nbsp;<strong>u</strong>
-to stop the simulation if the condition fails.
+to abort the simulation if the condition fails.
 
 </p>
 <p>
--- a/Modelica/Blocks/Logical.mo
+++ b/Modelica/Blocks/Logical.mo
@@ -929,7 +929,7 @@ The output <code>Q</code> is set by the input <code>S</code>, is reset by the in
 </html>"));
   end RSFlipFlop;
 
-  block AssertCondition "Asserts that input condition u is true"
+  block AssertCondition "Assert that input u is true"
     extends Modelica.Blocks.Interfaces.partialBooleanSI;
     parameter String message = getInstanceName() + "'s input is false"
       "Assertion message to be displayed if the input u is false";
@@ -945,14 +945,16 @@ The output <code>Q</code> is set by the input <code>S</code>, is reset by the in
       Documentation(
         info="<html>
 <p>
-This block is used by connecting a&nbsp;(possibly) <strong>time varying</strong>
-input condition for the Boolean input variable&nbsp;<strong>u</strong>.
-If this input variable is <strong>false</strong>, the simulation will
+This block asserts that the Boolean input&nbsp;<strong>u</strong> is <strong>true</strong>.
+If&nbsp;<strong>u</strong> is <strong>false</strong>, the simulation will
 either abort with an error (if <strong>assertionLevel</strong> is set to
 <strong>error</strong>), or issue a&nbsp;warning and continue (if
 <strong>assertionLevel</strong> is set to <strong>warning</strong>).
-The <strong>message</strong> will be included in the message text of
-the error or warning.
+The <strong>message</strong> will be reported in either case.
+Typically, this block is used by connecting a&nbsp;(possibly) <strong>time
+varying</strong> outside condition and the input&nbsp;<strong>u</strong>
+to terminate the simulation if the condition fails.
+
 </p>
 <p>
 For further information, refer to the documentation of
--- a/Modelica/Blocks/Logical.mo
+++ b/Modelica/Blocks/Logical.mo
@@ -953,7 +953,7 @@ either abort with an error (if <strong>assertionLevel</strong> is set to
 The <strong>message</strong> will be reported in either case.
 Typically, this block is used by connecting a&nbsp;(possibly) <strong>time
 varying</strong> outside condition and the input&nbsp;<strong>u</strong>
-to terminate the simulation if the condition fails.
+to stop the simulation if the condition fails.
 
 </p>
 <p>
--- a/Modelica/Blocks/Logical.mo
+++ b/Modelica/Blocks/Logical.mo
@@ -933,7 +933,7 @@ The output <code>Q</code> is set by the input <code>S</code>, is reset by the in
     extends Modelica.Blocks.Interfaces.partialBooleanSI;
     parameter String message = getInstanceName() + "'s input is false"
       "Assertion message to be displayed if the input u is false";
-    parameter AssertionLevel assertionLevel=AssertionLevel.error
+    parameter AssertionLevel assertionLevel = AssertionLevel.warning
       "Assertion level";
   equation
     assert(u,message,assertionLevel);
--- a/Modelica/Blocks/Logical.mo
+++ b/Modelica/Blocks/Logical.mo
@@ -977,7 +977,7 @@ parameter \"terminationText\".
   block AssertCondition "Asserts that input condition u is true"
     extends Modelica.Blocks.Interfaces.partialBooleanSI;
     parameter String message=getInstanceName() + "'s input u is false"
-      "Assertion message to be displayed if input u is false";
+      "Assertion message to be displayed if the input u is false";
     parameter AssertionLevel assertionLevel=AssertionLevel.error
       "Assertion level";
   equation
--- a/Modelica/Blocks/Logical.mo
+++ b/Modelica/Blocks/Logical.mo
@@ -951,15 +951,14 @@ either abort with an error (if <strong>assertionLevel</strong> is set to
 <strong>error</strong>), or issue a&nbsp;warning and continue (if
 <strong>assertionLevel</strong> is set to <strong>warning</strong>).
 The <strong>message</strong> will be reported in either case.
-Typically, this block is used by connecting a&nbsp;(possibly) <strong>time
-varying</strong> outside condition and the input&nbsp;<strong>u</strong>
-to abort the simulation if the condition fails.
-
-</p>
-<p>
 For further information, refer to the documentation of
 <a href=\"modelica://ModelicaReference.Operators.'assert()'\">ModelicaReference.Operators.'assert()'</a>
 &ndash; an operator which this block is based on.
+</p>
+<p>
+Typically, this block is used by connecting a&nbsp;(possibly) <strong>time
+varying</strong> outside condition and the input&nbsp;<strong>u</strong>
+to abort the simulation if the condition fails.
 </p>
 </html>"));
   end AssertCondition;
--- a/Modelica/Blocks/Logical.mo
+++ b/Modelica/Blocks/Logical.mo
@@ -929,6 +929,34 @@ The output <code>Q</code> is set by the input <code>S</code>, is reset by the in
 </html>"));
   end RSFlipFlop;
 
+  block AssertCondition "Asserts that input condition u is true"
+    extends Modelica.Blocks.Interfaces.partialBooleanSI;
+    parameter String message = getInstanceName() + "'s input is false"
+      "Assertion message to be displayed if the input u is false";
+    parameter AssertionLevel assertionLevel=AssertionLevel.error
+      "Assertion level";
+  equation
+    assert(u,message,assertionLevel);
+    annotation (
+      Icon(graphics={
+          Text(
+            extent={{-90,40},{90,-40}},
+            textString="assert")}),
+      Documentation(
+        info="<html>
+<p>
+This block is used by connecting a&nbsp;(possibly) <strong>time varying</strong>
+input condition for the Boolean input variable&nbsp;<strong>u</strong>.
+If this input variable is <strong>false</strong>, the simulation will
+either abort with an error (if <strong>assertionLevel</strong> is set to
+<strong>error</strong>), or issue a&nbsp;warning and continue (if
+<strong>assertionLevel</strong> is set to <strong>warning</strong>).
+The <strong>message</strong> will be included in the message text of
+the error or warning.
+</p>
+</html>"));
+  end AssertCondition;
+
   block TerminateSimulation "Terminate simulation if condition is fulfilled"
 
     Modelica.Blocks.Interfaces.BooleanOutput condition=false
@@ -974,33 +1002,6 @@ parameter \"terminationText\".
 </html>"));
   end TerminateSimulation;
 
-  block AssertCondition "Asserts that input condition u is true"
-    extends Modelica.Blocks.Interfaces.partialBooleanSI;
-    parameter String message = getInstanceName() + "'s input is false"
-      "Assertion message to be displayed if the input u is false";
-    parameter AssertionLevel assertionLevel=AssertionLevel.error
-      "Assertion level";
-  equation
-    assert(u,message,assertionLevel);
-    annotation (
-      Icon(graphics={
-          Text(
-            extent={{-90,40},{90,-40}},
-            textString="assert")}),
-      Documentation(
-        info="<html>
-<p>
-This block is used by connecting a&nbsp;(possibly) <strong>time varying</strong>
-input condition for the Boolean input variable&nbsp;<strong>u</strong>.
-If this input variable is <strong>false</strong>, the simulation will
-either abort with an error (if <strong>assertionLevel</strong> is set to
-<strong>error</strong>), or issue a&nbsp;warning and continue (if
-<strong>assertionLevel</strong> is set to <strong>warning</strong>).
-The <strong>message</strong> will be included in the message text of
-the error or warning.
-</p>
-</html>"));
-  end AssertCondition;
   annotation (Documentation(info="<html>
 <p>
 This package provides blocks with Boolean input and output signals
--- a/Modelica/Blocks/Logical.mo
+++ b/Modelica/Blocks/Logical.mo
@@ -954,6 +954,9 @@ either abort with an error (if <strong>assertionLevel</strong> is set to
 The <strong>message</strong> will be included in the message text of
 the error or warning.
 </p>
+For further information, refer to the documantation of
+<a href=\"modelica://ModelicaReference.Operators.'assert()'\">ModelicaReference.Operators.'assert()&apos;</a>
+&ndash; an operator which this block is based on.
 </html>"));
   end AssertCondition;
 
--- a/Modelica/Blocks/Logical.mo
+++ b/Modelica/Blocks/Logical.mo
@@ -954,9 +954,11 @@ either abort with an error (if <strong>assertionLevel</strong> is set to
 The <strong>message</strong> will be included in the message text of
 the error or warning.
 </p>
-For further information, refer to the documantation of
-<a href=\"modelica://ModelicaReference.Operators.'assert()'\">ModelicaReference.Operators.'assert()&apos;</a>
+<p>
+For further information, refer to the documentation of
+<a href=\"modelica://ModelicaReference.Operators.'assert()'\">ModelicaReference.Operators.'assert()'</a>
 &ndash; an operator which this block is based on.
+</p>
 </html>"));
   end AssertCondition;
 
--- a/Modelica/Blocks/Logical.mo
+++ b/Modelica/Blocks/Logical.mo
@@ -976,17 +976,29 @@ parameter \"terminationText\".
 
   block AssertCondition "Asserts that input condition u is true"
     extends Modelica.Blocks.Interfaces.partialBooleanSI;
-    parameter String message=getInstanceName() + "'s input is false"
+    parameter String message = getInstanceName() + "'s input is false"
       "Assertion message to be displayed if the input u is false";
     parameter AssertionLevel assertionLevel=AssertionLevel.error
       "Assertion level";
   equation
     assert(u,message,assertionLevel);
-    annotation (Icon(graphics={Text(
+    annotation (
+      Icon(graphics={
+          Text(
             extent={{-90,40},{90,-40}},
-            textColor={0,0,0},
-            textString="assert")}), Documentation(info="<html>
-<p>This block is used by connecting a (possibly) <strong>time varying</strong> input condition for the Boolean input variable <strong>u</strong>. If this input variable is <strong>false</strong>, the simulation will either abort with an error (if <strong>assertionLevel</strong> is set to <strong>error</strong>), or issue a warning and continue (if <strong>assertionLevel</strong> is set to <strong>warning</strong>). The <strong>message</strong> will be included in the message text of the error or warning.</p>
+            textString="assert")}),
+      Documentation(
+        info="<html>
+<p>
+This block is used by connecting a&nbsp;(possibly) <strong>time varying</strong>
+input condition for the Boolean input variable&nbsp;<strong>u</strong>.
+If this input variable is <strong>false</strong>, the simulation will
+either abort with an error (if <strong>assertionLevel</strong> is set to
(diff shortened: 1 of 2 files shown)


Modelica/Blocks/Logical.mo
```diff
@@ -973,6 +973,41 @@ parameter \"terminationText\".
 
 </html>"));
   end TerminateSimulation;
+
+  block AssertCondition "Trigger error (or warning) and print message if assertion condition is not fulfilled"
+
+    Modelica.Blocks.Interfaces.BooleanOutput condition=true
```
henrikt-ma, line 979: Can anyone explain why this (and the same in `TerminateSimulation`) is an output and not an input? I mean, what's the point of having an output that can only have the value is `true` (or stays `false` in the case of `TerminateSimulation`). Sure, with `AssertionLevel.warning` it can change back and forth, but it would seem more useful to me if the blocks could be controlled via an input connector.

I suppose one could also arrange for a conditional input connector, allowing the same block to be controlled by either an expression or an input?

Modelica/Blocks/Logical.mo
```diff
@@ -973,6 +973,41 @@ parameter \"terminationText\".
...
+      "Trigger error (or warning) and print message when condition becomes false"
+      annotation (Dialog, Placement(transformation(extent={{200,-10},{220,10}})));
+    parameter String message="... assertion condition no longer true"
+      "Text that will be displayed when assertion condition is false";
+    parameter AssertionLevel assertionLevel=AssertionLevel.error
+      "Select warning (1) or error (2)";
```
henrikt-ma, line 985: What are the numbers in parentheses supposed to mean? Wouldn't this be more to the point?
```suggestion
      "Assertion level";
```

Modelica/Blocks/Logical.mo
```diff
@@ -973,6 +973,41 @@ parameter \"terminationText\".
 
 </html>"));
   end TerminateSimulation;
+
+  block AssertCondition "Trigger error (or warning) and print message if assertion condition is not fulfilled"
+
+    Modelica.Blocks.Interfaces.BooleanOutput condition=true
+      "Trigger error (or warning) and print message when condition becomes false"
+      annotation (Dialog, Placement(transformation(extent={{200,-10},{220,10}})));
```
henrikt-ma, line 981: I see that it comes from `TerminateSimulation`, but what is the purpose of the argument-less `Dialog` annotation?

Modelica/Blocks/Logical.mo
```diff
@@ -973,6 +973,41 @@ parameter \"terminationText\".
...
+          Rectangle(
+            fillColor={0,140,72},
+            fillPattern=FillPattern.Solid,
+            borderPattern=BorderPattern.Raised,
+            extent={{-194,-14},{-168,14}}),
+          Text(
+            textColor={0,0,255},
+            extent={{-200,22},{200,46}},
+            textString="%name")}), Documentation(info="<html>
+<p>In the parameter menu, a <strong>time varying</strong> expression can be defined via variable <strong>condition</strong>, for example &quot;condition = x &lt; 0&quot;, where &quot;x&quot; is a variable that is declared in the model in which the &quot;AssertCondition&quot; block is present. If this expression becomes <strong>false</strong>, the simulation will either abort with a message indicating the cause of the error (if assertion level is set to error) or not abort with a message indicating the cause of the warning (if assertion level is set to warning).</p>
```
henrikt-ma, line 1008: Can we avoid speaking of the _parameter menu_ for a more tool independent documentation?

Modelica/Blocks/Logical.mo
mestinso: > What are the numbers in parentheses supposed to mean? Wouldn't this be more to the point?

Assertion level is an enumeration, so after compilation in an FMU for example, the integer 1 corresponds to warning and 2 corresponds to error. I could go either way, but in some of my personal usage of enumeration parameters, I found this useful. Open to suggestions.

Modelica/Blocks/Logical.mo
mestinso: Without the **Dialog** annotation, the user is not able to specify **condition** via the parameterization/dialog window (at least in Dymola).

Modelica/Blocks/Logical.mo
mestinso: Any suggestions? This wording was pulled from the **TerminateSimulation** block.

henrikt-ma: Maybe it can just be omitted:
```suggestion
<p>The block is used by specifying a (possibly) <strong>time varying</strong> expression for the variable <strong>condition</strong>, for example &quot;condition = x &lt; 0&quot;, where &quot;x&quot; is a variable that is declared in the model in which the &quot;AssertCondition&quot; block is present. If this expression becomes <strong>false</strong>, the simulation will either abort with a message indicating the cause of the error (if assertion level is set to error) or not abort with a message indicating the cause of the warning (if assertion level is set to warning).</p>
```

Modelica/Blocks/Logical.mo
mestinso: While it may not fully answer the question, I'll give you my comments.

I expect I would rarely, if ever, use the external output connection as is. Maybe there are some unique use cases possible when using the warning assertion level as you hinted. With that said, I added it for two reasons: a) to be in alignment with the TerminateSimulation block, and b) because why not, it doesn't hurt and it may have some use cases that are helpful that I'm not aware of.

Alternatively, if the block had an input connector, I would say 99% of the time I would always just have a "BooleanExpression" source block immediately upstream, in which case the design as-is is preferred for simplicity and ease of use.

Another comment is that, to me, the normal usage of this block should feel comparable to writing an assert statement in the equation section, but the main difference is that I've done it in the graphical layer rather than in the equation section.

henrikt-ma: Sure, I can also imagine that the more common use case would be to write an expression, and the question of how to control the block reminds me of some recent mention of the addition variant of a `Gain` block: we currently have `Add`, `Procuct`, and `Gain`, but no `Shift` (or `Offset`, etc) block. Just as you say regarding the use of `AssertCondition`, it is quite inconvenient to always have to use the combination of `Constant` and `Add` in order to shift a signal.  The situation is not identical, as the `Gain` has a parameter that allows the factor to be modified at initialization, but I still think it illustrates the desire to have alternatives for both the input and declaration equation variants.

Hence, I agree that only supporting the use of an input connector would be poor design. This is what I have in mind:
```
block AssertCondition "Trigger error (or warning) and print message if assertion condition is not fulfilled"
  parameter Boolean enableInput = false "Enable use of input connector instead of condition expression";
  Modelica.Blocks.Interfaces.BooleanOutput condition = true if not enableInput "Trigger error (or warning) and print message when condition becomes false" annotation(Dialog(enable = not enableInput), Placement(transformation(extent = {{200, -10}, {220, 10}})));
  Modelica.Blocks.Interfaces.BooleanInput u if enableInput "Trigger error (or warning) and print message when input becomes false" annotation(Dialog(enable = enableInput), Placement(transformation(extent = {{200, -10}, {220, 10}}, origin = {0, 0}, rotation = 0), visible = true, iconTransformation(origin = {-200, 0}, extent = {{-20, -10}, {0, 10}}, rotation = 0)));
  parameter String message = "... assertion condition no longer true" "Text that will be displayed when assertion condition is false";
  parameter AssertionLevel assertionLevel = AssertionLevel.error "Select warning (1) or error (2)";
protected
  Modelica.Blocks.Interfaces.BooleanInput cond;
equation
  connect(condition, cond);
  connect(u, cond);
  assert(cond, message, assertionLevel);
  annotation(Icon(coordinateSystem(preserveAspectRatio = true, extent = {{-200, -20}, {200, 20}}, initialScale = 0.2, grid = {10, 10}), graphics = {Rectangle(visible = true, fillColor = {235, 235, 235}, fillPattern = FillPattern.Solid, lineThickness = 5, borderPattern = BorderPattern.Raised, extent = {{-200, -20}, {200, 20}}), Text(visible = not enableInput, origin = {-15.556, 0}, extent = {{-184.444, -15}, {215.556, 15}}, textString = "assert(%condition)"), Text(visible = true, textColor = {0, 0, 255}, extent = {{-200, 22}, {200, 46}}, textString = "%name"), Text(visible = enableInput, origin = {-15.556, 0}, extent = {{-184.444, -15}, {215.556, 15}}, textString = "assert")}), Documentation(info = "<html>
<p>In the parameter menu, a <strong>time varying</strong> expression can be defined via variable <strong>condition</strong>, for example &quot;condition = x &lt; 0&quot;, where &quot;x&quot; is a variable that is declared in the model in which the &quot;AssertCondition&quot; block is present. If this expression becomes <strong>false</strong>, the simulation will either abort with a message indicating the cause of the error (if assertion level is set to error) or not abort with a message indicating the cause of the warning (if assertion level is set to warning).</p>
</html>"));
end AssertCondition;
```

Note that I have also modified the icon to make it more distinguishable from the `TerminateSimulation`, and the contents of the icon are controlled by the `enableInput` parameter.

HansOlsson: To me both this and `TerminateSolution` somewhat break the separation of concerns.

The general idea is to have connections for propagating signals, and we have `Modelica.Blocks.Sources.BooleanExpression` (etc) if we want to write an expression and generate a signal.

Having (optional) expression-code in many blocks lead to less clear solutions. I will comment on one of them.

Modelica/Blocks/Logical.mo
```diff
@@ -973,6 +973,41 @@ parameter \"terminationText\".
...
+      "Trigger error (or warning) and print message when condition becomes false"
+      annotation (Dialog, Placement(transformation(extent={{200,-10},{220,10}})));
+    parameter String message="... assertion condition no longer true"
+      "Text that will be displayed when assertion condition is false";
+    parameter AssertionLevel assertionLevel=AssertionLevel.error
+      "Select warning (1) or error (2)";
+  equation
+    assert(condition,message,assertionLevel);
+    annotation (Icon(coordinateSystem(
+          preserveAspectRatio=true,
```
HansOlsson, line 989: This has the same issue as #3986 and highlights one of the issues of duplicated code.
```suggestion
          preserveAspectRatio=false,
```

HansOlsson:
Investigating origin:

TerminateSimulation was added in https://github.com/modelica/ModelicaStandardLibrary/commit/952520dc9fc8c89c511cbde7bcf4e4b050c7504f (2007-03-25)

And BooleanExpression in https://github.com/modelica/ModelicaStandardLibrary/commit/ee9a1a29568c2815a0e568719f25482a50d388c7 (2004-07-07 - MSL 2.1Beta1 )

Modelica/Blocks/Logical.mo
HansOlsson: Thinking more based on #3986 I believe that there is a clear graphical advantage of using BooleanExpression to generate the signal and then having an assert-block (and terminate-block); even if it at first seems somewhat odd.

The advantage is that we can have a symbol for assert (and terminate) with fix aspect ratio, and a pure expression-block that can be adapted to the size of the expression.

Modelica/Blocks/Logical.mo
HansOlsson: But FMI 1 for ModelExchange and FMI 2 can describe assertion-types for an FMU, so even in that case it shouldn't be needed.

Modelica/Blocks/Logical.mo
henrikt-ma: Compare `BooleanExpression`, where the `Dialog` isn't argument-less:
```
Modelica.Blocks.Interfaces.BooleanOutput y = false "Value of Boolean output" annotation(Dialog(group = "Time varying output signal"), …);
```

mestinso: Understood regarding the group annotation in BooleanExpression (and RealExpression, etc.), but still I can assure you that Dialog without an argument does have functionality in at least Dymola. It's effectively equivalent to doing `annotation(Dialog(group="Parameters"))` as far as I can tell. Though, I suppose with that in mind it could be argued it doesn't belong under a group called "Parameters" since it's not a parameter.

Modelica/Blocks/Logical.mo
mestinso: @henrikt-ma While I understand your suggestion, I personally think it overcomplicates what should be a simple block and the block visual/icon would seem a little out of place when using it with the input connector enabled. With that said, I think I'm coming around to the idea of just having an input connector always enabled and without the time-varying dialog input, as suggested by @HansOlsson. What are your thoughts?

@HansOlsson While I think it adds at least a *little* overhead to have to use two blocks for the typical use case (BooleanExpression + AssertCondition), I am coming around to the idea that maybe that approach is a bit more elegant overall.

Assuming there is an agreement, what is the best way to proceed? It seems to me that this block and `TerminateSimulation` should use the same overall approach. If agreed, it seems one of two options could be taken:
Option 1:
1) Complete this PR mostly as is (i.e. `AssertCondition` similar to `TerminateSimulation`)
2) Later, issue another PR that changes both `AssertCondition` and `TerminateSimulation` to the approach that uses only an input connector

Option 2:
1) Rework this PR to revise `AssertCondition` so that it uses only an input connector
2) Later, issue another PR that brings `TerminateSimulation` into alignment with `AssertCondition`

...this is all assuming there is agreement. Any other thoughts/suggestions?

HansOlsson: I would believe 2 is best. The reason is that when changing TerminateSimulation we need to consider changing all uses - and to avoid breaking compatibility we might even keep both; it's good if we don't have to do that for AssertCondition as well.

henrikt-ma: I don't have strong opinions on this matter.  I just note that I often find it cumbersome to use a `Constant` block when doing things that would have been solved compactly using the `Gain` block in case of multiplication, and that there would be a way to make one block covering both usage patterns.

Anyway, to always control the block via an input at least resolves the confusion about why the controlling expression is attached to an output connector.

Modelica/Blocks/Logical.mo
tobolar: > suggesting to use `annotation(Dialog(group))` rather than `annotation(Dialog)` ...

But not argument-less `group`, of course?!

henrikt-ma: Yes, while waiting for something like `Dialog(show = true)` to become standardized, one needs to make up some other information instead.  As `tab` defaults to `"General"`, perhaps the best alternative is to set `Dialog(tab = "General")`?

henrikt-ma: A problem with `enable` is that it is true by default, so one cannot assume that setting it to that value would have any effect.

Modelica/Blocks/Logical.mo
mestinso: Block has been restructured to use input signal rather than dialog input

Modelica/Blocks/Logical.mo
mestinso: Block has been restructured to use input signal rather than dialog, so this issue is no longer applicable

mestinso:
The block has been restructured based on the previous discussion. Note that the initial scale of the block is 0.04 to be in alignment with the default height of the BooleanExpression block, which this block is expected to be paired with typically. See the picture below (top is the default rather than the bottom):

image

henrikt-ma:
> I'm open to changes, but just wanted to make you aware of why I made the choices I did. Basically they aren't new choices, just following previous examples and choices already made.

OK, thanks for explaining the design.  Matching the style of `FallingEdge` makes sense, and then my personal preference would simply be to also use the same `initialScale`.  By doing so, font sizes will be at least somewhat consistent with other icons such as `FallingEdge`.

> * For the text, I also followed the precedent for all other blocks in Modelica.Blocks.Logical, which had the same text extent: {{-90,40},{90,-40}}. I could drop it down to +- 85 or 80, but it seemed more aligned to just copy the precedent already set. Note that Modelica.Blocks.Logical.FallingEdge has a similarly long word (falling) with the same text extent.

The problem with the `extent`, as I see it, is that text gets scaled to fill the width rather than the height, meaning that one gets all sorts of effective font sizes depending on what goes into the box.  Anyway, I don't think this is the PR where we're supposed to come up with new icon text design guidelines, so copying an existing style is probably the best way to proceed.

mestinso:
Ok, so based on the discussion I went ahead and removed the initial scale annotation. Now by default the scale will match the 2nd example in the picture above. 

I think the PR should be good to go at this point unless any additional concerns.

Modelica/Blocks/Logical.mo
```diff
@@ -973,6 +973,22 @@ parameter \"terminationText\".
 
 </html>"));
   end TerminateSimulation;
+
+  block AssertCondition "Asserts that input condition u is true"
+    extends Modelica.Blocks.Interfaces.partialBooleanSI;
+    parameter String message=getInstanceName() + "'s input u is false"
+      "Assertion message to be displayed if input u is false";
```
henrikt-ma, line 980: How about the more verbose and more grammatically correct?
```suggestion
      "Assertion message to be displayed if the input u is false";
```

Modelica/Blocks/Logical.mo
```diff
@@ -973,6 +973,22 @@ parameter \"terminationText\".
...
+    parameter AssertionLevel assertionLevel=AssertionLevel.error
+      "Assertion level";
+  equation
+    assert(u,message,assertionLevel);
+    annotation (Icon(graphics={Text(
+            extent={{-90,40},{90,-40}},
+            textColor={0,0,0},
+            textString="assert")}), Documentation(info="<html>
+<p>This block is used by connecting a (possibly) <strong>time varying</strong> input condition for the Boolean input variable <strong>u</strong>. If this input variable is <strong>false</strong>, the simulation will either abort with a message indicating the cause of the error (if the assertion level is set to error) or not abort with a message indicating the cause of the warning (if the assertion level is set to warning).</p>
```
henrikt-ma, line 989: Avoid risk of misinterpretation, and an attempt at more consistent use of `<strong>`:
```suggestion
<p>This block is used by connecting a (possibly) <strong>time varying</strong> input condition for the Boolean input variable <strong>u</strong>. If this input variable is <strong>false</strong>, the simulation will either abort with an error (if <strong>assertionLevel</strong> is set to <strong>error</strong>), or issue a warning and continue (if <strong>assertionLevel</strong> is set to <strong>warning</strong>). The <strong>message</strong> will be included in the message text of the error or warning.</p>
```

henrikt-ma (changes requested):
Function and icon seems fine.  Just leaving some suggestions about wording.

tobolar:
@beutlich Is it allowed to link to `ModelicaReference.Operators.'assert()'` in the documentation? This could be helpful since the component fully use it.

tobolar:
I have changed the default of `assertionLevel` paremeter since I believe the component will mostly be used to trigger a warning, not to terminate the simulation completely.

henrikt-ma:
> I have changed the default of `assertionLevel` paremeter since I believe the component will mostly be used to trigger a warning, not to terminate the simulation completely.

And you're sure it wouldn't make more sense to use the same default as the Modelica language `assert`?

tobolar:
@henrikt-ma No, I'm not. :-( I'll revert this.

tobolar:
@beutlich How about https://github.com/modelica/ModelicaStandardLibrary/pull/3985#issuecomment-1203712017 ?

beutlich:
> @beutlich Is it allowed to link to `ModelicaReference.Operators.'assert()'` in the documentation? This could be helpful since the component fully use it.

Why not? See also #2108.

Modelica/Blocks/Logical.mo
```diff
@@ -929,6 +929,34 @@ The output <code>Q</code> is set by the input <code>S</code>, is reset by the in
 </html>"));
   end RSFlipFlop;
 
+  block AssertCondition "Asserts that input condition u is true"
+    extends Modelica.Blocks.Interfaces.partialBooleanSI;
+    parameter String message = getInstanceName() + "'s input is false"
+      "Assertion message to be displayed if the input u is false";
```
HansOlsson, line 935: To me this use of getInstanceName is problematic for two reasons:
* It is redundant and encourages work-arounds for something that should be handled by tools.
* It seems complicated and thus users might want to leave it as is, instead of actually setting the message to something meaningful.

For the first point Dymola currently writes the following if it fails:
```
Error: The following error was detected at time: 0.5
  M2.m.assertCondition's input is false
  Failed condition: m.assertCondition.u
```
So the `getInstanceName()` doesn't add anything.

For the second item I instead modified the test-case to add a Xor connected to the input _and_ output of a not-gate and set the message:
`message="The input and output of Not shall be different"`

That specific test might not be the best, but I believe that something along those lines would better show how it is intended to be used - especially as using asserts in test-cases is a well-known practice in test-driven development.

So I would prefer something simpler:
```suggestion
    parameter String message = "Assert block failed"
      "Assertion message to be displayed if the input u is false; please adapt";
```

HansOlsson (approved):
The general idea is good, but I would like to change the message and the test-case as indicated.

Modelica/Blocks/Logical.mo
henrikt-ma: I am also in favor of not having `getInstanceName()` as part of the message.

(By the way, I guess there is no way that the user can restore the effect of a default `getInstanceName()` with the use of a modification, as any modification is going to be resolved as coming from some other component.)

Modelica/Blocks/Logical.mo
```diff
@@ -945,14 +945,16 @@ The output <code>Q</code> is set by the input <code>S</code>, is reset by the in
...
+If&nbsp;<strong>u</strong> is <strong>false</strong>, the simulation will
 either abort with an error (if <strong>assertionLevel</strong> is set to
 <strong>error</strong>), or issue a&nbsp;warning and continue (if
 <strong>assertionLevel</strong> is set to <strong>warning</strong>).
-The <strong>message</strong> will be included in the message text of
-the error or warning.
+The <strong>message</strong> will be reported in either case.
+Typically, this block is used by connecting a&nbsp;(possibly) <strong>time
+varying</strong> outside condition and the input&nbsp;<strong>u</strong>
+to terminate the simulation if the condition fails.
```
henrikt-ma, line 956: As there is another block for _terminating_ the simulation, shouldn't we avoid using the term _terminate_ in the current context?

Modelica/Blocks/Logical.mo
```diff
@@ -945,14 +945,16 @@ The output <code>Q</code> is set by the input <code>S</code>, is reset by the in
...
-If this input variable is <strong>false</strong>, the simulation will
+This block asserts that the Boolean input&nbsp;<strong>u</strong> is <strong>true</strong>.
+If&nbsp;<strong>u</strong> is <strong>false</strong>, the simulation will
 either abort with an error (if <strong>assertionLevel</strong> is set to
 <strong>error</strong>), or issue a&nbsp;warning and continue (if
 <strong>assertionLevel</strong> is set to <strong>warning</strong>).
-The <strong>message</strong> will be included in the message text of
-the error or warning.
+The <strong>message</strong> will be reported in either case.
+Typically, this block is used by connecting a&nbsp;(possibly) <strong>time
```
henrikt-ma, line 954: How about a paragraph break before normative description and typical usage?

Modelica/Blocks/Logical.mo
HansOlsson: "stop" could work. Short and without the additional meaning.

tobolar: I agree.

henrikt-ma: Or "abort", which has the additional meaning of not terminating successfully.

mestinso: Switched to stop.

Modelica/Blocks/Logical.mo
tobolar: I can even imagine to change the last two sentences, so that "For further information, ..." goes to the first paragraph and "Typically, this block..." is the second paragraph.
@henrikt-ma Would you make a commit?

Modelica/Blocks/Logical.mo
```diff
@@ -929,6 +929,41 @@ The output <code>Q</code> is set by the input <code>S</code>, is reset by the in
...
+<p>
+This block asserts that the Boolean input&nbsp;<strong>u</strong> is <strong>true</strong>.
+If&nbsp;<strong>u</strong> is <strong>false</strong>, the simulation will
+either abort with an error (if <strong>assertionLevel</strong> is set to
+<strong>error</strong>), or issue a&nbsp;warning and continue (if
+<strong>assertionLevel</strong> is set to <strong>warning</strong>).
+The <strong>message</strong> will be reported in either case.
+Typically, this block is used by connecting a&nbsp;(possibly) <strong>time
+varying</strong> outside condition and the input&nbsp;<strong>u</strong>
+to stop the simulation if the condition fails.
```
henrikt-ma, line 956: I think we should simply align with the specification on the wording:
> If the simulation is aborted, message indicates the cause of the error.

```suggestion
to abort the simulation if the condition fails.
```

mestinso:
Made the minor changes as requested (documentation paragraph changes, removal of getInstanceName(), and changing the wording from terminate to abort). Let me know if anything else, otherwise, looking to approve and merge this this here.

Modelica/Blocks/Logical.mo
```diff
@@ -929,6 +929,40 @@ The output <code>Q</code> is set by the input <code>S</code>, is reset by the in
...
+      Documentation(
+        info="<html>
+<p>
+This block asserts that the Boolean input&nbsp;<strong>u</strong> is <strong>true</strong>.
+If&nbsp;<strong>u</strong> is <strong>false</strong>, the simulation will
+either abort with an error (if <strong>assertionLevel</strong> is set to
+<strong>error</strong>), or issue a&nbsp;warning and continue (if
+<strong>assertionLevel</strong> is set to <strong>warning</strong>).
+The <strong>message</strong> will be reported in either case.
+For further information, refer to the documentation of
```
henrikt-ma, line 954: Just noting that the additional paragraph break suggested in https://github.com/modelica/ModelicaStandardLibrary/pull/3985#discussion_r953849225 is not here.  I don't have much of an opinion on whether it is a good idea; this is just an observation.

henrikt-ma (approved):
Looks good to me.

Modelica/Blocks/Logical.mo
mestinso: The "further information" part was added to the first paragraph. The new paragraph starts after that.

tobolar (approved):
Fine to me.

PR #3987, opened 2022-05-23
Set Evaluate = true for IdealTransformer.considerMagnetization

henrikt-ma:
Leaving this parameter unevaluated seems like making things overly complicated, and as far as I can see there is nothing that clearly requires the parameter to be treated structurally.  Hence, adding an explicit annotation will increase chances of models being treated equally efficiently across tools.

Of course, I would have preferred to declare considerMagnetization a constant with Dialog annotation instead, but I'm afraid the general preference is to use a parameter with Evaluate = true instead.

### DIFF
--- a/Modelica/Electrical/Analog/Ideal/IdealTransformer.mo
+++ b/Modelica/Electrical/Analog/Ideal/IdealTransformer.mo
@@ -3,7 +3,7 @@ model IdealTransformer "Ideal transformer core with or without magnetization"
   extends Modelica.Electrical.Analog.Interfaces.TwoPort;
   parameter Real n(start=1) "Turns ratio primary:secondary voltage";
   parameter Boolean considerMagnetization=false
-    "Choice of considering magnetization";
+    "Choice of considering magnetization" annotation(Evaluate = true);
   parameter SI.Inductance Lm1(start=1)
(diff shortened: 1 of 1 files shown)


henrikt-ma:
There's a similar situation in `Modelica.Thermal.FluidHeatFlow.BaseClasses.TwoPort` that would be interesting to discuss at the same time, namely the parameter `m` present in this `if`-equation condition:
```
  if m>Modelica.Constants.small then
    flowPort_a.H_flow + flowPort_b.H_flow + Q_flow = m*medium.cv*der(T);
  else
    flowPort_a.H_flow + flowPort_b.H_flow + Q_flow = 0;
  end if;
```

Would you say it's a Modelica tool "frontend" problem to realize that the condition of this tricky equation should be evaluated, or would you expect that index reduction is the stage when one should determine which parametric `if`-equation conditions to evaluate?

AHaumer (approved):
Fine with wem.
Although I don't get the connection to `Modelica.Thermal.FluidHeatFlow.BaseClasses.TwoPort`.

henrikt-ma:
Nobody has any thoughts regarding the question about `Modelica.Thermal.FluidHeatFlow.BaseClasses.TwoPort` above?

christiankral (approved):
OK with me.

PR #3998, opened 2022-06-20
Propagate the angle alpha also in the calculation of A 

dietmarw:
in HollowCylinderRadialFlux
(backport of #3997)


christiankral (approved):
Looks all good.

PR #3999, opened 2022-06-20
Fix release notes generation

beutlich:
The previously used docverter service for MD to HTML/PDF conversion was shut down and will not come back. This is an alternative implementation with local MD to HTML conversion and utilizing the docraptor service for HTML to PDF conversion.

It also fixes the GitHub Action introduced by #3886.

### DIFF
--- a/.github/workflows/generateReleaseNotes.yml
+++ b/.github/workflows/generateReleaseNotes.yml
@@ -30,7 +30,7 @@ jobs:
         with:
           python-version: 3.8
       - name: Install python packages
-        run: pip install --disable-pip-version-check --user docraptor==3.0.0 Jinja2==3.1.2 Markdown==3.4.4 requests PyYAML==6.0.1
+        run: pip install --disable-pip-version-check --user docraptor==3.1.0 Jinja2==3.1.3 Markdown==3.5.2 requests PyYAML==6.0.1
       - name: Generate MSL release notes from closed GitHub issues
         timeout-minutes: 3
         env:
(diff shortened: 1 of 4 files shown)


beutlich:
Here's the successful test run: https://github.com/beutlich/ModelicaStandardLibrary/actions/runs/2531080050

dietmarw (changes requested):
I don't think relying (again) on a third-party service here is non-optimal. Why not run the PDF/HTML generation directly in the GH actions environment?

beutlich:
OK, we can wait until a new release is scheduled.

beutlich:
> OK, we can wait until a new release is scheduled.

@Harisankar-Allimangalath Since a new release is about to be scheduled we need to decide how to fix the release notes generation script. The proposed changes work for me as indicated by #4251 (though the free docraptor service only enables 5 documents per months w/o watermark).

casella:
@beutlich there are 2 weeks left until we branch off. Do you think you and @Harisankar-Allimangalath can fix it by then?

beutlich:
> @beutlich there are 2 weeks left until we branch off. Do you think you and @Harisankar-Allimangalath can fix it by then?

Since I already spent quite some time on this PR I see three short-term options for the next MSL release:

1. Accept as is (with the dependency on the docraptor service and their free plan of 5 watermark-free PDF documents per month). If it will break some day in future, well, let's revisit the topic by then.
2. Remove/disable the PDF generation, that is only deploy the Markdown and HTML files of the auto-generated release notes.
3. Remove the script (and CI trigger) from the repository, which does not mean, it can't be run from somewhere else.

dietmarw (approved):
I go for the pragmatic solution and go for "1. Accept as is".

Harisankar-Allimangalath (approved):

PR #4006, opened 2022-06-29
Added new power block for raising an input to a user parameterized exponent

mestinso:
This block adds a key missing mathematical function that is not available in Modelica.Blocks.Math. It was previously discussed here: https://github.com/modelica/ModelicaStandardLibrary/discussions/3967

Note that ideally this block should be called "Power" while the existing block called "Power" should be called "Exponentiation". Potentially in the future a conversion script could be used to swap the naming.

### DIFF
--- a/Modelica/Blocks/Math.mo
+++ b/Modelica/Blocks/Math.mo
@@ -1759,6 +1759,42 @@ y = base <strong>^</strong> u;
 </html>"));
   end Power;
 
+  block Exponentiation "Output the input raised to an exponent"
+    extends Interfaces.SISO;
+    parameter Real exponent=2 "Exponent of power" annotation (Evaluate=true);
+  equation
+    y = u^exponent;
+    annotation (Icon(coordinateSystem(preserveAspectRatio=true, extent={{-100,-100},
+              {100,100}}), graphics={
+          Line(points={{0,-80},{0,68}}, color={192,192,192}),
+          Polygon(
+            points={{0,90},{-8,68},{8,68},{0,90}},
+            lineColor={192,192,192},
+            fillColor={192,192,192},
+            fillPattern=FillPattern.Solid),
+          Text(
+            extent={{-86,50},{-14,2}},
+            textColor={192,192,192},
+            textString="^"),
+          Line(points={{-80,60},{-70,27.2},{-60,-1.3},{-50,-25.3},{-40,-45},{-30,-60.3},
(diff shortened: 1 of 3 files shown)


beutlich (dismissed):
Please add a test model in ModelicaTest.Blocks together with the comparisonSignals.txt file of the reference variables.

mestinso:
> Please add a test model in ModelicaTest.Blocks together with the comparisonSignals.txt file of the reference variables.

Done

AHaumer:
Still missing: a test example.
Although you are right, the old Power block should be named Exponentiation and the new block should be named Power, 
the name change is problematic - mybe even with a conversion script.

mestinso:
> Still missing: a test example.
> 
> Although you are right, the old Power block should be named Exponentiation and the new block should be named Power, 
> 
> the name change is problematic - mybe even with a conversion script.

@AHaumer I added a test example in ModelicaTest. However, I didn't add an example within MSL itself because none of the other math blocks seem to have examples either. Let me know what you suggest. 

Regarding the conversion script, agreed with your concerns. Potentially if I use a temporary name (PowerNew or PowerFuture) it would make the conversion easier in the future (Power-->Exponentiation and PowerNew-->Power).

@HansOlsson any thoughts/suggestions on a future rename/conversion? Are direct name swaps possible with the current conversion script specification?

mestinso:
> A name swap should be possible with conversion scripts.
> 
> However, there's also the issue of explaining it and user understanding, and I think a name swap will just create a mess for that, and I would prefer if we didn't plan to do that.
> 
> At least they are not plug-compatible so users will at least notice if they make an error.

Any other naming suggestions? Assuming there is no issue on a direct conversion script name swap, then I think probably the best names are maybe as is in the PR or change to Pow.

A few additional thoughts/notes here:
- I would describe the current situation as confusing. I think the u^p operation is much more prevalent than a b^u operation (not counting the exp operation which is used all over the place). I know when I was first looking for a u^p operation block, I naturally grabbed the Power block and to my surprise it didn't behave as expected. Took me a while to realize one didn't exist.
- I suspect the existing Power block isn't used too heavily, but maybe i'm biased by my own use cases. With that said, I'm not seeing any usage whatsoever in MSL or Buildings or a few different commercial libraries at least.
- Given the above two comments, I don't think the a future conversion would hurt too many people or cause too much confusion, if anything hopefully it would reduce it long term and drive clarity.

mestinso:
@MartinOtter @AHaumer Any additional concerns here? Or can we approve and merge?

MartinOtter (approved):

PR #4017, opened 2022-07-31
Implementation of Gaussian pulse signal source

AHaumer:
When preparing lectures for this fall, I noticed that an important signal source is missing: Gaussian pulse.


christiankral:
OK, as I have no actual application of a Gaussian pulse source, so I just want to summarize and clarify its behavior:

- There is the option to create a sequence of pulses using the parameter `nperiod` 
- The pulses are not superimposed. Instead a strictly periodic signal with no overlap of the individual pulses is created. 
- So the behavior is different from creating a vector of single pulse sources with consecutive `startTime` and adding their outputs. 

@AHaumer I suppose this behavior is intended, isn't it?

Modelica/Blocks/Sources.mo
```diff
@@ -808,6 +808,67 @@ The Real output y is a pulse signal:
...
+      "Equivalent pulse duration";
+    parameter SI.Time period(final min=Modelica.Constants.small)=1
+      "Time for one period";
+    parameter Integer nperiod=1
+      "Number of periods (< 0 means infinite number of periods)";
+    parameter Real offset=0 "Offset of output signal y";
+    parameter SI.Time startTime=0.5 "Time instant of pulse maximum";
+  protected
+    SI.Time t0(start=startTime, fixed=true);
+  initial equation
```
tobolar, line 825: This line can be deleted.

Modelica/Blocks/package.mo
```diff
@@ -1576,6 +1576,40 @@ The output is constant from the beginning.
...
+    connect(pulse.y, integrator2.u) annotation (Line(
+        points={{-9,-20},{-9,-20},{8,-20}},
+        color={0,0,127},
+        smooth=Smooth.Bezier));
+    annotation (experiment(
+        StopTime=4,
+        Interval=0.001,
+        Tolerance=1e-06), Documentation(info="<html>
+<p>
+Compare a single Gaussian pulse with a single rectangular pulse with same amplitude and same area under the pulse.
```
christiankral, line 1608:
```suggestion
Compare a single Gaussian pulse with a single rectangular pulse with the same amplitude and the same area under the curve.

```

christiankral (approved):

Modelica/Blocks/Sources.mo
```diff
@@ -413,13 +413,13 @@ The Real output y is a cosine signal:
           transformation(
           extent={{-2,-2},{2,2}},
           origin={-80,-60})));
-    Blocks.Sources.Constant amplitude_constant(final k=constantAmplitude) if
-      useConstantAmplitude
+    Blocks.Sources.Constant amplitude_constant(final k=constantAmplitude)
+   if useConstantAmplitude
```
tobolar, line 417:
```suggestion
      if useConstantAmplitude
```

Modelica/Blocks/Sources.mo
AHaumer: which line???

HansOlsson: I guess line 849, the "initial equation" as the next line is "equation". will make better comment

See https://github.com/modelica/ModelicaStandardLibrary/pull/4017#discussion_r3986738490

Modelica/Blocks/Sources.mo
```diff
@@ -832,6 +832,68 @@ The Real output y is a pulse signal:
...
+    parameter SI.Time period(final min=Modelica.Constants.small)=1
+      "Time for one period";
+    parameter Integer nperiod=1
+      "Number of periods (< 0 means infinite number of periods)";
+    parameter Real offset=0 "Offset of output signal y";
+    parameter SI.Time startTime=0.5 "Time instant of pulse maximum";
+  protected
+    SI.Time t0(start=startTime, fixed=true);
+  initial equation
+  equation
```
HansOlsson, line 850:
```suggestion
  equation
```
Skip empty section.

Issue #4012, opened 2022-07-05
Improve rotational coupling


tobolar:
I don't understand the meaning of the component `Modelica.Mechanics.Rotational.Components.Coupling` introduced with #3786. Even reading the documentation does not help.

Considering the `Coupling` icon, I would expect similar (or even the same) behaviour like `Modelica.Mechanics.Rotational.Components.Disc`, since it makes an impression of a gear shift (formschüßige Kupplung). Regarding behaviour (equations), it behaves exactly like `Modelica.Mechanics.Rotational.Components.IdealGear` with `ratio=-1`.

The example `Modelica.Electrical.Machines.Examples.DCMachines.DCPM_Drive` even works fine with `Rotational.Components.IdealGear idealGear(ratio=-1)`.

tobolar:
I wonder how the real mechanism looks like. "two electric machines at the shaft" - this is what confuse me, because one shaft could hardly rotate in different directions.
The icon reminds me of a jaw coupling, but the behaviour of the component seems be different.

AHaumer:
Jaw coupling is the right idea:
![grafik](https://user-images.githubusercontent.com/6900273/179467049-1c1f41dc-33a1-4444-b0d1-f678978ca642.png)
For electric machines, positive direction (right) is defined clockwise when looking at the front of the shaft end. 
Therefore one machine rotates cw (right), the coupled machine ccw (left).

beutlich:
I am with @tobolar here: If Coupling simply is a specific paramerization of IdealGear, it should be either removed again or implemented as extension (e.g. short class definition) to make this dependency transparent.

AHaumer:
Well a coupling primarily is _not_ an IdealGear, so extending (including inheriting the icon) is not a good idea. 
Could use the IdealGear inside the coupling model with its own icon.

AHaumer:
The misunderstanding stems from the following:
Looking at the system, both coupled machines turn in the same direction. 
Looking at each of the machines separately, one machine turns left (clockwise), the other one turns right (counterclockwise). 
This affects the induced voltage as well as control of the machine. 
Control of one machine doesn't take into account the other machine (the load).

tobolar:
@AHaumer I can follow your explanation.
Considering again the example `Modelica.Electrical.Machines.Examples.DCMachines.DCPM_Drive` is also helpful: when connecting `dcpm1`directly to `dcpm2`, it would result in the same rotation of both machines. 

[...]

PR #3786, opened 2021-04-05
Rotational coupling

AHaumer:
I just realized that we have no component like an ideal rotational coupling to couple two electrical machines in a realistic way, i.e. when one machine (the devide under test) turns in positive direction and the other one (the dynamometer) has to turn in the opposite direction. To test and demonstrate the usage, I've implemented a nice example (in Electrical.Machines) that shows some other facts about test beds and their power supply that could be usefull for education.

PR #4019, opened 2022-08-05
RotationalCoupling

AHaumer:
Address #4012:
- Moved `RotationalCoupling` to `Modelica.Electrical.Machines.Utilities`.
- Used `IdealGear(ratio=-1)` inside the coupling.
- Improved documentation.


dietmarw (changes requested):
I think this implementation is much cleaner. However, the comparisonSignals and test models introduced in #3786 should get removed/updated

tobolar:
I'm convinced with this solution. 

But sorry for bother you again - I'm still stumbling upon the icon. To me, it makes an impression of a jaw coupling (https://github.com/modelica/ModelicaStandardLibrary/issues/4012#issuecomment-1186856437) which in fact enables some very small relative rotation between the two flanges `a` and `b` - in fact, it is limited mechanically. 
In contrast, considering the example `Modelica.Electrical.Machines.Examples.DCMachines.DCPM_Drive`, the relative angle between the two machines `dcpm` reaches several hunderts of radian.

We have obviously different points of view - while I'm focused on that one component only, you consider the whole drive test rig. Located in the `Electrical` library now, consider this as my last comment on this component and feel free to design the icon on your own.

christiankral:
To have the coupling more obvious in the simulation example it would make sense to add two rotational speed sensors. The doumentation could then refer `speedSensor1.w` and `speedSensor2.w`.

christiankral (approved):
OK for me.

casella:
@AHaumer can you please take care of @dietmarw's [comment]([url](url)), so we can close this issue for the 4.1.0 release? Thanks!

AHaumer:
@casella sorry for the delay, dietmarw's concerns are addressed.

dietmarw:
@AHaumer I can not see that you actually updated the comparison signals. That still needs to get fixed.

AHaumer:
Please @dietmarw bei a little bit more specific.
Comparison signals both of Modelica and ModelicaTest contain coupling.{w, tau}. 
What do you want to be updated in the comparison signals?

dietmarw:
@AHaumer What where you trying to fix with your commit? Your commit fixed another bug of a wrong name reference.
but not the comparison signals, which need to be put in the place correlating with the place of where the coupling now resides. I.e., under Electrical.Machines.... That is what my comment was about.

AHaumer:
> @AHaumer What where you trying to fix with your commit? Your commit fixed another bug of a wrong name reference. but not the comparison signals, which need to be put in the place correlating with the place of where the coupling now resides. I.e., under Electrical.Machines.... That is what my comment was about.

One comparisonSignals.txt resides in \Modelica\Resources\Reference\Modelica\Electrical\Machines\Examples\DCMachines\DCPM_Drive, this is correct I suppose?
The other comparisonSignals.txt resides in \ModelicaTest\Resources\Reference\ModelicaTest\Rotational\AllComponents, which matches the structure of the models.
Would you like to shift the part with the coupling from ModelicaTest.Rotational.AllComponents to an own test example in ModelicaTest.Electrical.Machines or to remove this part completely from AllComponents and from the matching comparionsSignals? What's the harm if it remains as it is?

casella:
@AHaumer since the RotationalCoupling model was moved from Modelica.Mechanics.Rotational to Modelica.Electrical.Machines.Utilities, I guess it makes sense that the corresponding test model is moved from ModelicaTest.Rotational.AllComponents to ModelicaTest.Electrical.Machines, since the coupling model now belongs to Electrical.Machines instead of Mechanical.Rotational.

That also requires to update the referenceSignals.txt of AllComponents, because some components will be removed from that model.

dietmarw:
Thanks @casella for the clarification. That is exactly what the issue is.

AHaumer:
Ok since the component is tested in Modelica.Electrical.Machines.Examples.DCMachines.DCPM_Drive, 
I'll remove it from ModelicaTest.Rotational.AllComponents and of course the corresponding comparisonSignals.

casella (approved):
LGTM

dietmarw (approved):

MatthiasBSchaefer:
This PR needs a back-port to maint/4.1.x

Esther-Devakirubai:
Backported by #4584

Issue #4023, opened 2022-08-29
Obsolete form of noDerivative in ModelicaReference


henrikt-ma:
This form of `noDerivative` is obsolete:
https://github.com/modelica/ModelicaStandardLibrary/blob/02657f5310f208a6c13f103a8fe97fc65d495dcb/ModelicaReference/package.mo#L711

I don't know when it had that form, but at least I checked 3.3r1 where the only allowed form is `noDerivative = inputVar1`.

PR #4024, opened 2022-08-30
Update obsolete form of noDerivative in ModelicaReference

dietmarw:
https://github.com/modelica/ModelicaStandardLibrary/issues/4023


HansOlsson (approved):

henrikt-ma (approved):
Looks good.

(Whether the use of `noDerivative` in the MSL is actually consistent with the semantic restrictions that come with it is another question.)

PR #4031, opened 2022-09-14
Fix HallSensor

### LINKED ISSUE #4030: Bug in a component: Modelica.Electrical.Machines.Sensors.HallSensor
The block Modelica.Electrical.Machines.Sensors.HallSensor has an error.  It uses rem() but it really should be mod(). 

The hall sensor should always be between 0 and 2*pi — no matter the direction of travel.  The electrical angle of a motor is always in that range — as soon as you reach 2 Pi you wrap back to zero and if you pass zero in the negative direction, you wrap to 2 pi. Note that if the angle is negative, the HallSensor gives a nonsensical output.

AHaumer:
Fix #4030  - thanks @ankitnaik177 for pointing this out.


dietmarw (approved):
Looks good.

MartinOtter (approved):
Fine. Note, this is slightly non-backwards compatible because the name of a component has changed (constantSpeed -> speed). Since this is in an example, it is acceptable

PR #4038, opened 2022-09-27
spliceFunction_der check limits

thorade:
there are two if-conditions that used a different value: 1 versus 0.9999999999
this PR makes sure the identical limit is used

this is a follow-up for #3885

### DIFF
--- a/Modelica/Media/Air/MoistAir.mo
+++ b/Modelica/Media/Air/MoistAir.mo
@@ -1429,15 +1429,16 @@ Specific entropy of moist air is computed from pressure, temperature and composi
       input Real deltax=1 "Region around x with spline interpolation";
       output Real out;
     protected
+	  constant Real lim = 0.9999999999;
       Real scaledX1 "x scaled to -1 ... 1 interval";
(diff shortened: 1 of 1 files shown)


Modelica/Media/Air/MoistAir.mo
```diff
@@ -1429,15 +1429,16 @@ Specific entropy of moist air is computed from pressure, temperature and composi
       input Real deltax=1 "Region around x with spline interpolation";
       output Real out;
     protected
+	  constant Real lim = 0.9999999999;
```
beutlich, line 1432: Description would be nice.

thorade:
Added Hans Olsson as possible reviewer, because to my understanding he developed the early/original version of this function.

HansOlsson (approved):
Looks good. 

But I realize that the 0.9999999 is ridiculously accurate. In double precision 0.97 (or 0.99 rounded) is good enough, and it seems it converges quadratically.

The reason to have it all is that there's a risk of round-off close to pi/2.

hubertus65 (approved):

PR #4057, opened 2022-11-14
Refer to SPDX-License-Identifier

beutlich:
Inspired by https://github.com/modelica/Reference-FMUs/issues/300 from @t-sommer it is sufficient to refer to the SPDX-License-Identifier instead of repeating the license text in every file.

### DIFF
--- a/Modelica/Resources/C-Sources/ModelicaFFT.c
+++ b/Modelica/Resources/C-Sources/ModelicaFFT.c
@@ -4,30 +4,7 @@
    Copyright (C) 2003-2010, Mark Borgerding
    All rights reserved.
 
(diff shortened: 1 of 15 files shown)


hubertus65 (approved):

HansOlsson (changes requested):
Exactly how to use SPDX seems unclear, and I would prefer if we had some authorative explanation for the details.

Note that https://reuse.software/faq/ seem to indicate that the SPDX-info in the individual files should be matched with a corresponding file in LICENSES-folder - and that is not yet the case here. (The License-files have other names, and include copyright etc.)

beutlich:
Ok, closing w/o merge, since this is not an important change.

PR #4066, opened 2022-12-14
Replaced state events by time events

### LINKED ISSUE #4065: Unnecessary potential state events in Modelica.Blocks.Sources.Pulse
The blocks `Modelica.Blocks.Sources.Pulse`, `Modelica.Blocks.Sources.SawTooth`, and `Modelica.Blocks.Sources.Trapezoid` seem to have an unnecessarily complicated triggering expression for the pulse. In e.g. Dymola 2023x this results in a state event. 

As the trigger time is known the expression could be reformulated in the standard time-event format: `time >= expr`. 

Suggestion: reformulate
```
when integer((time - startTime)/period) > pre(count) then
```
to
```
when time >= (pre(count) + 1)*period + startTime then
```

AHaumer:
Fixes #4065 - thanks @ErikHenningsson for pointing that out.
For now, I kept the old formulation as documentation //
Well it's an enhancement, not really a bugfix.
@dietmarw should that get back-ported?

### DIFF
--- a/Modelica/Blocks/Sources.mo
+++ b/Modelica/Blocks/Sources.mo
@@ -768,7 +768,9 @@ by a falling exponential signal:
     count := integer((time - startTime)/period);
     T_start := startTime + count*period;
   equation
+    //The following formulation causes a state event
     //when integer((time - startTime)/period) > pre(count) then
+    //A formulation causing a time event is more efficient:
     when time >= (pre(count) + 1)*period + startTime then
       count = pre(count) + 1;
       T_start = time;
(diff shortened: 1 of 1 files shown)


HansOlsson (approved):
Looks good

dietmarw (approved):
@AHaumer I don't see this as a bug (nothing was broken before) but simply a numerical enhancement. So backporting would be overkill if you ask me.

HansOlsson (approved):
Ok

MartinOtter (approved):

Issue #4089, opened 2023-03-12
Faulty parameter propagation in IntersectivePWM


christiankral:
In the model `Modelica.Electrical.PowerConverters.DCAC.Control.IntersectivePWMIntersectivePWM` there are issue with the propagation of three parameters:

### 1 Array size of `trapezoid`

```modelica
Modelica.Blocks.Sources.Trapezoid trapezoid[3](...);
```

should be 

```modelica
Modelica.Blocks.Sources.Trapezoid trapezoid[m](...);
```

### 2 Start time porpagated to `trapezoid`

```modelica
  Modelica.Blocks.Sources.Trapezoid trapezoid[3](
    each final amplitude=uMax,
    each final width=0,
    each final nperiod=-1,
    each final offset=-uMax/2,
    each final rising=0.5/f,
    each final falling=0.5/f,
    each final period=1/f,
    final startTime={startTime - 1.25 + (if refType == ReferenceType.Triangle1
         then 0 else k)/m for k in 0:m - 1}/f)
         if (refType==ReferenceType.Triangle1 or refType==ReferenceType.Triangle3);
```

The units of the expression `final startTime=startTime - 1.25 + (if refType == ReferenceType.Triangle1 then 0 else k)/m for k in 0:m - 1}/f` do not match as startTime is added 1.25 and other real expressions.

### 3 Start time porpagated to `sawTooth` 

Similar issue as 2. The propagation reads `final startTime={startTime - 1.5 + (if refType == ReferenceType.Sawtooth1 then 0 else k)/m for k in 0:m - 1}/f`

---

I understand the different phase start times associated with `Triangle1` and `Triangle3` but I can only suspect that the term `1.5` shall ensure that 

- all the reference signals are fully up at `time >= startTime`
- the first reference signal start with half the maximum value

So my understanding for a three phase saw tooth PWM is:

![sawTooth3](https://user-images.githubusercontent.com/4184218/224538401-306f8851-2709-4b7b-9b22-0c52004248c6.png)

AHaumer:
According to my investigations, the timing of the carrier signals is as shown here:

<img width="307" height="486" alt="Image" src="https://github.com/user-attachments/assets/3ee36163-38dd-4bd6-95c7-7357e5c8fd3c" />

PR #4090, opened 2023-03-12
Fix propagation of start time in IntersectivePWM

christiankral:
Refs #4089


christiankral:
I think it also makes sense to update the documentation to more clearly explain what the difference between the different phase shift strategies is.

AHaumer (approved):
fine with me, thanks!

AHaumer:
Don't know why continuous-integration/ jenkins... fails?

henrikt-ma:
> Don't know why continuous-integration/ jenkins... fails?

As far as I can tell we have a green light, at least as of now.

TManikantan:
@henrikt-ma @AHaumer i think it requires approval form second reviewer for merging.I am not able to add @christiankral for some reason as a reviewer,he is the only other library officer other than @AHaumer who had already given his approval. Can you suggest someone who can review and approve the tickets?

henrikt-ma:
I'm sorry, I'm not the right person to recommend reviewers in this project.

christiankral:
The main purpose of this PR is not the documentation (which I provided so far), but the fix of the propagation of the start time as described in #4089. The actual implementation 
1. does not propagate correct units
` final startTime={startTime - 1.25 + (if refType == ReferenceType.Triangle1 then 0 else k)/m for k in 0:m - 1}/f) ...`
2. and does not (at all) calculate the correct start times.

The following graph shows how I think that the original implementation was intended.

sawTooth3

Two issues have to be resolved before moving forward with this PR:
1. Based on the shown graph: @AHaumer Is this the way how start time shall be implemented? On my opinion the first triangle signal shall start 1/f later than shown, don't you agree? Otherwise the actual start time suggests an instance in time earlier than possibly intended.
2. The meaning of the start time shall also be clarified: At the start time all three bridge branches shall already be supplied with firing signals, but no longer than necessary (this is why I think the first signal shall start 1/f later).

After carifying these issues, we can go ahead with providing a code fix for this BUG (units are wrong and times are not determined correctly! So this fix then shall rather go into maintenance ...

TManikantan:
> The main purpose of this PR is not the documentation (which I provided so far), but the fix of the propagation of the start time as described in #4089. The actual implementation
> 
> 1. does not propagate correct units
>    ` final startTime={startTime - 1.25 + (if refType == ReferenceType.Triangle1 then 0 else k)/m for k in 0:m - 1}/f) ...`
> 2. and does not (at all) calculate the correct start times.
> 
> The following graph shows how I think that the original implementation was intended.
> 
> sawTooth3
> 
> Two issues have to be resolved before moving forward with this PR:
> 
> 1. Based on the shown graph: @AHaumer Is this the way how start time shall be implemented? On my opinion the first triangle signal shall start 1/f later than shown, don't you agree? Otherwise the actual start time suggests an instance in time earlier than possibly intended.
> 2. The meaning of the start time shall also be clarified: At the start time all three bridge branches shall already be supplied with firing signals, but no longer than necessary (this is why I think the first signal shall start 1/f later).
> 
> After clarifying these issues, we can go ahead with providing a code fix for this BUG (units are wrong and times are not determined correctly! So this fix then shall rather go into maintenance ...

@AHaumer would you please comment on the suggestions made by @christiankral

AHaumer:
We are missing a review, and the checks seem to be stuck

casella:
@AHaumer the missing reviewer should obviously be @christiankral, which has expressed some doubts about this PR, e.g. [here](https://github.com/modelica/ModelicaStandardLibrary/pull/4090#issuecomment-1542420069). Please provide him some feedback at your earliest convenience.

@GallLeo please change the admin rights so I can give @christiankral write access to the repo and make him a reviewer.

AHaumer:
As far as I see the PR was created by @christiankral and he can't review his own PR.
So @casella please add your review and it's done.
We have a lack of reviewers ...

christiankral:
@AHaumer Up to now the bug reported in #4089 is not yet fixed through this PR. My request was: Could you please fix this issue the way I proposed it verbally in #4089, considering, that you agree with my proposal.

AHaumer:
This is not real a blocker, and it seems that it is not used a lot. 
It's better to spend some time to get a clean solution instead of having a quick and dirty solutionh.
Therefore: Shift the milestone.

AHaumer (approved):
LGTM

christiankral:
> The main purpose of this PR is not the documentation (which I provided so far), but the fix of the propagation of the start time as described in #4089. The actual implementation
> 
> 1. does not propagate correct units
>    ` final startTime={startTime - 1.25 + (if refType == ReferenceType.Triangle1 then 0 else k)/m for k in 0:m - 1}/f) ...`
> 2. and does not (at all) calculate the correct start times.

Kindly asking @AHaumer to 
- clarify if the described behavior is the intended behavior 
- and provide a fix to the problem of the wrong units

In the actual implementation the wrong units, however, lead to a "wrong" behavior as it has no physical background. 

> The following graph shows how I think that the original implementation was intended.
> 
> sawTooth3

christiankral:
To be clear: The work on this PR has not yet started. We need a fix to the wrong propagation of start times to merge this PR.

AHaumer (approved):
fine with me, sorry for the delay

AHaumer:
Sorry @christiankral for the long delay.
I'm just now finding the time to understand this PR.
I would sketch the timing (for sawtooth) as follows:
PWM
So more or less we agree.
Yes you're right the propagation of startTime is wrong.
For sawtooth it should be:
```
startTime={startTime - 1.5/f + 
  (if refType == ReferenceType.Sawtooth1 then 0 else k)/m/f for k in 0:m - 1}

```
Testing with f=1/3 and startTime=0.5 gives:
grafik
For triangle it should be:
```
startTime={startTime - 1.25/f + 
  (if refType == ReferenceType.Triangle1 then 0 else k)/m/f for k in 0:m - 1}

```
Testing with f=1/3 and startTime=0.5 gives:
grafik
So all `m=3` signals start at the desired startTime, for `time < startTime` they are "out of reach" (no fire signal).

Please accept my apologies for the delay - we might have a phone call to clarifiy how we shall proceed.
IMHO I should correct both formulas, and the PR should get backported.
As far as I see the only change in this PR is instantiating the trapezoid[m] insteat of trapezoid[3].
This is no big deal since m=3 is a constant. I could incorporate this into my PR.
Anything to clarify in the documentation?

We should also discuss the parameter uMax because the sawtooth and the triangle are shifted uMax/2 below zero.

I have enclosed a small test with which I derived the above results.
[PC40894090.zip](https://github.com/user-attachments/files/32070119/PC40894090.zip)

AHaumer:
Replaced by #4812

PR #4105, opened 2023-03-31
Add StoichiometricCoefficient next to StoichiometricNumber

henrikt-ma:
As _stoichiometric coefficient_ and _stoichiometric number_ sound very similar while not being the same (difference in sign convention), they should both be defined so that one doesn't fall into the trap of using `StoichiometricNumber` for a stoichiometric coefficient.


HansOlsson:
I tried to see how important this addition is, and found that while there's no use in MSL it is used in https://github.com/MarekMatejak/Chemical - and I might be mistaken but it seems it is using StoichometricNumber as StoichometricCoefficient - and in that case we need to coordinate the change.

Modelica/Units.mo
```diff
@@ -781,6 +781,8 @@ end UsersGuide;
         min=0);
     type StoichiometricNumber = Real (final quantity="StoichiometricNumber",
           final unit="1");
+    type StoichiometricCoefficient = Real (final quantity="StoichiometricCoefficient",
```
HansOlsson, line 784:
```suggestion
    type StoichiometricCoefficient = Real (final quantity="StoichiometricCoefficient",min=0,
```
(Well, normally they are `>0` but we cannot write that - and in case someone has a general formula it doesn't hurt to allow 0.)

HansOlsson: The reason wasn't just to be pedantic with the min-value but also to have as a differentiator and reminder of which one is which.

henrikt-ma: Sure.  It's a good idea.

henrikt-ma:
> I tried to see how important this addition is, and found that while there's no use in MSL it is used in https://github.com/MarekMatejak/Chemical - and I might be mistaken but it seems it is using StoichometricNumber as StoichometricCoefficient - and in that case we need to coordinate the change.

Wouldn't it be possible to just consider this a library issue?  At the point of introducing the new quantity in the MSL, it won't be in use anywhere (at least not in a way that MSL can be blamed for).  Then, shouldn't it be up to library authors to take advantage of the distinction to ensure quantity safety, while at the same time paying attention to reduce "damage" to component connectability?

HansOlsson:
> Wouldn't it be possible to just consider this a library issue? At the point of introducing the new quantity in the MSL, it won't be in use anywhere (at least not in a way that MSL can be blamed for). Then, shouldn't it be up to library authors to take advantage of the distinction to ensure quantity safety, while at the same time paying attention to reduce "damage" to component connectability?

Possible, but we can also see if we get feedback from users of these types, e.g., from @MarekMatejak

MarekMatejak:
Yes, you are right. Now, the usage in Chemical library is wrong, sorry.  StoichiometricCoefficient instead of StoichiometricNumber in this case could be better and more intuitive.
image
If not defined in MSL, I can still define it in library level. These quantities are not critical for Chemical library, because there are used only for component parameters of chemical processes.

hubertus65 (approved):

MarekMatejak:
Why stoichiometric coefficients or stoichiometric numbers can not be negative? I can imagine the component, where they can be determined dynamically together with dirrection of chemical process. In thi case the sign of these variables can change.  If "min=0" is removed it allows also to define chemical process only by one array of substance connectors, where substrates and products are together.

henrikt-ma:
> Why stoichiometric coefficients or stoichiometric numbers can not be negative? I can imagine the component, where they can be determined dynamically together with dirrection of chemical process. In thi case the sign of these variables can change. If "min=0" is removed it allows also to define chemical process only by one array of substance connectors, where substrates and products are together.

I am not an expert, but feel obliged to answer as I initiated this PR.  The point is that the signedness is the key difference between stoichiometric coefficients and stoichiometric numbers.  This is why the only the new quantity `"StoichiometricCoefficient"` comes with `min = 0`, while the already existing quantity `"StoichiometricNumber"` has no requirement on being non-negative.

My understanding of chemical modeling in Modelica is too limited for understanding the need of having an array of substance connectors, or why such substance connectors would need to be stoichiometric coefficients rather than stoichiometric numbers, for that matter.

henrikt-ma:
Setting milestone 4.1.0 as I think we just need to bring this up for decision at the next MAP-Lib monthly in January.

henrikt-ma:
Regarding the discussion about `min=0`, please note that starting with `min=0` and possibly removing it in the future will be easier than the other way around.

casella:
As far as I understand from the domain expert (@MarekMatejak), it would be good to also have StoichiometricCoefficient, but there is no reason to introduce min = 0 there, as you could have arrays of StoichiometricCoefficients that include both reactants and products, which will of course have opposite signs, as the former are consumed and latter are produced.

I would suggest to follow @MarekMatejak's advice to remove `min = 0` from the type definition of StoichiometricCoefficient and then merge this PR. I don't see the point of introducing this (unneeded) constraint now, to remove it later on.

arunkumar-narasimhan:
I have removed `min=0` based on @casella's suggestion. 

Could you please review? @casella @hubertus65

HansOlsson:
> As far as I understand from the domain expert (@MarekMatejak), it would be good to also have StoichiometricCoefficient, but there is no reason to introduce `min = 0` there, as you could have arrays of StoichiometricCoefficients that include both reactants and products, which will of course have opposite signs, as the former are consumed and latter are produced.
> 
> I would suggest to follow @MarekMatejak's advice to remove `min = 0` from the type definition of StoichiometricCoefficient and then merge this PR. I don't see the point of introducing this (unneeded) constraint now, to remove it later on.

Sorry to interrupt, but I don't think this is good and I don't interpret the comment in that way - but a more as a general discussing about pros and cons of having min=0.

As far as I understand the only difference between StoichiometricCoefficient and StoichiometricNumber is exactly that StoichiometricCoefficient is non-negative; https://en.wikipedia.org/wiki/Stoichiometry#Stoichiometric_coefficient_and_stoichiometric_number

Adding a new type that is identical to a previous one except that the new one has min=0, and then not adding the min=0 just seems confusing to me.

With the min-value it becomes easier to explain for users:
* If you want to have reactants and products in one array use StoichiometricNumber and use different signs for them.
* If you have them in different arrays use StoichiometricCoefficient and make them all non-negative (even if you could use StoichiometricNumber as currently done).

hubertus65 (commented):
I agree with Hans that the (min=0) should be in here, and will approve a review if it is. @casella, I think yoou are wrong there, please check the link provided by @HansOlsson.

MarekMatejak:
I do not see any problem with "min=0" for StoichiometricCoefficient if the user in MSL can chose between StoichiometricCoefficient and StoichiometricNumber (= without "min=0" ... https://goldbook.iupac.org/terms/view/S06025 .. can be negative).

HansOlsson (approved):
ok

PR #4113, opened 2023-04-06
Fix unit errors in BatteryDischargeCharge

henrikt-ma:


qlambert-pro:
That's a duplicate of 4101

PR #4127, opened 2023-05-04
Type cast result of malloc

rfranke:
This is common practice in MSL. It is mandatory when compiling external C code as C++.
Also mind the later cast in "return (void*) table" that is neither required by C nor by C++, nevertheless expressing the intention.

See also: https://github.com/search?q=repo%3Amodelica%2FModelicaStandardLibrary%20malloc&type=code


rfranke:
I have signed the CLA long time ago. For some reason this made it to modelica/fmi-standard but not to modelica/ModelicaStandardLibrary. See: https://github.com/modelica/fmi-standard/commits?author=rfranke

dietmarw:
@rfranke To make things easy it would help if just approved it again in given [link](https://cla-assistant.io/modelica/ModelicaStandardLibrary?pullRequest=4127). Then the CI system can approve your PR automatically. Thank you.

dietmarw (approved):

HansOlsson (approved):
Looks ok.
The only potential downside would be is if there's no definition of malloc (due to outdated automatic int return in C).

HansOlsson:
It now fails on `groovy.lang.Binding` - what/why is that?

sjoelund:
> It now fails on `groovy.lang.Binding` - what/why is that?

Seems to be that Jenkins thinks it's both building a branch and a pull request. It's not mandatory to pass the branch part since we aren't really using them for pull requests in most cases, right?

HansOlsson:
> > It now fails on `groovy.lang.Binding` - what/why is that?
> 
> Seems to be that Jenkins thinks it's both building a branch and a pull request. It's not mandatory to pass the branch part since we aren't really using them for pull requests in most cases, right?

Yes, branches are normally in other repos; that's why it normally doesn't matter. We should maybe change the branch-check to only consider maint-branches?

sjoelund:
I'll try the option to exclude PRs from that testing:
image

PR #4129, opened 2023-05-11
Fix position prescribed friction

m-kormann:
The friction models in `Modelica.Mechanics.Rotational` and `Modelica.Mechanics.Translational` are not designed to be connected to prescribed positions. But sometimes the model structure (e. g. a crankshaft) can lead to such a constellation without intention.

In such a case I found out that depending on the direction of movement change (Forward - Stuck - Backward vs. Backward - Stuck - Forward) the simulation in Dymola errors with a singular inconsistant scalar system. Until now we had a tool-specific discussion with @HansOlsson and Erik Henningsson, but the debate should happen on MSL level. Here our preliminary results:

The locked state is not physical when the velocity is prescribed and this is reflected by a singular linear system in the generated code. It should preferably not be reached at all. Although, it could be argued that it should be allowed as an intermediate iteration in the event iterations when going from Forward to Backward or vice versa. Then it is up to the tool to recover from the singular intermediate state.

The equation for `sa` is:

```
  a_relfric/unitAcceleration = if locked then 0 else if free then sa else 
    if startForward then sa - f0_max/unitForce else if startBackward then 
    sa + f0_max/unitForce else if pre(mode) == Forward then sa - f0_max/
    unitForce else sa + f0_max/unitForce;
```

When going from Forward to Backward we get

```
sa = a_relfric/unitAcceleration - f0_max/unitForce,
```

which allows for triggering `startBackward` and keeping `locked = false`. We actually get exactly the same value for `sa` when going from Backward to Forward, which makes it difficult to trigger `startForward` (we would want the latter term to be added rather than subtracted). As a consequence `startForward` is not triggered and we enter into `locked = true` and are exposed to the singularity.

The problem is that when `pre(mode) == Stuck`, then we get the same equation for `sa` regardless which direction we are going and this equation is biased towards going from Forward to Backward.

In this PR the `else` statement in the equation for `a_relfric/unitAcceleration` is modified to consider the direction of movement change by the signum of `v_relfric`, which allows both sides to be handled correctly.


DagBruck (approved):
This seems to be the minimal change to fix an issue.

m-kormann:
Could you provide a model that fails with the changes in this PR for testing? I compared all MSL examples involving one of the partial friction models and they deliver the same results with and without the changes in this PR.

DagBruck:
Can we continue with this issue? What is stopping us?

m-kormann:
@tobolar Do you have a model that fails with the changes in this PR?

DagBruck (approved):

DagBruck:
At the MAP-LIB meeting Jan 9, @MartinOtter said he would set up a meeting to discuss this issue.

MartinOtter (approved):
I wrongly assumed that sign(v) triggers events. However, this is not the case, but sign(v) is defined as "Expands into noEvent(if v > 0 then 1 else if v < 0 then -1 else 0)".
Since the else-branch "else sa - sign(v_relfric)*f0_max/unitForce;" can only be called
if neither locked nor startForward nor startBackward, v_relfric is guaranteed to be > 0 or < 0 outside of an event iteration - so during integration between events all is fine,
and this formulation does not introduce new events (as I previously thought).
The behavior during event iteration is not fully clear to me.

Since Jakub Tobolar tested this element with the PowerTrain library (that has many usages of the friction element) and all tests are successful and there are cases with predefined position movements where this change gives a successful simulation, its fine for me to accept this change.

Still, intuitively for a Coulomb friction element: When the position (say the rotation angle) is a pre-defined, continuous function, then at der(phi)=0, the friction torque tau_fric is no longer well-defined and -tau_max <= tau_fric <= tau_max, so mathematically no unique solution exists at this point. With the previous implementation of the friction element, the simulation failed. With this change, the friction torque is arbitrarily set to either tau_max or -tau_max and this gives a discontinuous change of the friction torque; in some applications this seem to give a meaningful solution.

beutlich:
This PR was reverted by #4255.

PR #4133, opened 2023-05-17
Replace 'Inline=false' with late inlining

### LINKED ISSUE #3908: Late inlining of Modelica.Mechanics.MultiBody.Frames.Internal.maxWithoutEvent
The current implementations of `maxWithoutEvent`, `maxWithoutEvent_d` and `maxWithoutEvent_dd` in `Modelica.Mechanics.MultiBody.Frames.Internal` all have the annotation `Inline=false`. This is not good. They should have `lateInline = true` annotated instead, to still enable inlining for performance optimization; after all, their implementations are just expressions.

@MartinOtter, @tobolar : Can you confirm this?

tobolar:
Close  #3908


TManikantan:
@MartinOtter @beutlich @christoff-buerger would you please review the fix?

christoff-buerger (approved):
Looks good.

MartinOtter (approved):

PR #4148, opened 2023-06-13
Fix false calculation of w (omega) when resolveInFrame == frame_resolve

tobolar:
Close #4145 

@beutlich Shall this be backported to v3.2.3 as well?
Oh, I see there is https://github.com/modelica/ModelicaStandardLibrary/wiki/Onboarding-MAP-Lib-Project-Leader-and-Deputy#msl-github-issues-and-pull-requests.
Will try to prepare a proper back-port once this PR is reviewed.

Backport to v3.2.3: #4154

PR #4154, opened 2023-06-13
Fix false calculation of w (omega) when resolveInFrame == frame_resolve

tobolar:
Back-port #4148 from master to maint/3.2.3.


beutlich (commented):
Why did you not apply the ModelicaTest changes of the original commit?

Harisankar-Allimangalath:
@beutlich can you please review the changes made here , Thanks

beutlich (approved):

PR #4155, opened 2023-06-13
Added units to mathematical constants

mestinso:
Added units to mathematical constants

- Facilitates unit checking for equations that use these variables

### DIFF
--- a/Modelica/Constants.mo
+++ b/Modelica/Constants.mo
@@ -7,11 +7,11 @@ package Constants
   import Modelica.Units.NonSI;
 
   // Mathematical constants
-  final constant Real e=Modelica.Math.exp(1.0);
-  final constant Real pi=2*Modelica.Math.asin(1.0); // 3.14159265358979;
-  final constant Real D2R=pi/180 "Degree to Radian";
-  final constant Real R2D=180/pi "Radian to Degree";
-  final constant Real gamma=0.57721566490153286061
+  final constant Real e(final unit="1") = Modelica.Math.exp(1.0);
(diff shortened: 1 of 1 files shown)


Modelica/Constants.mo
```diff
@@ -7,11 +7,11 @@ package Constants
...
-  final constant Real e=Modelica.Math.exp(1.0);
-  final constant Real pi=2*Modelica.Math.asin(1.0); // 3.14159265358979;
-  final constant Real D2R=pi/180 "Degree to Radian";
-  final constant Real R2D=180/pi "Radian to Degree";
-  final constant Real gamma=0.57721566490153286061
+  final constant Real e(final unit="1") = Modelica.Math.exp(1.0);
+  final constant Real pi(final unit="1") = 2*Modelica.Math.asin(1.0); // 3.14159265358979;
+  final constant Real D2R(final unit="rad/deg") = pi/180 "Degree to Radian";
+  final constant Real R2D(final unit="deg/rad") = 180/pi "Radian to Degree";
+  final constant Real gamma(final unit="1") = 0.57721566490153286061
```
henrikt-ma, line 14: Since the whole declaration is final, the `final` on the modification is redundant and I would have preferred to leave it out.  However, I see that the other declarations also add `final` to the modification, so it makes more sense to remove all of them at once in a future PR.

henrikt-ma (approved):
Even though this PR goes beyond only adding `unit = "1"` (contrary to https://github.com/modelica/ModelicaStandardLibrary/issues/4046#issuecomment-1589560363), I am fine with all the changes.

HansOlsson (approved):
Ok

PR #4163, opened 2023-06-28
Removed out-of-date/incorrect comment about epsilon_0

mestinso:
While reviewing https://github.com/modelica/ModelicaStandardLibrary/pull/4161, I was reminded that one of the comments was no longer correct. Minor update to the commentary in this PR to resolve. Note that another clarification statement/comment could be added for the reader, however, may need to wait until https://github.com/modelica/ModelicaStandardLibrary/pull/4161 is merged. Otherwise, this should be a quick review and merge.

### DIFF
--- a/Modelica/Constants.mo
+++ b/Modelica/Constants.mo
@@ -28,7 +28,7 @@ package Constants
   // (name, value, description from https://www.bipm.org/en/CGPM/db/26/1/, effective from May 20, 2019)
   // The values for c, q, h, k, N_A are exact and part of the basis of the SI-system
   // Note that the elementary charge uses the common alternate name q since e was taken.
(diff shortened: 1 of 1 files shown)


HansOlsson (approved):

henrikt-ma (commented):
I trust that @HansOlsson made a good review.

mestinso:
> I trust that @HansOlsson made a good review.

Agreed, but I added you because two reviews are required. Let me know if any other suggestions. Thanks.

henrikt-ma (approved):
OK, then I add that I additionally trust Wikipedia, that claims `epsilon_0` to be defined in terms of `mu_0`.  Since the latter is not among the exact ones, neither should the former be.

Issue #3523, opened 2020-03-24
Jumps in signal at very tiny time intervals in PolyphaseTwoLevel_RL reference results


maltelenz:
In the [reference results](https://github.com/modelica/MAP-LIB_ReferenceResults/blob/v4.0.0-beta.2/Modelica/Electrical/PowerConverters/Examples/DCAC/PolyphaseTwoLevel/PolyphaseTwoLevel_RL/PolyphaseTwoLevel_RL.csv) for `Modelica.Electrical.PowerConverters.Examples.DCAC.PolyphaseTwoLevel.PolyphaseTwoLevel_RL` there are two subsequent jumps at an extremely short distance from each other. You can see this in rows 3433-3436 in the linked csv file (can't direct link because github restrictions on file size).

The variables in question are `voltageSensor.v[1]` through `voltageSensor.v[3]`. Plotted:
![image](https://user-images.githubusercontent.com/329963/77441414-b46be980-6de9-11ea-8e66-96e8c6fa146a.png)
Please ignore the labeling on the x axis (bug in SystemModeler).

As you can see, SystemModeler (blue) only has a single event iteration leading to the complete jump, while the reference result (orange) takes two jumps very close to each other.
The reference reports jumps at `0.030250000000000003` and `0.030250000000115126`, a distance of approximately 1.15e-13 apart.

Is there anything in the model that explains the two separate jumps? Is it expected? Do other tools reproduce these two jumps?

AHaumer:
I can't reproduce that with Dymola, I can't zoom in beyond a resolution of ~1e-7.
I can't say that this is expected, but IMHO it's possible - depending on tolerance and event handling. 
There are time events (fundemantalWaveVoltage, fundemantalWaveCurrent, signalPWM) and state events (signalPWM, inverter). It could happen that a state event is iterated within a very short time span before or after a time event and both results are stored. Does the behaviour change with another choice of the solver and / or the tolerance?
Is the difference marked by the comparison tool? 
IMHO it's neglectible. Sorry I don't have time to dog deeper - we are 150% busy organizing online content for our students.

beutlich:
Clsoing as worksforme, finally.

henrikt-ma:
Instead of closing as _worksforme_, shouldn't the correct action be to remove the `voltageSensors.v` signals from the comparison variables?  It would comply with @AHaumer's suggestion to not dig deeper, and would still leave us with 44 less problematic variables to be compared.

[...]

PR #4167, opened 2023-07-07
Remove voltageSensor.v from the comparison list

TManikantan:
I have removed variables voltageSensors.v from the comparison list as per https://github.com/modelica/ModelicaStandardLibrary/issues/3523#issuecomment-760084305

### DIFF
--- a/Modelica/Resources/Reference/Modelica/Electrical/PowerConverters/Examples/DCAC/PolyphaseTwoLevel/PolyphaseTwoLevel_RL/comparisonSignals.txt
+++ b/Modelica/Resources/Reference/Modelica/Electrical/PowerConverters/Examples/DCAC/PolyphaseTwoLevel/PolyphaseTwoLevel_RL/comparisonSignals.txt
@@ -43,6 +43,4 @@ inverter.ac.pin[6].i
 currentSensor.i[1]
 currentSensor.i[2]
 currentSensor.i[3]
-voltageSensor.v[1]
-voltageSensor.v[2]
-voltageSensor.v[3]
+


maltelenz:
@christiankral It is the reference result from Dymola that shows the weird jumps at very tiny time intervals, while System Modeler has fewer jumps, as shown in the linked issue.

The `inverter.fire_p[:]` do have quite a few jumps in System Modeler, but I guess that's expected? Plot:

image

christiankral:
I wonder how the reference signals are created (by Dymola). Is the result mat file simply converted to a CSV file? As @AHaumer commented in https://github.com/modelica/ModelicaStandardLibrary/issues/3523#issuecomment-603776842 there is no obvious way to reproduce the jumps by zooming into the plots.

Possibly @HansOlsson can provide some feedback here: 
- How can the jumps described in #3523 be reproduced in a Dymola simulation? 
- Are the jums also present in the signals `inverter.fire_p[:]`?

HansOlsson:
Yes, the result has a very small differences in time, so you need double precision to resolve it. You can store result files in double precision in Dymola, Simulation>Setup>Output>Format>Double precision

You cannot zoom that hard in the plot window at the moment (you can see there's something odd), but it you use the table-window you see something like:

| |0.03025|0.030250000000115|
|-|-|-|
|`inverter.fire_p[4]`|true, false | false|
|`inverter.fire_p[6]`|true | true, false|
|`voltageSensor.v[4]`|16.666665416937562, -66.66666788306244 | -66.66666788306243, -50.000001223062434
|`voltageSensor.v[6]`|16.666668071707548, 33.33333473170754 | 33.33333473170756, -49.99999856829243

Looking more closely the reason is:
```
Expression signalPWM[4].greaterEqual.u1 < signalPWM[4].greaterEqual.u2 became false ( (signalPWM[4].greaterEqual.u1)-(signalPWM[4].greaterEqual.u2) = 5.55112e-16 )
  Variable inverter.transistor_p.idealGTOThyristor[4].off = 1 at time 0.03025
  Variable inverter.transistor_n.idealGTOThyristor[4].off = 0 at time 0.03025
Expression inverter.transistor_n.idealGTOThyristor[4].s < 0 became true ( (inverter.transistor_n.idealGTOThyristor[4].s)-(0) = -0.122306 )
  Variable inverter.transistor_n.idealGTOThyristor[4].off = 1 at time 0.03025
Expression inverter.diode_n.idealDiode[4].s < 0 became false ( (inverter.diode_n.idealDiode[4].s)-(0) = 1.22306e-06 )
  Variable inverter.diode_n.idealDiode[4].off = 0 at time 0.03025
Shorter attempt for mixed equation system 0.9993892096873109
Expression inverter.diode_n.idealDiode[4].s < 0 became true ( (inverter.diode_n.idealDiode[4].s)-(0) = -0.0610778 )
  Variable inverter.diode_n.idealDiode[4].off = 1 at time 0.03025
Expression inverter.diode_n.idealDiode[4].s < 0 became false ( (inverter.diode_n.idealDiode[4].s)-(0) = 3057.66 )
  Variable inverter.diode_n.idealDiode[4].off = 0 at time 0.03025
Shorter attempt for mixed equation system 0.001019974969863034
  Variable inverter.transistor_n.fire[4] = 1 at time 0.03025
  Variable inverter.transistor_p.fire[4] = 0 at time 0.03025
Iterating to find consistent restart conditions.
      during event at Time :  0.03025
Expression signalPWM[6].greaterEqual.u1 < signalPWM[6].greaterEqual.u2 became false ( (signalPWM[6].greaterEqual.u1)-(signalPWM[6].greaterEqual.u2) = 1.15123e-10 )
  Variable inverter.transistor_n.idealGTOThyristor[6].off = 0 at time 0.03025
Expression inverter.transistor_p.idealGTOThyristor[6].s < 0 became false ( (inverter.transistor_p.idealGTOThyristor[6].s)-(0) = 50 )
Expression inverter.diode_p.idealDiode[6].s < 0 became true ( (inverter.diode_p.idealDiode[6].s)-(0) = -5e+06 )
  Variable inverter.diode_p.idealDiode[6].off = 1 at time 0.03025
Shorter attempt for mixed equation system 0.001000027834150455
  Variable inverter.transistor_n.fire[6] = 1 at time 0.03025
  Variable inverter.transistor_p.fire[6] = 0 at time 0.03025
Iterating to find consistent restart conditions.
      during event at Time :  0.03025000000011513
```
My analysis:
- The cause of the problem is that the different PWM signals are slightly out of sync causing a slight difference in timing. 
- A less accurate crossing function handling would miss the first events, and thus only get one event.
- Whether this difference is a significant for the voltageSensor-signal depend on how you measure errors. If you integrate the errors it wouldn't matter (I think), but with the current tube-handling I believe it will cause problems.

Additionally:
- I don't know exactly what causes the difference in PWM signals, and whether it is correct as currently written.
- Slightly out of sync signals can be really problematic for models, especially in combination with diode-bridges.
- I don't know if the PWM signal generation should changed to avoid this issue by coupling the signals in some way

henrikt-ma (approved):
Looks good, and we even got a newline at the end of the file!

Harisankar-Allimangalath:
@christiankral given the reply from @HansOlsson  how should we proceed with this .

HansOlsson:
Looking more I realize that there's no integration error; it is just that the signals with f1=50Hz and phase -pi/6 and 7pi/6 don't match up perfectly when sampled at f=1000Hz (when they cross at sin(...)=-0.5), unless one uses infinite precision pi.

I don't know how PWMs in practice handle this.

beutlich:
> Overall I am willing to accept this PR, but I would like to have a comment of @AHaumer here first.

Added @AHaumer as reviewer.

AHaumer (changes requested):
IMHO the differences (1 or 2 steps) are not caused by the model - for me it seems like a tool issue.
I dislike the idea of removing signal from comparisonSignals because two tools show different event iteration. 
Should we really remove all non-trivial signals to avoid problems?

casella:
I am not an expert in power converters, but my understanding of this issue is that the simulation of PWM systems will always be potentially critical from the point of view of tool comparisons: there are a gazillion events, some of them should be synchronous but actually aren't (due to numerical approximations) and they trigger instantaneous and large voltage changes. Getting those jagged voltages waveforms to match perfectly across tools that use different numerical strategies seems quite difficult to me. 

The question is, is that a problem, do we care and does it matter? I think the answer is no, no, and no.

I would say that the goal of these models is not to predict with absolute precision the triggering instants (which in real life is affected by some jitter, BTW), but rather to predict if the generated waveforms are OK, _according to some integral criterion_: if a certain switch is triggered 1ns later, there is an instantaneous large voltage error, but the impact this has on the actual load (e.g., a motor) in terms of transferred energy is totally negligible. I don't think we should split hairs in 16 parts by asking all tools to produce _exactly_ the same results on these logic signals. BTW, in real life there will probably be some stray capacitance effects, so those voltage jumps won't actually be jumps. The difference between two very close steps and one step is pure fiction, compared to the real waveforms. Again, we don't care about that, what we simulate is an idealization which, as Dirk explained well in his plenary in Aachen, can lead to some numerical issue. The problem is the idealization, not the numerical issue.

From this point of view, the inductor currents seem to me a much better indicator to compare the simulation results across tools, _because they have inherent integral action_, due to their "current inertia", which smooths out these small time differences _that are totally negligible from a practical point of view._ If a tool messes up with the solution (not just by a 1 ns jitter) of the voltage triggering logic, the outcome will inevitably be different current waveforms, which will be revealed by the comparison.

This is a particular case of a principle that we are always using when automatically selecting reference trajectories: if a model has state variables, only use them as references. The rationale is that state variables have inertia, which inherently smoothes out small errors on the time axis; on the other hand, if there are serious and not too short discrepancies on some algebraic variables, these eventually also change the state trajectories.

According to this principle, if I have a heat exchanger model I would probably only check the fluid temperatures (which have inertia) - I wouldn't care about very small jitter errors in the flow rates, which are often also dirtier due to numerical noise.

If we are fine with that principle, then we should be happy to only compare the currents and go ahead with this PR.

My 2 cts.

AHaumer (approved):
ok following the elaborate explanation of @casella I agree with sticking to the currents and deleting the voltages from the list of comparison signals.

christiankral (approved):
OK now.

Harisankar-Allimangalath:
@christiankral @AHaumer we have some errors in the check so it can't be merged and how should we proceed then ?

beutlich:
> @christiankral @AHaumer we have some errors in the check so it can't be merged and how should we proceed then ?

All green again.

PR #4176, opened 2023-08-08
Fix (and rename) parameters of Modelica.Electrical.Digital.Examples.Utilities.DFF

### LINKED ISSUE #4174: Unused parameters in Modelica.Electrical.Digital.Examples.Utilities.DFF
The parameters Tdel and QInit are not used. 
I assume they should just be propagated to the RSFF1.

beutlich:
Closes #4174.

Note, that I renamed the two parameters to match the identifiers of the underlying classes. No conversion script is needed for such example components.


hubertus65 (approved):

henrikt-ma (commented):
While I often think that `final` parameter propagations are good, I think someone like an MSL sub-library officer is needed to know whether it is the right thing to do here.

beutlich:
> While I often think that `final` parameter propagations are good, I think someone like an MSL sub-library officer is needed to know whether it is the right thing to do here.

Someone needs to take over, because the original [library officers](https://github.com/modelica/MAP-LIB_ProjectRules/blob/master/MAP-LIB_ProjectRules.md#annex-3-library-officers) of Electrical.Digital are no loger active.

hubertus65:
Yes, that is exactly the problem that the former library officers are no longer active.

henrikt-ma:
> Yes, that is exactly the problem that the former library officers are no longer active.

Right.  This is a reminder of one of the unavoidable problems we run into due the monolithic MSL architecture.

hubertus65:
People leaving due to job change/ retirement has nothing to do with that. We have exactly per-sublibrary officers, so arguably this is the only part of MSL that is not monolithic in terms of practice.

MartinOtter (approved):

GallLeo:
MAP-LIB group during monthly meeting:
Just formal issue of no library officer.
Proposal seems fine.

PR #4177, opened 2023-08-08
Delete successful and skipped workflow runs by GitHub Actions (CI)

beutlich:
See #4134.


tobolar (approved):
Looks good.

beutlich:
One effect of the workflow removal is that open PRs are now missing their required check results from GitHub Actions:

grafik

I asked in https://github.com/dmvict/clean-workflow-runs/issues/5 for some improval.

PR #4187, opened 2023-08-31
Remove newlines at end of Text primitive

maltelenz:
Newline behavior in Text primitives is undefined, and in these cases seem to serve no purpose regardless.


AHaumer (approved):
as discussed at MAP-Lib monthly 2023-12-12

HansOlsson (approved):
Seems good.

PR #4195, opened 2023-09-18
Add all defined outputs of the test example to reference signals

tobolar:


HansOlsson (approved):
Ok, makes sense for better testing.

But the harmonic part is just weird: we internally generate re and im parts, then in harmonic convert to polar, and then outside the harmonic component we convert back to re and im, and then to polar again!

tobolar:
>  we internally generate re and im parts, then in harmonic convert to polar, and then outside the harmonic component we convert back to re and im, and then to polar again!

Probably to test also `PolarToRectangular` and `RectangularToPolar` in just one test?

beutlich (approved):
Thanks!

PR #4205, opened 2023-10-18
Remove redundant diagram text from Analog examples 

christiankral:
The full description is provided in #4186

Redundant text is removed from analog simulation examples.


HansOlsson (changes requested):
The basic idea is good, but:
- Some commits are from a separate PR and shouldn't be included.
- There's something weird with file-endings, I believe you accidentally deleted the end-of-line at the end of the file.

After that is cleared up I can review the rest.

maltelenz:
I agree, this is great!

I guess I'll wait with reviewing until Hans comments above have been considered.

christiankral:
Yeah, right. Something went wrong here. I will close this PR an re-open a new one.

Issue #4233, opened 2023-11-19
Old value for universal gas constant used for calculation of specific gas constant


Parti-Gyle:
I noticed that the specific gas constant stated in the data records for the ideal gases (`SingleGasesData.mo`), has been calculated with a universal gas constant value of $`R=8.3144720`$; which I believe would have been the value for $`R`$ when the Media library was written. 

However, in the McBride B.J., Zehe M.J., and Gordon S. (2002) (*NASA Glenn coefficients*) paper they use a value $`R=8.314510`$ (Appendix A). Yet since then, the universal gas constant has been redefined $`R=N_A.k=8.31446261815324\; \: J\cdot K^{-1}\cdot mol^{-1}`$.

I know we're talking about a difference in the 5th/6th significant figure, but should all the specific gas constants be recalculated $R_S$? And if so, what value of the universal gas constant $`R`$ should be used?

I'm happy to submit a pull request the the relevant changes if it's deemed necessary.

beutlich:
Thanks for raising this issue.

I am not sure if we really need the calculated value for R_s, or rather can live with an expression `final R_s=Modelica.Constants.R/MM`.

Parti-Gyle:
I agree that would remove the need to update all the specific gas constants whenever the universal gas constant in `Modelica.Constants.mo` gets updated.  However, $R$ in MSL is a different value to that which was used in the NASA paper and this difference will propagate to the thermodynamic properties calculated with the NASA Glenn coefficients.  

I know the error will be small, using a newer value of $R$, but the thermodynamic properties calculated will differ from what the original NASA model produces.

Should the specific gas constants not be recalculated with `$R=8.314510`$, to remain true to the NASA source?

mestinso:
I did a cursory glance here and based on my experience with fluid property routines such as this one, yes normally one should always use the universal gas constant that was used by the authors of the routine. In this case, that appears to be the 8.314510 value.

So, the R_S value in each data record should probably be replaced by MM/R_NASA2002, where R_NASA2002=8.314510.

...moving forward, it will be nice when all routines use the now truly constant new universal gas constant value. But for all routine created before ~2020, yes, I think we will have these other/older gas constant values in the mix. (Note that the alternative is much messier).

With that said, probably best to get agreement/confirmation from the author/officer for this portion of the library.

hubertus65:
Thanks Matthis, for the link, quite insightful. I don't find this easy, trivial to decide.

[...]

PR #4238, opened 2023-12-10
R_s for each ideal gas record now calculated with R_NASA

Parti-Gyle:
Recalculated all the `R_s` values for each Ideal gas record in SingleGasesData, using `R=8.314510` specified in the original 
McBride B.J., Zehe M.J., and Gordon S. (2002) (NASA Glenn coefficients) paper.

See #4233 for discussion.

### DIFF
--- a/Modelica/Media/IdealGases/Common/SingleGasesData.mo
+++ b/Modelica/Media/IdealGases/Common/SingleGasesData.mo
@@ -13,7 +13,7 @@ package SingleGasesData "Ideal gas data based on the NASA Glenn coefficients"
     ahigh={-330992.637,982.0086420000001,1.381179917,0.0006170899989999999,-1.6881146e-007,
         2.008826848e-011,-5.627285655e-016},
     bhigh={27267.19171,14.56862733},
-    R_s=77.07991789980736);
+    R_s=77.08027018157344);
 
   constant IdealGases.Common.DataRecord Agplus(
     name="Agplus",
@@ -27,7 +27,7 @@ package SingleGasesData "Ideal gas data based on the NASA Glenn coefficients"
     ahigh={-53274323.49999999,131071.0631,-109.8820208,0.0482600276,-1.093661557e-005,
         1.26383591e-009,-5.852542535e-014},
     bhigh={-743912.2509999999,844.6266189999999},
-    R_s=77.08030991764042);
+    R_s=77.08066220119817);
 
   constant IdealGases.Common.DataRecord Agminus(
     name="Agminus",
@@ -39,7 +39,7 @@ package SingleGasesData "Ideal gas data based on the NASA Glenn coefficients"
     blow={17665.65919,5.8696798},
     ahigh={0,0,2.5,0,0,0,0},
     bhigh={17665.65919,5.8696798},
-    R_s=77.07952588596174);
+    R_s=77.07987816593619);
 
   constant IdealGases.Common.DataRecord Air(
     name="Air",
@@ -53,7 +53,7 @@ package SingleGasesData "Ideal gas data based on the NASA Glenn coefficients"
     ahigh={241521.443,-1257.8746,5.14455867,-0.000213854179,7.06522784e-008,-1.07148349e-011,
         6.57780015e-016},
     bhigh={6462.26319,-8.147411905},
-    R_s=287.0512249529787);
+    R_s=287.0525368759184);
 
   constant IdealGases.Common.DataRecord AL(
     name="AL",
@@ -67,7 +67,7 @@ package SingleGasesData "Ideal gas data based on the NASA Glenn coefficients"
     ahigh={-29208.20938,116.7751876,2.356906505,7.73723152e-005,-1.529455262e-008,
         -9.97167026e-013,5.053278264e-016},
     bhigh={38232.8865,6.600920155},
-    R_s=308.1541163442944);
+    R_s=308.1555247147142);
 
   constant IdealGases.Common.DataRecord ALplus(
     name="ALplus",
@@ -80,7 +80,7 @@ package SingleGasesData "Ideal gas data based on the NASA Glenn coefficients"
     ahigh={-4181.18325,-9.94855727,2.548615878,-5.878760040000001e-005,
         3.132291294e-008,-7.74889463e-012,7.27444769e-016},
     bhigh={109101.1485,3.48866729},
-    R_s=308.1603819910326);
+    R_s=308.1617903900885);
 
   constant IdealGases.Common.DataRecord ALminus(
     name="ALminus",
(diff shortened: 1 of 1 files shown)


beutlich:
Thanks!

Did you change all 1241 manually? Or do you have some automation script?

Might be desirable to have a revison (or info) annotation stating the used value for R.

mestinso:
It seems to me that it would have been better if, instead of recalculating and entering each R value 1 by one,  you just  replaced the R values with an expression in its place (`MM/R_NASA2002`). This way, the R value is more directly traceable to the original source and someone (like yourself) doesn't have to reverse engineer what the universal value must have been in order to catch a bug.

beutlich:
> It seems to me that it would have been better if, instead of recalculating and entering each R value one by one, you just replaced the R values with an expression in its place (`R_NASA2002/MM`).

H, having an expression is exactly what I first proposed in https://github.com/modelica/ModelicaStandardLibrary/issues/4233#issuecomment-1819417794.

Parti-Gyle:
I didn't change all the values of `R_s` to an expression, as I didn't think having an old value for R that any user could inadvertently reference was a good idea.  There should only be one value for the constant R, stored in the constants package.  As suggested [#4233](https://github.com/modelica/ModelicaStandardLibrary/issues/4233#issuecomment-1851590325), these ideal gas routines should ideally use the now fixed ideal gas constant.  But, this would require re-calculating all the polynomial coefficients for every species, which I'm not prepared to do. 

I think this is a reasonable compromise.  But, please let me know what changes are required to get this merged. 
Thanks

mestinso:
Agreed with your concern, but this is why the constant should live within the relevant subpackage and should generally not be used outside that subpackage. Note that this isn't the first spot in MSL where different universal R values are being used instead of the one in Modelica.Constants. See an 8.314371 R value here: Modelica.Media.Air.ReferenceMoistAir.Utilities.Water95_Utilities or the same 8.314510 R value here: Modelica.Media.Air.ReferenceAir.Air_Utilities.Basic.

So, for example, something like following could be done:
image
It should be clear from the constant name and its location that this constant is intended for local use.

One annoying thing about this implementation is that while it's much more traceable where the R_s value came from, it's a little more annoying to make a script to automatically make the change since the expression has the instance name in it.

hubertus65:
See my comments in #4233. I am unsure. 

- If we add more advanced mixture routines later, I think the advice from NIST to only use the latest R for everything is sound. 
- That would require reverting the proposed change. Often, the NASA data is the only data available. for some fluids. 
- My first thought was yes, the "local R 2002" would be the right choice, but not after reading the discussion at NIST.

mestinso:
> See my comments in #4233. I am unsure.
> 
>     * If we add more advanced mixture routines later, I think the advice from NIST to only use the latest R for everything is sound.
> 
>     * That would require reverting the proposed change. Often, the NASA data is the only data available. for some fluids.
> 
>     * My first thought was yes, the "local R 2002" would be the right choice, but not after reading the discussion at NIST.

As mentioned above (https://github.com/modelica/ModelicaStandardLibrary/pull/4238#issuecomment-1861789063), I would point out that this isn't the only spot in MSL where "local R" values are used. If we are proposing using the new final R value here, we should probably use the new final R value everywhere. Why be inconsistent? Arguably that is a bigger decision beyond this PR (which just fixes a bug with the original NASA gas implementation).

With that said, I think it's worth pointing out that there are some key differences and considerations in MSL compared to REFPROP that may lead to different decisions from what NIST decided:
1) As far as I know, we aren't facing any current pain points due to different gas constants being used. Whereas they cite live issues related to "enhanced mixture functions" among other things. For us, any pain points are hypothetical ones for functionalities not yet implemented.
2) NIST still maintains a flag that the user can turn on to go back to the original R value. This is important for a) routine verification, b) if max accuracy is required (rare), or c) if the user wants to avoid any sort of change across REFPROP releases. In our case, I don't think we are going to support any sort of flag (new gas constant vs original gas constant), so if we switch the gas constant now we have introduced differences from the primary reference with no easy way to revert on demand. This means that a user or developer cannot easily verify whether the routine is implemented correctly. There will be (minor) differences and one cannot be sure if those are differences due to the R value difference, some bug, or both. REFPROP doesn't have this issue because they can fall back to the original routine on demand for verification.
3) Lastly I would comment that Eric Lemmon expressed hesitation (ie it's not a black-and-white decision) and they didn't even fully make the switch (unless I'm misinterpreting his comment): I think he mentioned oxygen, water, nitrogen, argon, and others that still use the original R value.

Given these facts and considerations, I don't see much upside in MSL to switch over to the new R value across the board, whereas I do see various downsides. With that said, as new routines are released which displace the old ones, I think it would be a great goal to adopt them and eventually reach a future state with routines that only use the new R value. Maybe an MSL 5.0 goal....

hubertus65:
Ok,Matt, @mestinso, I buy your arguments and am fine with the proposed change. Thanks for your efforts! @GallLeo, please be aware that this change might lead to regression test failures for any test involving ideal gases, depending on the tolerance chosen. THEY CAN BE IGNORED, if sufficiently small.

hubertus65 (approved):
After some through thinking about this case, I agree with the proposed changes.

beutlich:
> After some through thinking about this case, I agree with the proposed changes.

Just to be on the safe side: Which means, the PR needs to be updated first to not use/introduce R_NASA, right?

casella:
Thorny issue indeed, when constants change all the time. BTW, this also happens with other constants, such as the critical pressure of water, which changed several times during my career. The discussion split between this PR and #4233 makes things easier 😅.

As @beutlich, I am also a bit confused about exactly what is the final decision. I'll try to summarize what I understood of the whole discussion:
- The original SingleGasesData.mo written by  @hubertus65 in the days of yore contained literal constants for R_s
- @Parti-Gyle noticed that those values were computed with the current value of R = 8.3144720 circa 2000 AD, but in fact the NASA paper used R = 8.314510, so he proposed that we use that value instead, and prepared this PR accordingly
- In the meantime, the Bureau Internationale de Poids et Mésures determined an ultimate value of R = ka*N, never to be changed in the future.

So, I understand we now have three choices:
1. Leave SingleGasesData as it is. No regressions, but it's nether using the ultimate value of R, nor the original one of the Glenn paper
2. Merge this PR to make SingleGasesData consistent with the Glenn paper, tough it uses a "local" value of R, as other fluid models in MSL do. Very small regressions possible
3. Modify the PR so that it uses the ultimate value or R = Modelica.Constants.R instead of the NASA Glenn value. Even smaller regressions possible.

I guess both @mestinso and @hubertus65 are both in favour of 2., as I also am, for what is worth. Please confirm before we proceed.

Then, try to get ChatGPT-4 to figure this out automatically from all the posts. Just kidding 😄

mestinso:
> I understand @mestinso and @hubertus65 are both in favour of 2., as I also am, for what it's worth. Please confirm before we proceed.

:+1: That's correct: in favor of 2.

beutlich:
> I understand @mestinso and @hubertus65 are both in favour of 2., as I also am, for what it's worth. Please confirm before we proceed.

Thanks for clarification.

beutlich (approved):

Harisankar-Allimangalath:
Merging this PR

PR #3825, opened 2021-06-03
Add mass fraction functions to Modelica.Media

wischhusen:
I made an attempt to resolve #3650. Please check.

PR #4243, opened 2024-01-03
- added partial massFraction to partialMedium

### LINKED ISSUE #3650: Modelica Media Interface is information incomplete w.r.t. independent mass fractions
The Modelica Media package provides an abstract interface to the thermodynamic state of the Medium by defining a set of "set" and "get" like functions. So far, so good.

Unfortunately it occured to me that this interface is incomplete. Using the "get"-like functions, I cannot fully retrieve the thermodynamic state because there seems to be no function for the (independent) mass fractions. When you start using the Media models with different interfaces than in Standard fluid, you start to notice.

Irrespective of what a modeller wants to do, I think it is a good idea to make the abstract interface complete. It can also be easily fixed. 

I propose to add the following function to PartialMedium

  replaceable partial function massFraction "returns independent mass fraction"
    input ThermodynamicState state "Thermodynamic state record";
    output MassFraction Xi[nXi] "(independent) Mass Fraction";
  end massFraction;

Then add to PartialPureSubstance the implementation

redeclare replaceable function massFraction "Return independent mass Fraction"
    input ThermodynamicState state "Thermodynamic state record";
    output MassFraction Xi[nXi] "(independent) Mass Fraction";
  algorithm 
    Xi := fill(0,0);
  end massFraction;

and last to PartialMixtureMedium

replaceable function massFraction "Return independent mass Fraction"
    input ThermodynamicState state "Thermodynamic state record";
    output MassFraction Xi[nXi] "(independent) Mass Fraction";
  algorithm 
    Xi := state.X[1:nXi];
  end massFraction;

Objections? Other ideas? Shall I branch and raise a merge request?

We need this for our new fluid library.

dzimmer:
This is a second pull request replacing #3825 which attempted to resolve #3650.
#3825 should be discarded and archived consequently.

This pull request is based on the most recent master of the MSL andI have taken the critic of #3825 into account:
- The comments now speak of "mass fractions"  instead of "mass fraction" in order to be consistent with the ot´ther comments
- I have extended the examples of ModelicaTest to include calls of this function at 3 examples to cover the pureSubstance case as well as the reduced and non-reduced Mixture case.

The added function is till called massFraction not massFractions for the following reasons:
- to be consistent with the outputType
- to be compatible with the ThermoFluidStream implementation already using this function
- to be compatible with the XRG extensions to their media where they have implemented it.

The following changes are included in the pull request.

- implemented massFraction for partialPureSubstance
- implemented massFraction for PartialMixtureMedium
- added 3 test cases for ModeliceTest covering pure / reduced mixture / non-reduced misture

### DIFF
--- a/Modelica/Media/package.mo
+++ b/Modelica/Media/package.mo
@@ -4539,6 +4539,12 @@ This function computes an isentropic state transformation:
               X));
     end specificEnthalpy_psX;
 
+    replaceable partial function massFraction "Return independent mass fractions (if any)"
+      extends Modelica.Icons.Function;
+      input ThermodynamicState state "Thermodynamic state record";
+      output MassFraction Xi[nXi] "(independent) Mass fractions";
+    end massFraction;
+	
     type MassFlowRate = SI.MassFlowRate (
         quantity="MassFlowRate." + mediumName,
         min=-1.0e5,
@@ -4722,6 +4728,14 @@ are described in
               T,
               fill(0, 0)));
     end density_pT;
+
+    redeclare replaceable function massFraction "Return independent mass fractions (if any)"
+      extends Modelica.Icons.Function;
+      input ThermodynamicState state "Thermodynamic state record";
+      output MassFraction Xi[nXi] "Independent mass fractions";
+    algorithm
+      Xi := fill(0,0);
+    end massFraction;
 
     redeclare replaceable partial model extends BaseProperties(final
         standardOrderComponents=true)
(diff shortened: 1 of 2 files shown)


Modelica/Media/package.mo
```diff
@@ -4539,6 +4539,12 @@ This function computes an isentropic state transformation:
               X));
     end specificEnthalpy_psX;
 
+    replaceable partial function massFraction "Return independent mass fractions (if any)"
+      extends Modelica.Icons.Function;
+      input ThermodynamicState state "Thermodynamic state record";
+      output MassFraction Xi[nXi] "(independent) Mass fractions";
```
beutlich, line 4545:
```suggestion
      output MassFraction Xi[nXi] "Independent mass fractions";
```

MartinOtter (approved):

Modelica/Media/package.mo
```diff
@@ -5095,6 +5109,14 @@ to the above list of assumptions</li>
       MassFraction[nX] X(start=reference_X)
         "Mass fractions (= (component mass)/total mass  m_i/m)";
     end ThermodynamicState;
+	
+	redeclare replaceable function massFraction "Return independent mass fractions (if any)"
```
tobolar, line 5113:
```suggestion

    redeclare replaceable function massFraction "Return independent mass fractions (if any)"
```

tobolar: Tabs replaced by spaces.

hubertus65 (approved):

casella:
LGTM

PR #4247, opened 2024-01-10
Update version, uses annotation and copyright year

casella:
This PR updates all version and uses annotation to 4.1.0. It also updates copyright dates to 2024. Finally, it removes a spurious version annotation in a sub-package.

PR #4251, opened 2024-01-10
Create Modelica.UsersGuide.ReleaseNotes.Version_4_1_0

beutlich:
The ResolvedGitHubIssues documents are created from my branch fix-release-notes-generation (while Modelica/Resources/Documentation/Generate-ReleaseNotes.py is broken and #3999 still is open). No need to merge them now, since they should be updated immediately before tagging.


AHaumer (approved):
ok

casella:
Thanks @beutlich! Two questions:
- does this PR replace #3999 as a backup in case we can't fix the Python script
- should we merge for the tagging of the alpha.1 version and then update with beta and rc versions?

Adding @arunkumar-narasimhan and @Harisankar-Allimangalath to the loop.

beutlich:
@casella We have/had multiple kind of release notes:

1. This PR (and its follow-up issue #4260) care about the manually edited release notes.
1. PR #3999 is about the resolved GitHub issues. Which is nice to have (once all issues/PRs are properly labeled). Thus, it does not replace the manual editing of release notes.
1. There also is the pretty useful MSL version comparison (which can be another source of input for the manual editing).

arunkumar-narasimhan:
@casella, @beutlich, My understanding is that #4251 and #3999 are independent PRs, so we should review and merge #4251 and then #4247 to proceed with branching while working on #3999. 

if so, could you please review this PR @casella?

Harisankar-Allimangalath:
@beutlich from your understanding also this PR is independent of #3999  right , so once we get this approval , we can proceed with branching while working on #3999 right ?

Harisankar-Allimangalath (approved):

PR #4252, opened 2024-01-10
Partially revert 18320a6

beutlich:
MSL 4.0.0 never was based on MoLang 3.6. Reverting that change of 18320a68b295c9b0be7a5fade0420487ec0059dc.


HansOlsson (approved):
Good. It is 4.1.0 that will be based on 3.6, whereas 4.0.0 has always been on 3.4.

maltelenz (approved):

PR #4254, opened 2024-01-11
Reintroduce Arrow as a visualization shape

tobolar:
Reintroduce `Arrow` as a common visualization object, similar to a box or a sphere. In the MultiBody, the `Arrow` is now used for visualization of axes of coordinate system in 
- `MultiBody.Visualizers.FixedFrame`
- `MultiBody.World`

To better understand the idea, see also https://github.com/modelica/ModelicaStandardLibrary/issues/3931#issuecomment-1594362054

This PR bases on #4248 and is, thus, a draft only. It shall be rebased, checked and changed to a full PR after merging that other PR. 

Note: I didn't change `DoubleArrow` since I consider this being an obsolete model.

### DIFF
--- a/Modelica/Mechanics/MultiBody/Visualizers/Advanced/Arrow.mo
+++ b/Modelica/Mechanics/MultiBody/Visualizers/Advanced/Arrow.mo
@@ -5,6 +5,7 @@ model Arrow
   import Modelica.Mechanics.MultiBody.Types;
   import Modelica.Mechanics.MultiBody.Frames;
   import T = Modelica.Mechanics.MultiBody.Frames.TransformationMatrices;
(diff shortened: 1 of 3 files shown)


tobolar:
I've realized the usage of `Mechanics.MultiBody.Visualizers.Advanced.Arrow` inside of `Mechanics.MultiBody.World` leads to recursion issue since `Arrow` itself uses `outer world`. A solution could be to define `Mechanics.MultiBody.Visualizers.Internal.Arrow` which do not uses world.

casella:
@tobolar this change could have other side effects that we are not aware of, and potentially break some aspects of the MultiBody library. 

I think it would be wiser to postpone this to MSL 4.2.0, so it can be thoroughly tested with multiple tools.

PR #4255, opened 2024-01-12
Revert "Fix position prescribed friction"

MartinOtter:
Reverts modelica/ModelicaStandardLibrary#4129

When simulating a model with the changed friction model the following error appears in Dymola 2023x:

Compiling and linking the model (Visual C++). 

dsmodel.c
dsmodel.c(176): error C2065: "ModelicaStandardTables_CombiTable1D_init3": nichtdeklarierter Bezeichner
dsmodel.c(176): error C2064: Ausdruck ergibt keine Funktion, die 337 Argumente bernimmt

Error generating Dymosim.

PR #4256, opened 2024-01-12
Revert "Merge pull request #4255 from modelica/revert-4129-fix-friction"

m-kormann:
The fix from #4129 has been accidentally reverted by #4255 without being the reason for the compiling errors.

The compilation errors should be fixed by recompilation of the binaries as discussed in #3911, #4178 and #4250.

### DIFF
--- a/Modelica/Mechanics/Rotational/Interfaces/PartialFriction.mo
+++ b/Modelica/Mechanics/Rotational/Interfaces/PartialFriction.mo
@@ -54,9 +54,9 @@ equation
     mode) == Backward or startBackward);
 
   a_relfric/unitAngularAcceleration = if locked then 0 else if free then sa
-     else if startForward then sa - tau0_max/unitTorque else if
-    startBackward then sa + tau0_max/unitTorque else if pre(mode) ==
-    Forward then sa - tau0_max/unitTorque else sa + tau0_max/unitTorque;
+    else if startForward then sa - tau0_max/unitTorque
+    else if startBackward then sa + tau0_max/unitTorque
+    else sa - sign(w_relfric)*tau0_max/unitTorque;
 
   /* Friction torque "tau" has to be defined in a subclass. Example for possible
      realization of a clutch utilizing interpolation by ExternalCombiTable1D see:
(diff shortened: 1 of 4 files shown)


DagBruck (approved):
An improvement compared to the status-quo. As discussed in the earlier PR 4129.

MartinOtter (commented):
I have re-analyzed the solution and propose a slightly different fix. I would have included this fix in the pull request, but I don't manage to quickly do it (and do not have enough time to figure this out). So, someone has to include this solution.

Issue with current proposal:

  a_relfric/unitAngularAcceleration = if locked then 0 else if free then sa
   else if startForward then sa - tau0_max/unitTorque
   else if startBackward then sa + tau0_max/unitTorque
   else sa - sign(w_relfric)*tau0_max/unitTorque;

Assume mode == Forward, so forward sliding, and wrelfric crosses zero. Then previously, the equation "mode = .... w_relfric > 0 ..." triggered a state event at w_relfric=0 and during finding the root w_relfric=0, all signals are continuous. With the change above, this is no longer the case, because sa and a_relfric are computed with "a_relfric = sa - sign(w_relfric)..." during root finding, so a_relfric and sa are discontinuous during root finding (which is bad).

This can be fixed by reformulating the equations to:
  a_relfric/unitAngularAcceleration = if locked then 0 else if free then sa
   else if startForward then sa - tau0_max/unitTorque
   else if startBackward then sa + tau0_max/unitTorque
   else if pre(mode) == Forward then sa - tau0_max/unitTorque
   else if pre(mode) == Backward then sa + tau0_max/unitTorque
   else sa - sign(w_relfric)*tau0_max/unitTorque;

The difference can be seen in the test example ModelicaTest.Rotational.TestFrictionPosition with a small number of output points ("number of intervals = 10" in Dymola).

Proposed solution
grafik

New solution:
grafik

tobolar:
> I have re-analyzed the solution and propose a slightly different fix.

@MartinOtter I have implemented your suggestion in https://github.com/m-kormann/ModelicaStandardLibrary/pull/1
@m-kormann Please merge https://github.com/m-kormann/ModelicaStandardLibrary/pull/1 in order to be visible here.

MartinOtter (approved):

PR #4257, opened 2024-01-12
Update Continuous.mo

hubertus65:
Better suggestion to fix init issue for Integrators in Modelica.Blocks.Continuous. WARNING: @MartinOtter, I currently don't have a setup to test this immediately, so this is UNTESTED. Please test before committing. It would be nice to have a test when this is in a loop with a non-linear plant + steady-state initialization.

### DIFF
--- a/Modelica/Blocks/Continuous.mo
+++ b/Modelica/Blocks/Continuous.mo
@@ -21,7 +21,7 @@ package Continuous "Library of continuous control blocks with internal states"
         Dialog(group="Initialization"));
     parameter Real y_start=0 "Initial or guess value of output (= state)"
       annotation (Dialog(group="Initialization"));
-    extends Interfaces.SISO;
+    extends Interfaces.SISO(y(start = if initType == Init.SteadyState then y_start), fixed = if initType == Init.SteadyState then false else true);
     Modelica.Blocks.Interfaces.BooleanInput reset if use_reset "Optional connector of reset signal" annotation(Placement(
       transformation(
         extent={{-20,-20},{20,20}},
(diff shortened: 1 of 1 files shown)


HansOlsson (changes requested):
This will not work for two reasons:
- The default for the start-attribute is now "empty", not 0.0 as in Modelica 3.4. Thus for non-steady-state there will be a new start-value, that may cause problems.
- There are also initial equations that weren't modified, so there will now be both a start-value with `fixed=true` and an initial equation - which is even more problematic.

For Dymola this is likely to be worse than before, since we could previously down-prioritize a start-value if there was an identical (possibly inactive) initial equation (even if not standardized it could at least be explained) - now they differ so we cannot do that.

It _may_ be possible to find a solution that only provides a non-fixed start-value when needed, I will check later.

MartinOtter (commented):
As Hans explained, your proposed solution is not good.

A (non-backwards) compatible good solution would be, as performed in Rotational.Components. Inertia, where no start or initial equations are defined at all, but "showStartAttribute=true" is provided for the relevant variables often used for initialization. This is much clearer and the user has a good GUI to select the most important initialization options (and then it is completely clear in the user interface which values are set with fixed = true, fixed = false or not set).

Since it seems overkill to make such a non-backwards compatible change now (which requires an appropriate, probably non-trivial, conversion script), I would suggest to keep the merged solution, but correct the description text for y_start:
   Real y_start= 0 "Initial value of output (= state), if initType=initialState or initialOutout)"
instead of the current
   Real y_start = 0 "Initial or guess value of output (= state)"

If the block appears in a nonlinear algebraic loop and y is an iteration variable, then, as usual, the user has to make a modifier (y(start=xxx)). One might also explain this in the documentation.

PR #4267, opened 2024-01-16
Add translation templates (new by Modelica Language Specification 3.6)

beutlich:
Now that we enabled (soft) MoLang 3.6 features, we can easily add the translation template files. Of course, they need to be updated each time before tagging a release. Therefore the strings are sorted by identifier to have it diff-friendly.

@HansOlsson @henrikt-ma Please check in Dymola and WSM, respectively, if the files are correctly generated.


beutlich:
@casella Feel free to update the milestone if this is not yet considered for MSL 4.1.0. I didn't want to miss the opportunity though what's possible with new MoLang 3.6.

tobolar:
> Of course, they need to be updated each time before tagging a release.

Not only the original english texts shall be updated. Ideally, also all translations. (So maybe we would need "Language officers" as well. :-D  )

And, in oder to populatize the multilingual feature, making public a routine for generation of pot-files from english documentation would be beneficial.

HansOlsson:
I don't think we should add it for this release (possibly except the file-settings):
- They are quite large files - increasing the size of the distribution.
- It doesn't add anything for users _in itself_.
- And I don't think we can realistically expect actual translations for _this release_.

BTW: Dymola 2024x can generate such templates with
`generateTranslationFile("Modelica", "Modelica.pot");`
And there's some way of reading them as well and (partially) using them - contact me if needed.

I guess we should document that more.
--
Additionally it is somewhat complicated to compare the files due to the sorting. I would prefer if we avoid that sorting.

I checked Complex and to me it seems as if the templates are based on an earlier version (e.g. "<blockquote><pre>c3 = sum(conj(c1[k]) * c2[k] for k in 1:n)\n" from `Complex.'*'.scalarProduct` was missing).

beutlich:
> BTW: Dymola 2024x can generate such templates with
> `generateTranslationFile("Modelica", "Modelica.pot");`

Nice!

When opening the generated POT file in Poedit translation tool I get a warning on duplicate items:

grafik

tobolar:
> BTW: Dymola 2024x can generate such templates with
`generateTranslationFile("Modelica", "Modelica.pot");`
And there's some way of reading them as well and (partially) using them - contact me if needed.

Nice. My first try here https://github.com/tobolar/Credibility/tree/multilingualDescriptions

beutlich:
> > BTW: Dymola 2024x can generate such templates with
> > `generateTranslationFile("Modelica", "Modelica.pot");`
> > And there's some way of reading them as well and (partially) using them - contact me if needed.
> 
> Nice. My first try here https://github.com/tobolar/Credibility/tree/multilingualDescriptions

Yes, that was also my assumption that both template and translations should be part of the repository.

henrikt-ma (commented):
Unfortunately, we don't have support for library translations yet in System Modeler, so I'm unable to try this out.

AHaumer:
Discussed this in today's MAP-Lib meeting, not realistic for 4.1.0 -> shift the milestone

PR #4268, opened 2024-01-16
Make third-party licenses explicit

beutlich:
I'd prefer to have the utilized third-party OSS licenses explicit rather than mixed-up with the MA related licenses.

Will also add it to https://github.com/modelica/ModelicaSpecification/pull/2900.


maltelenz (approved):
I don't see any problems with this.

casella (approved):
LGTM

Issue #4193, opened 2023-09-18
Inconsistent use of `tol` in `ModelicaTest.MultiBody.Forces.Damper`


qlambert-pro:
The variable used as tolerance margin to decide whether two `Real` values are equal is used to compare position and speed at the same time.

We could turn `tol` into a `constant` as we don't expect the infer the unit of constants based on their usage or maybe use `Modelica.Constants.eps`?

qlambert-pro:
Same issue in `ModelicaTest.MultiBody.Forces.Damper2`.

tobolar:
I don't understand the point of the issue. Can you explain in more detail? 

Regarding `ModelicaTest.MultiBody.Forces.Spring` and `ModelicaTest.MultiBody.Forces.Spring2`, there is a mess in using `tol`and `assert`. I will prepare a fix.

henrikt-ma:
This has to do with unit checking.  When unit inference is applied, the `tol` should get an inferred unit which is consistent with how it is used, but this gets problematic if it is used inconsistently unit-wise.  What @qlambert-pro said about using a constant instead of a parameter has to do with the current idea that unit inference in Modelica should not be allowed for constants, so that a constant with empty unit will behave just as a literal number.

tobolar:
@henrikt-ma Many thanks. I didn't fully follow the unit discussion.

Then I guess both suggestions (changing to `constant` and equal to `Modelica.Constants.eps`) are fine. 
(At least Dymola didn't indicated any issue.)

PR #4276, opened 2024-01-22
Fix assertions

tobolar:
Note: for `ModelicaTest.MultiBody.Forces.Spring`, this PR changes values of reference signals `prismatic1.s` and `prismatic1.v` over time. So regression check shall be broken => reference signals shall be recalculated/re-generated.

No such behaviour expected for `ModelicaTest.MultiBody.Forces.Spring2`.

Refs #4193

### DIFF
--- a/ModelicaTest/MultiBody.mo
+++ b/ModelicaTest/MultiBody.mo
@@ -3699,7 +3699,7 @@ a linear damper is connected here.
       extends Modelica.Icons.Example;
       inner MultiBody.World world
         annotation (Placement(transformation(extent={{-96,40},{-76,60}})));
-      parameter Real tol=1e-4;
+      constant Real tol=Modelica.Constants.eps;
       MultiBody.Parts.FixedTranslation fixedTranslation(
           animation=false, r={1,0,0})
         annotation (Placement(transformation(extent={{-26,40},{-6,60}})));
@@ -3734,14 +3734,14 @@ a linear damper is connected here.
         v(fixed=true),
         n={0,1,0},
         s(start=-1, fixed=true),
-        a(fixed=true, start=-0.81)) annotation (Placement(transformation(
+        a(fixed=true, start=9-world.g)) annotation (Placement(transformation(
             extent={{-10,-10},{10,10}},
             rotation=270,
             origin={-4,16})));
     equation
-      assert(body.r_0[2] - body.r_0[2] < tol,
+      assert(body.r_0[2] - body1.r_0[2] < tol,
         "Position of the bodies must be equal (less then tolerance)");
-      assert(body.v_0[2] - body.v_0[2] < tol,
+      assert(body.v_0[2] - body1.v_0[2] < tol,
         "Velocity of the bodies must be equal (less then tolerance)");
       connect(world.frame_b, fixedTranslation.frame_a) annotation (Line(
           points={{-76,50},{-26,50}},
--- a/ModelicaTest/MultiBody.mo
+++ b/ModelicaTest/MultiBody.mo
@@ -3787,7 +3787,6 @@ a linear damper is connected here.
       extends Modelica.Icons.Example;
       inner MultiBody.World world
         annotation (Placement(transformation(extent={{-96,40},{-76,60}})));
-      parameter Real tol=1e-4;
       MultiBody.Parts.FixedTranslation fixedTranslation(
           animation=false, r={1,0,0})
         annotation (Placement(transformation(extent={{-26,40},{-6,60}})));
@@ -3845,10 +3844,6 @@ a linear damper is connected here.
           animation=false, r={1,0,0})
         annotation (Placement(transformation(extent={{28,40},{48,60}})));
     equation
-      assert(body.r_0[2] - body.r_0[2] < tol,
-        "Position of the bodies must be equal (less then tolerance)");
-      assert(body.v_0[2] - body.v_0[2] < tol,
-        "Velocity of the bodies must be equal (less then tolerance)");
       connect(world.frame_b, fixedTranslation.frame_a) annotation (Line(
           points={{-76,50},{-26,50}},
           color={95,95,95},


HansOlsson (changes requested):
Apart from the previous proposal to have two different tolerances (rtol and vtol) there's another issue with the test:

It is still a one-sided test!

ModelicaTest/MultiBody.mo
```diff
@@ -3734,14 +3734,14 @@ a linear damper is connected here.
...
         n={0,1,0},
         s(start=-1, fixed=true),
-        a(fixed=true, start=-0.81)) annotation (Placement(transformation(
+        a(fixed=true, start=9-world.g)) annotation (Placement(transformation(
             extent={{-10,-10},{10,10}},
             rotation=270,
             origin={-4,16})));
     equation
-      assert(body.r_0[2] - body.r_0[2] < tol,
+      assert(body.r_0[2] - body1.r_0[2] < tol,
```
HansOlsson, line 3742:
```suggestion
      assert(abs(body.r_0[2] - body1.r_0[2]) < tol,
```
We should test that they are close enough.

ModelicaTest/MultiBody.mo
```diff
@@ -3734,14 +3734,14 @@ a linear damper is connected here.
...
         "Position of the bodies must be equal (less then tolerance)");
-      assert(body.v_0[2] - body.v_0[2] < tol,
+      assert(body.v_0[2] - body1.v_0[2] < tol,
```
HansOlsson, line 3744:
```suggestion
      assert(abs(body.v_0[2] - body1.v_0[2]) < tol,
```
And similarly.

ModelicaTest/MultiBody.mo
tobolar: This concerns far more examples. :-|
Will try to modify them all.

ModelicaTest/MultiBody.mo
```diff
@@ -2473,9 +2464,9 @@ menu (this sets \"fixed=false\" on parameter \"length\").
 
       extends Modelica.Icons.Example;
       parameter Real tol=1e-4;
-      SI.Force rod_f_diff[3]=rod1.frame_b.f - rod3.frame_b.f
+      SI.Force rod_f_diff[3]=abs(rod1.frame_b.f - rod3.frame_b.f)
```
HansOlsson, line 2467: I think I would have preferred to have this line (and the next similar one) unchanged and only have the abs in the assert-clause; it makes it easier to plot the difference.

ModelicaTest/MultiBody.mo
```diff
@@ -3734,14 +3740,14 @@ a linear damper is connected here.
         v(fixed=true),
         n={0,1,0},
         s(start=-1, fixed=true),
-        a(fixed=true, start=-0.81)) annotation (Placement(transformation(
+        a(fixed=true, start=9-world.g)) annotation (Placement(transformation(
```
tobolar, line 3743: @MartinOtter As I fixed the assert-clause below, it was triggered. So I changed `prismatic1.a.start` to far more precise value to fulfill the assertion. If this was not originally intended in the test model, then some other estimation shall be taken. But the original `start=-0.81` is too hight.

HansOlsson: I was a bit lost here. Having -0.81 as a magic number is weird, changing it to 9-world.g just seems like magic!
After some thinking and translating the model in Dymola I got that the model has the following parameters:

  ```
  parameter Modelica.Units.SI.TranslationalSpringConstant spring_c=10
    "Spring constant";
  parameter Modelica.Units.SI.Length spring_s_unstretched=0.1
    "Unstretched spring length";
  parameter Modelica.Units.SI.Position prismatic_s_start=-1
    "Relative distance between frame_a and frame_b";
  parameter Modelica.Units.SI.Mass m=1;
```

Adding them where appropriate allows us to write:
`a(fixed=true, start=spring_c*(-prismatic_s_start-spring_s_unstretched)/m - world.g)`

I don't know if there is an intuitive explanation for the formula, but at least it seems to reduce the magic (and not only does it work, it's also unit-correct).

**Review by HansOlsson — COMMENTED** [review]
As indicated I think that the Spring-formula now looks too much like magic.

File: ModelicaTest/MultiBody.mo (hunk 4 above)
Comment (tobolar) [reply ModelicaTest/MultiBody.mo:3835]: @HansOlsson Perfect, I like it! (Why didn't I just think of that.)

The question remains - was it an intention to cause the difference for the assert? Or is that precise value just fine now?

Comment (HansOlsson) [reply ModelicaTest/MultiBody.mo:3835]: I think the actual value is just fine now - but I would prefer replacing the magic 9 by a non-magic formula as described above. I don't think the assert was supposed to trigger.

I believe the history of the model went something like this:
1. Original model with code added in some order:
   * Originally `g` was 9.81 (something like that was common when I was in school, I think we used 9.82 in Lund - but Munich is south of here).
   * Based on that value for `g` and the spring as given above the start-value was given as `9-g` and evaluated, where 9 corresponds to `spring_c*(-prismatic_s_start-spring_s_unstretched)/m`.
   * A non-working assert was added, it didn't trigger so it was assumed to work!
2. Then `g`  was replaced by the official 9.8065, but since the assert was broken it wasn't detected.
3. And here the assert is being corrected.

**Review by HansOlsson — APPROVED** [review]
Looking good

File: ModelicaTest/MultiBody.mo [hunk 5]
```diff
@@ -3699,7 +3795,11 @@ a linear damper is connected here.
       extends Modelica.Icons.Example;
       inner MultiBody.World world
         annotation (Placement(transformation(extent={{-96,40},{-76,60}})));
-      parameter Real tol=1e-4;
+      parameter SI.Position rtol=1e-4;
+      parameter SI.Velocity vtol=1e-4;
+
+      final parameter SI.Position s_start = -1 "Initial relative distance between world frame and body";
```
henrikt-ma, line 3801: May I ask why the parameter is made final, but not making the modifications using it final?

tobolar: No particular reason for this. I just didn't touch the modifiers. The parameter itself is final just to retain its value.

PR #4277, opened 2024-01-22
Declare tolerance as MSL constant

### LINKED ISSUE #4193: Inconsistent use of `tol` in `ModelicaTest.MultiBody.Forces.Damper`

tobolar:
Close #4193

### DIFF
--- a/ModelicaTest/MultiBody.mo
+++ b/ModelicaTest/MultiBody.mo
@@ -3890,7 +3890,7 @@ a linear damper is connected here.
       extends Modelica.Icons.Example;
       inner MultiBody.World world
         annotation (Placement(transformation(extent={{-120,20},{-100,40}})));
-      constant Real tol=Modelica.Constants.eps;
+      constant Real tol=1e-4;
       MultiBody.Forces.Damper damper(d=10) annotation (
           Placement(transformation(
             extent={{-10,-10},{10,10}},
@@ -4078,7 +4078,7 @@ a linear damper is connected here.
       extends Modelica.Icons.Example;
       inner MultiBody.World world
         annotation (Placement(transformation(extent={{-120,20},{-100,40}})));
-      constant Real tol=Modelica.Constants.eps;
+      constant Real tol=1e-4;
       MultiBody.Forces.Damper damper(d=310) annotation (
           Placement(transformation(
             extent={{-10,-10},{10,10}},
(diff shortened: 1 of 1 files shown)


henrikt-ma (commented):
The change from 1e-4 to `Modelica.Constants.eps` is massive.  I made a small test with System Modeler, and to my big surprise the asserts were not triggered, despite the new tolerance being extremely tight.  I hope that someone with a deeper understanding of the dynamics here can approve the change to such tight tolerances.

What about all the other test models in `ModelicaTest.MultiBody` that still use 1e-4?

MartinOtter (commented):
I am strongly against this change.
Several different modeling ways are compared in a simulation. Signals are computed in the order of the relative tolerance, so the correct way would be something like tol = 10*relativeTolerance. However, this cannot be expressed in Modelica. 1e-4 is far good enough as an approximation

tobolar:
> Several different modeling ways are compared in a simulation. Signals are computed in the order of the relative tolerance, so the correct way would be something like tol = 10*relativeTolerance. However, this cannot be expressed in Modelica. 1e-4 is far good enough as an approximation

Ok. The tolerance is set again to 1e-4.

henrikt-ma (changes requested):
There are now lots of unrelated changes in graphic coordinates.  Please remove.

tobolar:
> There are now lots of unrelated changes in graphic coordinates. Please remove.

Sorry for this. Fixed now.

HansOlsson (commented):
I'm not sure this is a good idea.
I understand that we want to be less strict with unit-checking for such constants, and it matches that. (However, it is not decided yet so no need to rush.)

But if the test fails it does make sense to experiment with changing the tolerance, and thus having a constant seems problematic. (Except that the assert test is completely broken.)

Thus to me it would make sense to instead have two tolerances in the model: rtol (position) and vtol (velocity), even if they have the same value.
This is similar to the idea in #4051 to have different sources.

tobolar:
> Thus to me it would make sense to instead have two tolerances in the model: rtol (position) and vtol (velocity), even if they have the same value.

I can change this, retaining the same values. Reasonable also to give them units (position, velocity)?

And, which milestone shall we strive?

henrikt-ma:
> I can change this, retaining the same values. Reasonable also to give them units (position, velocity)?

Yes, this will ensure unit-awareness when viewing/modifying the parameters in context where units determined by inference are not available.

PR #4282, opened 2024-01-24
Add missing macro definitions

### LINKED ISSUE #4280: C code uses macros specified in internal/hidden ModelicaUtilities.h
The external C code in MSL uses macros. Example:

https://github.com/modelica/ModelicaStandardLibrary/blob/d0ff9bceb68621e4fc56f44a335e403947fd3c78/Modelica/Resources/C-Sources/ModelicaInternal.c#L150

These macros are defined in some `ModelicaUtilities.h` which has to be assumed is internal since it is in the `.CI` folder:

https://github.com/modelica/ModelicaStandardLibrary/blob/d0ff9bceb68621e4fc56f44a335e403947fd3c78/.CI/Test/ModelicaUtilities.h#L63-L127

However, as a tool vendor, I expect to provide my own `ModelicaUtilities.h`, which fulfills only the requirements given in the specification for such a file.

Should these macro definitions be moved out to a different `.h` in `Modelica/Resources/C-Sources`?

beutlich:
Closes #4280.

### DIFF
--- a/.CI/Test/ModelicaUtilities.h
+++ b/.CI/Test/ModelicaUtilities.h
@@ -31,14 +31,11 @@
 
 /* Utility functions which can be called by external Modelica functions.
 
-   These functions are defined in section 12.8.6 of the
-   Modelica Specification 3.0 and section 12.9.6 of the
-   Modelica Specification 3.1 and later.
+   These functions are defined in section 12.9.6 of the Modelica Specification 3.6.
 
    A generic C-implementation of these functions cannot be given,
    because it is tool dependent how strings are output in a
-   window of the respective simulation tool. Therefore, only
-   this header file is shipped with the Modelica Standard Library.
+   console or window of the respective simulation tool.
 */
 
 #ifndef MODELICA_UTILITIES_H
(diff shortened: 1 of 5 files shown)


beutlich:
Should we add the undefines to the other header files as well?

maltelenz (commented):
With this, I was able to build it cleanly (on Linux) with our vendor-provided `ModelicaUtilities.h`.

Should all of the define stuff be removed from `.CI/Test/ModelicaUtilities.h`, to more closely resemble a vendor-provided `ModelicaUtilities.h`, and avoid accidental reliance on these defines?

This comment in the `ModelicaUtilities.h` is not correct:
```
   A generic C-implementation of these functions cannot be given,
   because it is tool dependent how strings are output in a
   window of the respective simulation tool. Therefore, only
this header file is shipped with the Modelica Standard Library.
```
but to be fair, this PR didn't touch that part.

beutlich:
> This comment in the `ModelicaUtilities.h` is not correct:

It's updated now.

> Should all of the define stuff be removed from `.CI/Test/ModelicaUtilities.h`,

I prefer to have it as is. After all, it is test code only.

> Should we add the undefines to the other header files as well?

From what I checked, it is fine as it is.

HansOlsson (approved):

maltelenz (approved):

PR #4286, opened 2024-01-27
Fix warning: 'noreturn' function does return (CI)

beutlich:
This warning was observed when getting clean builds for #4265.


maltelenz (approved):

HansOlsson (approved):
Ok. And hope that no-one uses something like this in actual production code.

Harisankar-Allimangalath:
@maltelenz we have to backport this to the maint/4.1.0 right ?

beutlich:
This is test code only and not that relevant for back-porting.

PR #4294, opened 2024-01-31
Fix incorrect units in icons of to_cm2 and from_cm2

henrikt-ma:
Addressing user complaint.  As far as I can tell, the old placement of the unit text still works fine.


tobolar (approved):
Fine to me.

HansOlsson (approved):
Ok

PR #4348, opened 2024-02-28
Update CoupledClutches.mos

DagBruck:
Updated to Dymola command set from this millennium.

### DIFF
--- a/Modelica/Resources/Scripts/Dymola/Mechanics/Rotational/CoupledClutches.mos
+++ b/Modelica/Resources/Scripts/Dymola/Mechanics/Rotational/CoupledClutches.mos
@@ -1,6 +1,4 @@
-// Script generated: 06/21/00 15:01:45
+// Script updated for MSL 4.1.0
 
-translateModel("Modelica.Mechanics.Rotational.Examples.CoupledClutches")
-experiment StopTime=1.2
-simulate
+simulateModel("Modelica.Mechanics.Rotational.Examples.CoupledClutches", stopTime=1.2)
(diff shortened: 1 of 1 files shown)


dietmarw (approved):
LGTM

beutlich:
Would be better to merge to master and cherry-pick after merge.

AHaumer (approved):
ok
as @beutlich said, merge to master, then to maint\4.1.0

tobolar (approved):
Fine to me.

DagBruck:
It would help to have some document that describes "best practices". Git permits many different ways to use the system, branching, etc. I'm happy to march along, if somebody just points in the right direction.

DagBruck:
Thats for the pointer @dietmarw but I must admit I was confused by what it says:

> **Maintenance branch**
> 
> Name: "maint/4.0.x"
> 
> This branch contains the released Modelica Standard Library version (e.g., v4.0.0) where all bug-fixes since this release date are included (also consecutive BUGFIX versions 4.0.1, 4.0.2, etc., up to when a new MINOR or MAJOR release becomes available; i.e., there will not be any further BUGFIX versions (i.e., 4.0.x) of a previous release).

The way I read it, the maintenance branch is created after the new release has been created, and then you can apply bug fixes to 4.0.x versions. Is that correct?

What makes me confused is that we have `maint/4.1.0` define now, but 4.1 has not been released.

dietmarw:
Well naming the branch `4.1.0` is simply **wrong**. It should be `4.1.x`

DagBruck:
@HansOlsson suggested  I create a new pull request for the master branch: #4350.

PR #4378, opened 2024-04-02
Utilize constant EPSSIL in Modelica.Electrical.Spice3.Internal.Mos1.mos1ModelLineParamsInitEquations

beutlich:
Already utilized in other place.


christiankral (approved):

beutlich:
> As noted in #4377 we might also document that epssil means epsilon (di-electric constant) for Silicon, and add a proper unit.

I updated the descriptions accordingly but will change the units of EPSSIL and EPSSOX in a separate PR once this PR gets merged, because this PR does not introduce any simulation relevant changes.

PR #4390, opened 2024-04-16
Fixes #4389

### LINKED ISSUE #4389: Wrong input type in specificEnthalpy_pTX, temperature_phX, and density_phX
Consider this simple MWE:
```modelica
model TestFunctionDefault
  package Medium = Modelica.Media.Air.SimpleAir;
  Medium.SpecificEnthalpy h = Medium.specificEnthalpy_pT(1e5, 300);
end TestFunctionDefault;
```
This model compiles in Dymola 2024X and produces the expected result. However, when I try to compile it with OpenModelica, I get this error:
```
[1] 18:31:01 Translation Error
[Modelica.Media: 4706:7-4709:26]: Type mismatch for positional argument 3 in TestMedia.TestFunctionDefault.Medium.specificEnthalpy_pTX(X=fill(0.0, 0)). The argument has type:
  Integer[0]
expected type:
  Real[1]
```
This output looks a bit odd at first sight, what's wrong in calling `specificEnthalpy_pT(1e5, 300)`? I thought it was an OMC bug, but in fact it turns out to be an MSL issue.

`specificEnthalpy_pT`, as defined in [PartialPureSubstance](https://github.com/modelica/ModelicaStandardLibrary/blob/a877f527e29c47e344d2c05e393b65c6c4eee01b/Modelica/Media/package.mo#L4599), calls `specificEnthalpy_pTX(p, T, fill(0, 0))`, which is defined in [PartialMedium](https://github.com/modelica/ModelicaStandardLibrary/blob/a877f527e29c47e344d2c05e393b65c6c4eee01b/Modelica/Media/package.mo#L4446) as `specificEnthalpy(setState_pTX(p, T, X))`, so it calls `setState_pTX` with a third argument `X = fill(0, 0)`.

Unfortunately, setState_pTX for SimpleAir is defined in [PartialSimpleIdealGasMedium](https://github.com/modelica/ModelicaStandardLibrary/blob/a877f527e29c47e344d2c05e393b65c6c4eee01b/Modelica/Media/package.mo#L6085) as:
```modelica
  redeclare function specificEnthalpy_pTX
    "Return specific enthalpy from p, T, and X or Xi"
    input AbsolutePressure p "Pressure";
    input Temperature T "Temperature";
    input MassFraction X[nX] "Mass fractions";
    output SpecificEnthalpy h "Specific enthalpy at p, T, X";
  algorithm
    h := cp_const*(T - T0);
  end specificEnthalpy_pTX;
```
with input `MassFraction X[nX]` instead of `MassFraction X[:]`. Since by default `nX = size(substanceNames, 1) = 1`, it then expects an array of dimension one, but instead gets an array of dimension zero, coming from `fill(0,0)`.

I'm not sure why Dymola does not complain about this issue, but for sure it needs to be fixed. This holds for functions `specificEnthalpy_pTX`, `temperature_phX`, and `density_phX` in both `PartialSimpleMedium` and `PartialSimpleIdealGasMedium`.

casella:
See #4389.

### DIFF
--- a/Modelica/Media/package.mo
+++ b/Modelica/Media/package.mo
@@ -6082,7 +6082,7 @@ Note that the (small) influence of the pressure term p/d is neglected.
       extends Modelica.Icons.Function;
       input AbsolutePressure p "Pressure";
       input Temperature T "Temperature";
-      input MassFraction X[nX] "Mass fractions";
+      input MassFraction X[:] "Mass fractions";
       output SpecificEnthalpy h "Specific enthalpy";
     algorithm
       h := cp_const*(T - T0);
@@ -6098,7 +6098,7 @@ This function computes the specific enthalpy of the fluid, but neglects the (sma
       extends Modelica.Icons.Function;
       input AbsolutePressure p "Pressure";
       input SpecificEnthalpy h "Specific enthalpy";
-      input MassFraction X[nX] "Mass fractions";
+      input MassFraction X[:] "Mass fractions";
       output Temperature T "Temperature";
     algorithm
       T := T0 + h/cp_const;
@@ -6108,7 +6108,7 @@ This function computes the specific enthalpy of the fluid, but neglects the (sma
       extends Modelica.Icons.Function;
       input AbsolutePressure p "Pressure";
       input SpecificEnthalpy h "Specific enthalpy";
-      input MassFraction X[nX] "Mass fractions";
+      input MassFraction X[:] "Mass fractions";
       output Density d "Density";
     algorithm
       d := density(setState_phX(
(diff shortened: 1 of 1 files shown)


beutlich (commented):
Would be good to also have a test model for all affected function in ModelicaTest.Media.TestOnly to not break it again accidentically in the future.

HansOlsson (approved):
Seems ok.

However, the underlying design is a bit odd since setState_phX is defined as taking an input of size nX or nX-1.

But it seems too late to redesign that and I don't think a syntax such as `input MassFraction X[nX or nX-1];` would be helpful!

hubertus65 (approved):

Esther-Devakirubai:
Backporting to maintenance branch by #4453

PR #4393, opened 2024-04-19
Make type explicit for fill operator

beutlich:
This is derived from #4389. It helps tools to correctly deduce the array type, just by checking the argument of fill.

Note, that it already was a mixed usage of Integer zero and Real zero in Media functions.

### DIFF
--- a/Modelica/Media/package.mo
+++ b/Modelica/Media/package.mo
@@ -2819,7 +2819,7 @@ points, e.g., when an isentropic reference state is computed.
       s2 = Medium.specificEntropy_pTX(
               p,
               Ts,
-              fill(0, 0));
+              fill(0.0, 0));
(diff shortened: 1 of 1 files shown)


casella (approved):
I'm not sure I see the point of this PR, since all these functions have a third input which is a subtype of Real; as Modelica defines automatic conversions from Integer to Real, is there any difference?

Besides, the default is actually an empty array.

That said, I don't see anything wrong with this change, even if I don't get the point 😃

casella (approved):
LGTM

casella:
@Harisankar-Allimangalath can you please review this PR so that it gets merged in?

@Esther-Devakirubai this can also be ported to maint/v4.1.x. Thanks!

arunkumar-narasimhan (approved):

Esther-Devakirubai:
Backporting this to maint 4.1.x by #4462

PR #4406, opened 2024-05-22
Skip NULL checks for passed external objects

beutlich:
This is in line with https://github.com/modelica-3rdparty/Modelica_DeviceDrivers/issues/153 and specification change/clarification for https://github.com/modelica/ModelicaSpecification/issues/3500.

I labeled with `requires Modelica 3.7`, but I am actually not sure.

### DIFF
--- a/Modelica/Resources/C-Sources/ModelicaStandardTables.c
+++ b/Modelica/Resources/C-Sources/ModelicaStandardTables.c
@@ -196,6 +196,7 @@
 #include "uthash.h"
 #include "gconstructor.h"
 #endif
+#include <assert.h>
 #include <float.h>
 #include <math.h>
 #include <string.h>
@@ -870,9 +871,7 @@ void* ModelicaStandardTables_CombiTimeTable_init3(_In_z_ const char* fileName,
 
 void ModelicaStandardTables_CombiTimeTable_close(void* _tableID) {
     CombiTimeTable* tableID = (CombiTimeTable*)_tableID;
-    if (NULL == tableID) {
-        return;
-    }
+    assert(NULL != tableID);
     if (NULL != tableID->table && tableID->source == TABLESOURCE_FILE) {
 #if defined(TABLE_SHARE) && !defined(NO_FILE_SYSTEM)
         if (NULL != tableID->key) {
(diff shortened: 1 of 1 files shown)


Modelica/Resources/C-Sources/ModelicaStandardTables.c
```diff
@@ -865,9 +866,7 @@ void* ModelicaStandardTables_CombiTimeTable_init3(_In_z_ const char* fileName,
 
 void ModelicaStandardTables_CombiTimeTable_close(void* _tableID) {
     CombiTimeTable* tableID = (CombiTimeTable*)_tableID;
-    if (NULL == tableID) {
-        return;
-    }
+    assert(NULL != tableID);
```
henrikt-ma, line 869: Just wondering if simply `assert(tableID)` could be the more idiomatic way of checking that a pointer in C  is not null?

maltelenz (approved):
Seems fine to me.

I would say Modelica 3.7 is not required.

Modelica/Resources/C-Sources/ModelicaStandardTables.c
beutlich: Also possible. I prefer it explicit (and not implicitly casted to boolean).

henrikt-ma: I used to have the same preference, but many years I changed position because I felt I was working against the language design.  The top answer here covers my position rather well:
https://stackoverflow.com/questions/3825668/checking-for-null-pointer-in-c-c

beutlich: I also read there:

> Having a consistent style is sometimes more important than which style you pick.

This should be the case here. Let's consider it out of scope for this PR.

HansOlsson (approved):
Seems ok.
However, an alternative would be to have the assert first and then keep the previous checks - that would be slightly more robust.

I don't know if the NULL pointer checks are that costly.

beutlich:
We have it in M_DD without NULL checks for years. Thus it rather is an issue of code style than technical necessity.

Modelica/Resources/C-Sources/ModelicaStandardTables.c
henrikt-ma: That's right.  I didn't think of the numerous other non-null tests we already have.  Resolving.

beutlich:
> @casella @beutlich Should this be backported to maint/4.1.0?

No, not needed.

PR #4411, opened 2024-06-03
Add top-level License annotation

beutlich:


Complex.mo
```diff
@@ -278,6 +278,6 @@ Copyright &copy; 2010-2024, Modelica Association and <a href=\"modelica://Modeli
           radius=25.0), Text(
           textColor={255,255,255},
           extent={{-90,-50},{90,50}},
-          textString="C")}));
-
+          textString="C")}),
+License="modelica:/Modelica/Resources/Licenses/LICENSE_ModelicaStandardLibrary.txt");
```
tobolar, line 282:
```suggestion
    License="modelica:/Modelica/Resources/Licenses/LICENSE_ModelicaStandardLibrary.txt");
```

beutlich: Indentation is bad anyway when you check the entire annotation.

tobolar:
In the Modelica Spec v3.7-dev, there is both the annotation *License* and an annotation record of the same name (as a part of the *Protection* record). I'm confused...

Anyway, the suffix "-dev" feels like a development version of Spec. So why to implement it in MSL already now?

beutlich:
Yes, that's why it's labeled [`requires Modelica 3.7`](https://github.com/modelica/ModelicaStandardLibrary/labels/requires%20Modelica%203.7).

tobolar:
What I mean is that a development version (of Spec) is not a release and can still change. Maybe unlikely for the *License* annotation but who knows.

beutlich:
> What I mean is that a development version (of Spec) is not a release and can still change. Maybe unlikely for the _License_ annotation but who knows.

This just seems too unlikely given the voting results:

> Voting on MCP-0029 has ended.
>
> 100% are in favor.

See mail to MD from Fri, 10 May 2024 14:35:17 +0200.

PR #4420, opened 2024-06-18
Remove unreliable signals very close to zero

casella:
These signals only depend on numerical approximations and CSV compare cannot evaluate them reliably because it lacks information about their magnitude (nominal attribute).


casella:
Once approved, this should be cherry-picked to maint/v4.1.0 as well.

HansOlsson (commented):
I would very much prefer if we could update the comparison program instead.

AHaumer (approved):
I'm not a specialist in MultiBody but for me it looks reasonable.

casella:
@HansOlsson the more I think of it the more I see you point. These signals are supposed to be near zero, and if we don't check they may (in principle) stray away.

One could argue that they very likely can't do that without some of the non-zero components changing significantly, but at the end of the day this issue can be really fixed by just adding some sensible absolute tolerance to the CVS-compare tool.

I'll keep this on hold for the time being.

maltelenz:
I converted it to a draft PR to reflect the previous comment and avoid accidental merging.

PR #4426, opened 2024-06-19
Rebuild binaries

beutlich:
Rebuilt required after merging #4417.

Built from 715668b515099568a99de62f4dd3706f4a33555a with

  * MSVC VS 2005
  * GCC (TDM64-1) 5.1.0
  * GCC (SUSE Linux) 4.8.5


casella (approved):
Assuming the process was the same of the previous build, this should be OK.

HansOlsson (approved):
Seems good. As expected the sizes are fairly similar.

PR #4490, opened 2024-10-27
Revise CTest configuration

beutlich:
* `include(CTest)` dependency is not required and considered controversial because it introduces extra build targets. Instead `enable_testing()` is good enough. I read about it [here](https://discourse.cmake.org/t/is-there-any-reason-to-prefer-include-ctest-or-enable-testing-over-the-other/1905) and [here](https://gitlab.kitware.com/cmake/cmake/-/issues/21730) from the CMake experts, for example.
* Introduce `MODELICA_BUILD_TESTING:BOOL=ON` as a dedicated option to build the tests. 
* Add folder property globally and additionally to test targets to achieve better structuring in VS.

### DIFF
--- a/Modelica/Resources/BuildProjects/CMake/options.cmake
+++ b/Modelica/Resources/BuildProjects/CMake/options.cmake
@@ -21,3 +21,11 @@ option(
 
 # Option whether to install ModelicaExternalC
 option(MODELICA_INSTALL_EXTC "Install ModelicaExternalC library" OFF)
+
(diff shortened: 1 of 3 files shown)


casella:
@beutlich do you have suggestions for reviewers?

maltelenz (approved):

beutlich:
@casella How to proceed which PRs where hardly anyone feels comfortable to review it?

HansOlsson (approved):

PR #4492, opened 2024-10-27
Remove dead link to old OM wiki

beutlich:


casella (approved):

casella:
In fact, all the information in that section of the documentation is completely obsolete, and should be either updated or removed

casella:
I guess this should also be backported to maint/4.1.0

casella:
In fact, changing history a-posteriori is kind of weird, but I guess in this case we have to.

casella:
@beutlich would you mind back-porting this to maint/4.1.x? Thanks!

beutlich:
> @beutlich would you mind back-porting this to maint/4.1.x? Thanks!

See #4501.

PR #4495, opened 2024-10-27
Add function Math.nearestInteger

beutlich:
Cherry-picked from #3247 as proposed by https://github.com/modelica/ModelicaStandardLibrary/pull/3247#pullrequestreview-1001948116.


casella (approved):
LGTM

Modelica/Blocks/Math.mo
```diff
@@ -1939,7 +1939,7 @@ Otherwise the input angle <code>u</code> is wrapped to the <a href=\"https://en.
     Interfaces.IntegerOutput y "Connector of Integer output signal" annotation (
        Placement(transformation(extent={{100,-10},{120,10}})));
   equation
-    y = if (u > 0) then integer(floor(u + 0.5)) else integer(ceil(u - 0.5));
+    y = Modelica.Math.nearestInteger(u);
```
henrikt-ma, line 1942: This should be rejected since `Modelica.Math.nearestInteger(u)` is a discrete-time `Integer`.

henrikt-ma (changes requested):
Needs more thought regarding event generation.

Modelica/Blocks/Math.mo
henrikt-ma: Even if the function had `GenerateEvents = true`, that wouldn't change the expression variability of the result, right?  Hence, isn't the only way to solve this to ensure that all inputs passed to the user-defined function are discrete-time?  For instance, one could pass both `integer(floor(u + 0.5))` and `integer(ceil(u - 0.5))` to a user-defined function which would be able to return a discrete-time `Integer`, but that function wouldn't be an abstraction worthy a place in the MSL, in my opinion.

henrikt-ma (changes requested):
(Just restoring my _Request changes_ after changing my mind back and forth.)

Modelica/Blocks/Math.mo
HansOlsson: > This should be rejected since `Modelica.Math.nearestInteger(u)` is a discrete-time `Integer`.

I assume you mean "non-discrete time".
Note that this is true even if the function would have GenerateEvents (at least currently).

So, the proper solution seems to be to have GenerateEvents influence variability in the specification, and then it might be legal - but that's too late for MSL 4.1.0.

henrikt-ma: Yes, I'll edit the original comment above to avoid confusion.

beutlich: I rebased and skipped that bad commit.

HansOlsson (commented):
The general idea is good, but when using it outside of functions it becomes messy.

It seems we need the annotation `GenerateEvents=true` _and_ update such functions in terms of variability, e.g., along the lines of https://github.com/modelica/ModelicaSpecification/pull/3610

beutlich:
> The general idea is good, but when using it outside of functions it becomes messy.

I removed the usage in block `Modelica.Blocks.Math.RealToInteger`. @casella @HansOlsson Please add your review again.

Modelica/Math/nearestInteger.mo
```diff
@@ -0,0 +1,44 @@
...
+Math.nearestInteger(-0.5);                    // = -1
+</pre></blockquote>
+
+<h4>Note</h4>
+
+<p>
+This function does the same conversion as the block
+<a href=\"modelica://Modelica.Blocks.Math.RealToInteger\">RealToInteger</a>.
+</p>
+</html>"));
```
HansOlsson, line 43:
```suggestion
</html>"), GenerateEvents=true);
```
Regardless of https://github.com/modelica/ModelicaSpecification/pull/3610 this makes sense to me.

HansOlsson (approved):
LGTM.
Combined with https://github.com/modelica/ModelicaSpecification/pull/3610 it will be generally useful.

One could possibly add "away from zero" already in description as the preferred method depends, https://en.wikipedia.org/wiki/Rounding (But it may also be too long.)

HansOlsson (approved):

PR #4500, opened 2024-11-05
Ensure that tests (via CTest) are built and run (CI)

beutlich:


casella:
@beutlich what problem does this PR address? Any suggestion for reviewers?

beutlich:
The ctest option [--no-tests](https://cmake.org/cmake/help/latest/manual/ctest.1.html#cmdoption-ctest-no-tests)=error was added to not silently accept CI builds with `-DBUILD_TESTING=OFF` (`ON` is default) (which happened to my accidently when working on CMake configs of other reprositories I was contributing).

casella (approved):
LGTM

maltelenz (approved):
Without having any deeper insight into how CI is used in this project, this seems sensible, and I trust @beutlich to know what is needed.

Issue #2849, opened 2019-03-07
Improve LAPack deployment with MSL


beutlich:
As discussed with @hubertus65 at the Regensburg design meeting:

MSL is currently not self-contained to the LAPack sources and binaries. There only is a note in the readme, that each tool needs to support LAPack.

https://github.com/modelica/ModelicaStandardLibrary/blob/ea28ae3e29bf0ed9c68189100292a0001151143e/Modelica/Resources/C-Sources/readme.txt#L51-L54

beutlich:
Obviously, most Modelica tool-vendors worked-around this issue of a non-self-contained MSL w.r.t. LAPACK sources and binaries.

PR #4513, opened 2024-12-28
Add License annotation to external functions

beutlich:
This PR comes along with #4411 (after merging Mo-2900) and makes the license texts of the external functions/objects explicit such that FMU exporting tools no longer need to distribute all files of `modelica:/Modelica/Resources/Licenses/` along with the FMU, but only the actually used license texts.

Edit: Dependency on lapack (and the corresponding license texts) still is an exception since it is not seen as MSL dependency, but as tool requirement (see #2849).

Edit 2: Needs to be aligned with #4240.

## Summary by CodeRabbit

- **Chores**
  - Added explicit license information to external C function annotations across multiple packages, enhancing metadata about third-party and internal library usage.
  - No changes to function signatures, logic, or user-facing behavior; only license references were added for greater transparency regarding external dependencies.


HansOlsson (approved):
LGTM

PR #4514, opened 2025-01-02
Bump copyright year to 2025

beutlich:


HansOlsson (commented):
I don't think this is needed in general, and we likely should reconsider how we list the years to reduce the work in the future.

First and foremost copyright according to the Berne convention, https://en.wikipedia.org/wiki/List_of_copyright_duration_by_country is 50 or more years (70 in EU, Australia, Canada, Japan, and the US) after the author's death; so I don't see that updating the publication year matters.

Additionally some claim that one should only bump the year when there are actual changes.

One possibility would be to remove copyright from non-top .mo files, but keep them in the top ones and in other files. For the other files we could either bump them now, bump them when changed, or remove the end-year.

beutlich:
You are probably right. I just did as we did the previous 25 years w/o much questioning.

maltelenz (commented):
I agree with @HansOlsson . Should we even go so far as to remove the year completely everywhere? As far as I can tell (I'm not a lawyer) it shouldn't be required?

I don't have any strong opinions here, but I'm in favor of things that reduce manual work.

beutlich:
> but I'm in favor of things that reduce manual work.

Honestly, it is not much manual work at all, just search and replace.

casella:
@beutlich we could merge this on master, where it will already be obsolete when we branch off 4.2.0. Or, we branch it on master and then back-port it, every time with two reviews and more discussion. Not sure.

beutlich:
Yes, please. Merge and backport.

nithish-selvan:
@beutlich  I can be a reviewer :). Looks good! Closed duplicate #4623

maltelenz (approved):
Consensus seems to be that we want to do it like this (at least for 4.1.0).

HansOlsson (approved):
Ok, but change it after 4.1.0 so that we don't have all of these copyright years that need update.

maltelenz:
> @maltelenz is the broken `html_documentation_checks` something to worry?

Probably not, but the failure (some cmake version requirement not fulfilled) means we don't get whatever checks `html_documentation_checks` usually performs.

@beutlich ?

nithish-selvan:
> > @maltelenz is the broken `html_documentation_checks` something to worry?
> 
> Probably not, but the failure (some cmake version requirement not fulfilled) means we don't get whatever checks `html_documentation_checks` usually performs.
> 
> @beutlich ?

It worked now.

maltelenz:
Hah, I guess the stupid AI bot fulfilled a purpose (by retriggering the check through its comment).

beutlich:
> @beutlich ?

Well, see #4606 (still being unresolved).

PR #4535, opened 2025-02-15
Restore and modernize ModelicaUtilities.h

beutlich:
Same as #4487, but

* targeting master branch
* adapting CMake configuration files
* modernizing noreturn function attribute/specifier as propsed by @HansOlsson


casella (approved):
LGTM, also after MAP-Lib discussion

casella:
@maltelenz please back-port this to maint/4.1.x as agreed during the MAP-Lib meeting

PR #4536, opened 2025-02-16
Fix typo

beutlich:
As reported by https://github.com/modelica/ModelicaStandardLibrary/actions/runs/13351228193/job/37288029202#step:5:50


tobolar (approved):
Fine to me.

PR #4543, opened 2025-02-25
Fixed Utilities functions of ReferenceMoistAir to also work with Xi

### LINKED ISSUE #4512: Issue with ReferenceMoistAir Beyond 500°C
Hi,

I recently faced an issue when using Modelica.Media.Air.ReferenceMoistAir.
When using the BaseProperties (so all volume based components, like the boundary or vessel), there is a "Index out of bound" issue happening only when the temperature is beyond 500°C.

I used Dymola 25x and MSL 4.0.0 (I have the same issue in Dymola 2020x with MSL 3.2.4).

Please find below a very simple example reproducing the error:
```mo
model BugReferenceMoistAir
  //The model just bug (Index out of bounds) when temperature reaches 500°C
  package Medium=Modelica.Media.Air.ReferenceMoistAir;
  Medium.BaseProperties medium;
equation 
  medium.p=1e5;
  medium.T=273.15+time*1000;
  medium.Xi=Medium.reference_X[1:Medium.nXi];

  annotation (uses(Modelica(version="4.0.0")));
end BugReferenceMoistAir;
```
And the error message given by Dymola:
> Error: The following error was detected at time: 0.5
>   Index out of bounds
>   Failed condition: (j >= 1) && (j <= a.dims[0])
> The stack of functions is:
> Modelica.Media.Air.ReferenceMoistAir.Utilities.h_dis_pTX
> Modelica.Media.Air.ReferenceMoistAir.Utilities.h_pTX
> Modelica.Media.Air.ReferenceMoistAir.Utilities.u_pTX(100000.0, medium.T, medium.Xi)

The error triggered by **Modelica.Media.Air.ReferenceMoistAir.Utilities.h_dis_pTX()** comes from the fact that the functions (like Modelica.Media.Air.ReferenceMoistAir.Utilities.u_pTX()) in **BaseProperties** are called using Xi instead of X. Indeed, the function Modelica.Media.Air.ReferenceMoistAir.Utilities.h_dis_pTX() that is triggering the error is using X[1] and X[2] (obviously not available when using only Xi).
I duplicated the package Modelica.Media.Air.ReferenceMoistAir, and I modified all calls in BaseProperties to use X instead of Xi and it solved the error.

I looked in different Medium.BaseProperties, and all calls to functions using Xi... maybe we should correct that ?

casella:
Fixes #4512 by avoiding to use X[2] and X_der[2] in all functions that get X[:] as input (could be either X or Xi).

### DIFF
--- a/Modelica/Media/Air/ReferenceMoistAir.mo
+++ b/Modelica/Media/Air/ReferenceMoistAir.mo
@@ -2820,7 +2820,7 @@ for region 2.
           if (T >= 773.15) then
             cp := X[1]*
               Modelica.Media.Air.ReferenceMoistAir.Utilities.IF97_new.cp_pT(pd,
-              T) + X[2]*Modelica.Media.Air.ReferenceAir.Air_Utilities.cp_pT(pl,
+              T) + (1 - X[1])*Modelica.Media.Air.ReferenceAir.Air_Utilities.cp_pT(pl,
               T) + Modelica.Media.Air.ReferenceMoistAir.Utilities.cp_dis_pTX(
                   p,
                   T,
@@ -2828,7 +2828,7 @@ for region 2.
           else
             cp := X[1]*
               Modelica.Media.Air.ReferenceMoistAir.Utilities.IF97_new.cp_pT(pd,
-              T) + X[2]*Modelica.Media.Air.ReferenceAir.Air_Utilities.cp_pT(pl,
+              T) + (1 - X[1])*Modelica.Media.Air.ReferenceAir.Air_Utilities.cp_pT(pl,
               T);
           end if;
         else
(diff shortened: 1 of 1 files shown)


HansOlsson (commented):
I find this quite confusing, even if correct.
Especially as there's also air - and the logic is in general that it's the last one (not 2) that can be generated.

I would thus prefer to either introduce an extra X-variable as 
X2:={X[1],1-X[1]};
Or just comment on the logic on the specific lines.

arunkumar-narasimhan:
Would this line634 require a change as well? This appears to be the only instance in the file that is using state.X[2], not explicitly used as X[2] however.
`    MM := 1/(state.X[1]/steam.MM + state.X[2]/dryair.MM);`

casella:
> Would this line634 require a change as well? This appears to be the only instance in the file that is using state.X[2], not explicitly used as X[2] however. ` MM := 1/(state.X[1]/steam.MM + state.X[2]/dryair.MM);`

This is not necessary. setState_xx() takes in either X or Xi but then fills state.X completely all time.

arunkumar-narasimhan (approved):

Esther-Devakirubai:
Backported by #4546

PR #4547, opened 2025-02-25
Fixes for #4424

### LINKED ISSUE #4424: Completely specify initial conditions for DynamicPipeEnergyConservationCheck and DynamicPipeEnergyConservationCheck2
The models 
```
ModelicaTest.Fluid.TestComponents.Pipes.DynamicPipeEnergyConservationCheck
ModelicaTest.Fluid.TestComponents.Pipes.DynamicPipeEnergyConservationCheck2
```
do not have fully specified initial conditions. These should be added to make sure that the initialization problem is not tool-dependent. This is probably also the reason why the first of the two models currently [fails](https://libraries.openmodelica.org/branches/master/ModelicaTest_trunk/files/ModelicaTest_trunk_ModelicaTest.Fluid.TestComponents.Pipes.DynamicPipeEnergyConservationCheck.err) in OpenModelica.

casella:
Added small pressure losses to avoid index reduction with fixed pressure boundaries. Changed to explicit steady state initialization to avoid unspecified initial conditions.

Fixes #4424

### DIFF
--- a/ModelicaTest/Fluid/TestComponents/Pipes/DynamicPipeEnergyConservationCheck2.mo
+++ b/ModelicaTest/Fluid/TestComponents/Pipes/DynamicPipeEnergyConservationCheck2.mo
@@ -25,7 +25,7 @@ model DynamicPipeEnergyConservationCheck2
     modelStructure=Modelica.Fluid.Types.ModelStructure.av_b,
     useInnerPortProperties=true)
     annotation (Placement(transformation(extent={{-10,40},{10,60}})));
-  inner Modelica.Fluid.System system(energyDynamics=Modelica.Fluid.Types.Dynamics.SteadyStateInitial)
+  inner Modelica.Fluid.System system(energyDynamics=Modelica.Fluid.Types.Dynamics.FixedInitial)
     annotation (Placement(transformation(extent={{-100,80},{-80,100}})));
   Modelica.Fluid.Sources.MassFlowSource_h boundary(
     nPorts=1,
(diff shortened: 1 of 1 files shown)


casella:
This works in Dymola but not in OpenModelica 😢

casella:
There is something fishy in the way the pressure loss models work, I'm getting much smaller pressure losse than I expected, which makes the model ill-posed. I'll investigate more.

casella:
In the meantime, I changed the parametrization of the pressure losses to use a quadratic model and get about 0.1 bar pressure loss. I also switched to FixedInitial to avoid convergence problems. This works both in Dymola and OMC. 

@maltelenz could you please check if it also works in WSM?

HansOlsson (approved):

maltelenz:
`ModelicaTest.Fluid.TestComponents.Pipes.DynamicPipeEnergyConservationCheck2` simulates to completion in WSM.

The new results do not match the previously suggested reference (I assume this is expected).

There are still transients in the beginning, I don't know if this is expected.

casella:
> `ModelicaTest.Fluid.TestComponents.Pipes.DynamicPipeEnergyConservationCheck2` simulates to completion in WSM.

Good.
 
> The new results do not match the previously suggested reference (I assume this is expected).

Yes, since I added some pressure losses the pressures are different. But that doesn't matter, since the point of this model is to run the asserts. I'll ask LTX to re-generate the reference results in #4424.

> There are still transients in the beginning, I don't know if this is expected.

Yes, it is. I tried steady-state initialization but I got into numerical trouble with OMC, so I though we could keep the same default behaviour we had previously with Dymola (i.e., fix all the states to their initial values), only make it explicit. There is a relaxation transient, then the asserts are only applied at the end. This is also improved by PR #4555.

So, you can approve this without problem. Thanks!

maltelenz (approved):
Works in System Modeler, and the results seem as expected per previous comment.

Esther-Devakirubai:
Backported by #4566

PR #4555, opened 2025-02-26
Add numerical tolerance to floating-point number comparison in ModelicaTest.Fluid.TestComponents.Machines.TestLinearPower

### LINKED ISSUE #4554: Asserts on Real values with test models are too tight
There are two tests in ModelicaTest.Fluid.TestComponents that have too tight asserts. One is Machines.TestLinearPower:

https://github.com/modelica/ModelicaStandardLibrary/blob/b4a4b0336f8759418e916f29491f14789d9db905/ModelicaTest/Fluid/TestComponents/Machines/TestLinearPower.mo#L7

which is a numerical hazard, because `p` is computed by a linear equation, so the actual computed value could be something like 23.000000000001 or 29.999999999999 or something like that, which would of course be OK but would trigger the assertion failure.

The other one is ModelicaTest.Fluid.TestComponents.Pipes.DynamicPipeEnergyConservationCheck2:

https://github.com/modelica/ModelicaStandardLibrary/blob/b4a4b0336f8759418e916f29491f14789d9db905/ModelicaTest/Fluid/TestComponents/Pipes/DynamicPipeEnergyConservationCheck2.mo#L128-L131

where it is not clear whether the second part of the assertions will ever be checked, since 
https://github.com/modelica/ModelicaStandardLibrary/blob/b4a4b0336f8759418e916f29491f14789d9db905/ModelicaTest/Fluid/TestComponents/Pipes/DynamicPipeEnergyConservationCheck2.mo#L148

casella:
As discussed during the MAP-Lib meeting of 24/02/2025, the first issue in #4554 is fixed by allowing a bit of numerical tolerance. The second issue is addressed by explicitly calling assert() only at the end of the simulation, when the steady-state has been reached.

### DIFF
--- a/ModelicaTest/Fluid/TestComponents/Machines/TestLinearPower.mo
+++ b/ModelicaTest/Fluid/TestComponents/Machines/TestLinearPower.mo
@@ -4,6 +4,6 @@ model TestLinearPower
   Modelica.Units.SI.Power p;
 equation
   p = Modelica.Fluid.Machines.BaseClasses.PumpCharacteristics.linearPower(V_flow_nominal={1,2}, W_nominal={3,5}, V_flow=11);
-  assert(p >= 23 and p <= 23, "Error in linearPower");
+  assert(abs(p - 23) < 1e-9, "Error in linearPower");
 annotation(experiment(StopTime=1));
 end TestLinearPower;
--- a/ModelicaTest/Fluid/TestComponents/Machines/TestLinearPower.mo
+++ b/ModelicaTest/Fluid/TestComponents/Machines/TestLinearPower.mo
@@ -4,6 +4,6 @@ model TestLinearPower
   Modelica.Units.SI.Power p;
 equation
   p = Modelica.Fluid.Machines.BaseClasses.PumpCharacteristics.linearPower(V_flow_nominal={1,2}, W_nominal={3,5}, V_flow=11);
-  assert(abs(p - 23) < 1e-9, "Error in linearPower");
+  assert(Modelica.Math.isEqual(p, 23.0, 1e-9), "Error in linearPower");
 annotation(experiment(StopTime=1));
 end TestLinearPower;
--- a/ModelicaTest/Fluid/TestComponents/Pipes/DynamicPipeEnergyConservationCheck2.mo
+++ b/ModelicaTest/Fluid/TestComponents/Pipes/DynamicPipeEnergyConservationCheck2.mo
@@ -126,6 +126,7 @@ equation
   total_A_V_B_a = pipeA_V_B.H_flows[1] + pipeA_V_B.m_flows[1]*pipeA_V_B.flowModel.vs[1]^2/2;
   total_A_V_B_b = pipeA_V_B.H_flows[end] + pipeA_V_B.m_flows[end]*pipeA_V_B.flowModel.vs[end]^2/2;
   when terminal() then
+    assert(time > 495, "Steady state was not yet reached at StopTime");
     assert(Modelica.Math.isEqual(total_AV_B_a, total_AV_B_b, 1), "Energy not conserved!");
     assert(Modelica.Math.isEqual(total_A_VB_a, total_A_VB_b, 1), "Energy not conserved!");
     assert(Modelica.Math.isEqual(total_AV_VB_a, total_AV_VB_b, 1), "Energy not conserved!");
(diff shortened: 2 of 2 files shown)


HansOlsson (approved):
LGTM

ModelicaTest/Fluid/TestComponents/Machines/TestLinearPower.mo
```diff
@@ -4,6 +4,6 @@ model TestLinearPower
   Modelica.Units.SI.Power p;
 equation
   p = Modelica.Fluid.Machines.BaseClasses.PumpCharacteristics.linearPower(V_flow_nominal={1,2}, W_nominal={3,5}, V_flow=11);
-  assert(p >= 23 and p <= 23, "Error in linearPower");
+  assert(abs(p - 23) < 1e-9, "Error in linearPower");
```
beutlich, line 7:
```suggestion
  assert(Modelica.Math.isEqual(p, 23.0, 1e-9), "Error in linearPower");
```

ModelicaTest/Fluid/TestComponents/Pipes/DynamicPipeEnergyConservationCheck2.mo
```diff
@@ -125,10 +125,12 @@ equation
   total_AV_VB_b = pipeAV_VB.H_flows[end] + pipeAV_VB.m_flows[end]*pipeAV_VB.flowModel.vs[end]^2/2;
   total_A_V_B_a = pipeA_V_B.H_flows[1] + pipeA_V_B.m_flows[1]*pipeA_V_B.flowModel.vs[1]^2/2;
   total_A_V_B_b = pipeA_V_B.H_flows[end] + pipeA_V_B.m_flows[end]*pipeA_V_B.flowModel.vs[end]^2/2;
-  assert(time < 500 or Modelica.Math.isEqual(total_AV_B_a, total_AV_B_b, 1), "Energy not conserved!");
-  assert(time < 500 or Modelica.Math.isEqual(total_A_VB_a, total_A_VB_b, 1), "Energy not conserved!");
-  assert(time < 500 or Modelica.Math.isEqual(total_AV_VB_a, total_AV_VB_b, 1), "Energy not conserved!");
-  assert(time < 500 or Modelica.Math.isEqual(total_A_V_B_a, total_A_V_B_b, 1), "Energy not conserved!");
+  when terminal() then
```
beutlich, line 128: I am not sure about this change.

* The previous intention was that steady-state is reached at 500 s. Now, if the simulation stop time is changed to a lower value, assumption might be wrong.
* In SimulationX `when terminal` might not be called in all solver conditions because by default a simulation run is meant to be continued from the reached stop-time.

Hence, I am reluctant with this proposed change.

beutlich: Would `when time >= 500 then` work instead?

ModelicaTest/Fluid/TestComponents/Machines/TestLinearPower.mo
casella: Yeah, I mean, thanks to inlining it's exactly the same, only with some added abstraction. Does it really matter?

ModelicaTest/Fluid/TestComponents/Pipes/DynamicPipeEnergyConservationCheck2.mo
casella: >     * The previous intention was that steady-state is reached at 500 s.

Indeed.

> Now, if the simulation stop time is changed to a lower value, assumption might be wrong.

Of course, but why should anyone change it? I mean, if we want to be super-pedantic we could add another assert in the `when terminal()` clause that makes sure enough time has elapsed, e.g. 

```modelica
assert(time >= 495, "Not enough time to reach steady-state");
```

This will make sure that the test actually fails if the simulation time is shortened. To play safe, I would avoid asserting `time >=500`, otherwise we're back to square one, i.e., we are no longer getting tolerance to floating point comparisons, which is the whole point of this PR. Note that the steady state is already reached at time = 450, so there is really no reason to be tight on that condition.

Would you like to add that to make the intention clearer?

>     * In SimulationX `when terminal` might not be called in all solver conditions because by default a simulation run is meant to be continued from the reached stop-time.

I'm not sure I understand what you mean. [MLS Sect. 3.7.5](https://specification.modelica.org/maint/3.5/operators-and-expressions.html#event-related-operators-with-function-syntax:

>  terminal()
>
>  Returns true at the end of a successful analysis.
>
>   [Hereby, terminal() ensures an event at the end of successful simulation.]

So, either the simulation is not successful, i.e. it fails (that would be a tool issue, if the tool cannot actually run the simulation for some reason), or it succeeds, and then terminal becomes true and the assert section must be executed. The final goal is to avoid the possibility that the model runs successfully while a wrong energy balance is not caught. Do I miss something?

> Would when time >= 500 then work instead?

No, that would in fact risk that the actual checking of the energy balances, _which is the whole point of this model_, is not triggered and terminates successfully
- because the simulation ends at time = 499.99999999999 because of solver rounding errors
- because someone changed StopTime to some value less than 500 (as per your previous comment)
even though the energy balances are actually not checked and may actually be wrong.

casella: @beutlich, I added an assert to make sure that enough time has elapsed. I'm still not sure about your concerns about when terminal(), but to me the semantics of that in the specification is crystal clear.

Please add your review so we can merge this in and move on with the release. Thanks!

casella:
@beutlich please check the resolved conversation. Is this PR OK for you now?

beutlich (approved):

Esther-Devakirubai:
Backported by #4581

PR #4562, opened 2025-03-04
Attempt to fix #4559 by increasing RonDiode and GoffDiode

### LINKED ISSUE #4559: Regression: SMPM_Braking in MSL v4.1.0-dev fails due to chattering in OpenModelica
Modelica.Magnetic.FundamentalWave.Examples.BasicMachines.SynchronousMachines.SMPM_Braking in MSL 4.0.0 [simulated successfully](https://libraries.openmodelica.org/branches/master/Modelica_4.0.0/files/Modelica_4.0.0_Modelica.Magnetic.FundamentalWave.Examples.BasicMachines.SynchronousMachines.SMPM_Braking.sim) in OpenModelica, with results matching the reference files.

It [fails instead](https://libraries.openmodelica.org/branches/master/Modelica_4.1.0-beta.om/files/Modelica_4.1.0-beta.om_Modelica.Magnetic.FundamentalWave.Examples.BasicMachines.SynchronousMachines.SMPM_Braking.sim) in MSL 4.1.0 with the following error message:

```
Chattering detected around time 0.523614405406..0.523625675045 (100 state events in a row with a total time delta less than the step size 0.0001). This can be a performance bottleneck.
The zero-crossing was: diodeBridge2mPulse.diode_n.idealDiode[2].s < 0.0
```

This may be a tool issue, but was apparently triggered by some change in the library at some point. I will do some bisection to figure out when it happened.

@maltelenz does it work for you in WSM?

casella:
This attempts to fix #4559 by making the system less brutally nonlinear when approaching zero voltage and current on the load towards the end of the transient.

### DIFF
--- a/Modelica/Magnetic/FundamentalWave/Examples/BasicMachines/SynchronousMachines/SMPM_Braking.mo
+++ b/Modelica/Magnetic/FundamentalWave/Examples/BasicMachines/SynchronousMachines/SMPM_Braking.mo
@@ -149,13 +149,13 @@ equation
   connect(starM.pin_n, grounding.p)
     annotation (Line(points={{-70,-10},{-70,5},{-70,20}}, color={0,0,255}));
   annotation (experiment(StopTime=0.8, Interval=1E-4, Tolerance=1e-06), Documentation(
-        info="<html>
+        info= "<html>
 <p>
 A synchronous machine with permanent magnets starts braking from nominal speed by feeding a diode bridge,
 which in turn feeds a braking resistor.
 Since induced voltage is reduced proportional to falling speed, the braking resistance is set proportional
 to speed to achieve constant current and torque.</p>
 
-<p>Default machine parameters are used</p>
+<p>Default machine parameters are used. The RonDiode and GoffDiode parameters of the diode bridge are both increased to 1e-4 to avoid numerical issues when the variable resistor voltage gets close to zero.</p>
 </html>"));
 end SMPM_Braking;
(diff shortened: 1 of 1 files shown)


maltelenz (approved):
As described in my [comment](https://github.com/modelica/ModelicaStandardLibrary/issues/4559#issuecomment-2700465223) this seems good from a "simulate nicely in WSM" standpoint.

I want to make clear that I have no expertise on the model itself or the domain.

casella:
> I want to make clear that I have no expertise on the model itself or the domain.

Sure. For that, we wait for @AHaumer's or @christiankral's input.

casella:
I also have a (slightly rusty by now) degree in Electronic Engineering _cum laude_ 😄

AHaumer:
@casella the whole discussing was a little bit misleading, there are 2 examples of SMPM braking using a rectifier:

- Modelica.Electrical.Machines.Examples.SynchronousMachines.SMPM_Braking
- Modelica.Magnetic.FundamentalWave.Examples.BasicMachines.SynchronousMachines.SMPM_Braking
Should we apply the same solution to the other example, too?

AHaumer (approved):
LGTM
but IMHO the same solution should be applied to  Modelica.Electrical.Machines.Examples.SynchronousMachines.SMPM_Braking

Esther-Devakirubai:
Backported by #4570

casella:
> @casella the whole discussing was a little bit misleading, there are 2 examples of SMPM braking using a rectifier:
> 
>     * Modelica.Electrical.Machines.Examples.SynchronousMachines.SMPM_Braking
>     * Modelica.Magnetic.FundamentalWave.Examples.BasicMachines.SynchronousMachines.SMPM_Braking

Good point. I got a regression on the second one, but if the two examples are similar it may make sense to also fix the other one. In fact, as I discussed [here](https://github.com/modelica/ModelicaStandardLibrary/issues/4559#issuecomment-2704012622), we could even think of changing the default values for the underlying switching supply models.

I think it's best to fix the regressions locally in 4.1.0 where they show up, and then consider changing the default in MSL 4.2.0. That requires some analysis of side effects and is not something I'd do in a rush.

@AHaumer do you agree?

AHaumer:
@casella these two examples are pretty much the same except for the machine models which are 100% compatible speaking if connectros and parameters (only the internal representation is different).

PR #4572, opened 2025-03-06
Same patch with Ron and Goff

AHaumer:
If @casella agrees after testing with OM and @maltelenz agrees after testing with WSM, we should get this into 4.1.0 
(means @Esther-Devakirubai please backport to 4.1.x and @GallLeo @MatthiasBSchaefer please re-generate reference results). 
This would help to achieve more stability for 4.1.0.


maltelenz (approved):
I ran it in WSM on one platform, and quantitatively the results look similar to the (old) reference, but of course they differ in the details. The largest change was in `capacitor.v`:

image

Approving with the usual caveat that I have no understanding of the model.

casella (approved):
LGTM

Esther-Devakirubai:
Backported by #4579

PR #4589, opened 2025-03-11
Fix TDM-GCC build and update binaries

beutlich:
f86bce097a317fdfd151f44d6dfc9056a80d81db requires new binaries.


beutlich:
Note that [402ea29](https://github.com/modelica/ModelicaStandardLibrary/pull/4589/commits/402ea29706c01c8a7f3d067814add90ee599e9e7) also should go to master branch.

casella (approved):
As decided during the last MAP-Lib meeting, we should update the binaries in the maintenance branch

casella:
@AHaumer, @maltelenz can you please review so we get this done in time for the .rc tagging next Tuesday? Thanks!

Issue #4639, opened 2025-05-08
Improvements to Modelica.Math.Vectors.interpolate


DagBruck:
I suggest the following fairly simple performance improvements. They should be fully backwards compatible.
1. Special handling of `nx=2`
   - This is a small change that avoids the entire double iteration to find an index.
   - This case is quite common.
2. Default `iLast` calculated using linear interpolation.
   - If `iLast` is not specified by the caller, then we do a linear search from the start. This is very inefficient for all but the smallest tables.
   - Calculating an initial guess value by linear interpolation is a a reasonable alternative.
   - As far as I can tell, there is very little active use of `iLast` so we often fall back to the inefficient default.
3. Remove assertion x2>x1.
   - The test is not very useful because it just tests the two points where our index search stopped. It does not test the input requirement that x is non-decreasing.
   - The test clutters the code.
 
I also considered two more extensive changes, but do not propose them.
- Replace the linear search by binary search. This has big potential, but the implementation is complicated because the abscissa may contain several identical values. The documentation even defines what should happen.
- Remove the iLast argument and the return value iNew. This would be a significant simplification, especially because the function would just return a single number, not a pair of numbers, Also, almost nobody seems to use iLast/iNew. This would obviously be a non-compatible change.

DagBruck:
Another issue with `interpolate` is that the documentation seems self-contradictory. The description of the function argument `x` says:

> Abscissa table vector (strict monotonically increasing values required)

The documentation later says (and it makes sense to me, in order to support a step change):

> If x has two or more identical values then interpolation utilizes the x-value with the largest index.

My proposal is to remove "strict" from the former.

PR #4641, opened 2025-05-12
Update Math.Vectors.interpolate

DagBruck:
Performance improvements as described in https://github.com/modelica/ModelicaStandardLibrary/issues/4639.

### DIFF
--- a/Modelica/Math/package.mo
+++ b/Modelica/Math/package.mo
@@ -497,10 +497,10 @@ can be provided as third argument of the function. Default is \"eps = 0\".
   function interpolate "Interpolate linearly in a vector"
     extends Modelica.Icons.Function;
     input Real x[:]
-      "Abscissa table vector (strict monotonically increasing values required)";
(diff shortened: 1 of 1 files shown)


HansOlsson (commented):
Looks fairly good.
My only concern is that it seems as if the interface is underspecified.
Consider `Vectors.interpolate({0, 2, 2}, {1,2,3}, 4)`
It says that should extrapolate through the last two points, but since they have the same x-value it is unclear what "extrapolation" means.
The previous variant seems to have asserted on that (possibly by mistake due to left-over code), now we return 4.
I think the assert might be preferable - and that it should be documented.

DagBruck:
Well, the documentation is contradictory and the check appears to be inconsistent. See example in https://github.com/modelica/ModelicaStandardLibrary/issues/4639#issuecomment-2865094598.

casella (commented):
@DagBruck would you mind adding some test cases of the specially handled cases to ModelicaTest as well?

Thanks!

PR #4643, opened 2025-05-15
ImprovedOpAmpLimited

AHaumer:
This PR replaces #4485 which will be closed but we still can look up the discussions.
The following explanation could be used for future release notes:
Modelica.Electrical.Analog.Ideal.IdealizedOpAmpLimited has been declared as deprecated since some of the results (Modelica.Electrical.Analog.Examples.OpAmps) could not be compared between different tools, because the usage of noEvent() and smooth() for saturation prevented exact events in switching applications. 
As a replacement Modelica.Electrical.Analog.Ideal.ImprovedOpAmpLimited has been implemented, thoroughly tested with different tools and used in the above mentioned examples. Saturation has been designed simpler, events are triggered by two Boolean variables that indicate saturation. Additionally a firstOrder models rise time of the OpAmp. Initialization is now without homopotopy acting on the result of the firstOrder (before saturation acts on the output).

### DIFF
--- a/Modelica/Electrical/Analog/Examples/OpAmps/Differentiator.mo
+++ b/Modelica/Electrical/Analog/Examples/OpAmps/Differentiator.mo
@@ -40,8 +40,7 @@ equation
   connect(der_.n2, vOut.n)
     annotation (Line(points={{10,-10},{40,-10}}, color={0,0,255}));
   annotation (Documentation(info="<html>
-<p>This is a (inverting) differentiating amplifier. Resistance R can be chosen, capacitance C is defined by the desired time constant resp. frequency.</p>
-<p>Note: <code>vOut</code> measure the negative output voltage.</p>
+<p>This is a (inverting) differentiating amplifier. Resistance R can be chosen, capacitance C is defined by the desired time constant resp. frequency.</p
 </html>"),
     experiment(
       StartTime=0,
--- a/Modelica/Electrical/Analog/Examples/OpAmps/Differentiator.mo
+++ b/Modelica/Electrical/Analog/Examples/OpAmps/Differentiator.mo
@@ -1,11 +1,11 @@
 within Modelica.Electrical.Analog.Examples.OpAmps;
 model Differentiator "Differentiating amplifier"
   extends Modelica.Icons.Example;
-  parameter SI.Voltage Vin=5 "Amplitude of input voltage";
-  parameter SI.Frequency f=10 "Frequency of input voltage";
+  parameter Modelica.Units.SI.Voltage Vin=5 "Amplitude of input voltage";
+  parameter Modelica.Units.SI.Frequency f=10 "Frequency of input voltage";
   Modelica.Electrical.Analog.Basic.Ground ground
     annotation (Placement(transformation(extent={{-20,-40},{0,-20}})));
-  Sources.TrapezoidVoltage vIn(
+  Modelica.Electrical.Analog.Sources.TrapezoidVoltage vIn(
     V=2*Vin,
     rising=0.2/f,
     width=0.3/f,
@@ -13,8 +13,7 @@ model Differentiator "Differentiating amplifier"
     period=1/f,
     nperiod=-1,
     offset=-Vin,
-    startTime=-(vIn.rising + vIn.width/2))
-    annotation (Placement(
+    startTime=-(vIn.rising + vIn.width/2)) annotation (Placement(
         transformation(
         extent={{-10,-10},{10,10}},
         rotation=270,
--- a/Modelica/Electrical/Analog/Examples/OpAmps/Differentiator.mo
+++ b/Modelica/Electrical/Analog/Examples/OpAmps/Differentiator.mo
@@ -1,8 +1,8 @@
 within Modelica.Electrical.Analog.Examples.OpAmps;
 model Differentiator "Differentiating amplifier"
   extends Modelica.Icons.Example;
-  parameter Modelica.Units.SI.Voltage Vin=5 "Amplitude of input voltage";
-  parameter Modelica.Units.SI.Frequency f=10 "Frequency of input voltage";
+  parameter SI.Voltage Vin=5 "Amplitude of input voltage";
+  parameter SI.Frequency f=10 "Frequency of input voltage";
   Modelica.Electrical.Analog.Basic.Ground ground
     annotation (Placement(transformation(extent={{-20,-40},{0,-20}})));
   Modelica.Electrical.Analog.Sources.TrapezoidVoltage vIn(
--- a/Modelica/Electrical/Analog/Examples/OpAmps/Differentiator.mo
+++ b/Modelica/Electrical/Analog/Examples/OpAmps/Differentiator.mo
@@ -40,7 +40,7 @@ equation
   connect(der_.n2, vOut.n)
     annotation (Line(points={{10,-10},{40,-10}}, color={0,0,255}));
   annotation (Documentation(info="<html>
-<p>This is a (inverting) differentiating amplifier. Resistance R can be chosen, capacitance C is defined by the desired time constant resp. frequency.</p
+<p>This is a (inverting) differentiating amplifier. Resistance R can be chosen, capacitance C is defined by the desired time constant resp. frequency.</p>
 </html>"),
     experiment(
       StartTime=0,
--- a/Modelica/Electrical/Analog/Examples/OpAmps/package.order
+++ b/Modelica/Electrical/Analog/Examples/OpAmps/package.order
@@ -12,7 +12,7 @@ VoltageFollower
 Comparator
 InvertingSchmittTrigger
 SchmittTrigger
+LCOscillator
 Multivibrator
 SignalGenerator
-LCOscillator
 OpAmpCircuits
--- a/Modelica/Electrical/Analog/Examples/OpAmps/package.order
+++ b/Modelica/Electrical/Analog/Examples/OpAmps/package.order
@@ -1,6 +1,5 @@
 NonInvertingAmplifier
 InvertingAmplifier
-DifferentialAmplifier
 Adder
 Subtracter
 Differentiator
@@ -8,6 +7,8 @@ Integrator
 LowPass
 HighPass
 ControlCircuit
+MeasureRiseTime
+DifferentialAmplifier
 VoltageFollower
 Comparator
 InvertingSchmittTrigger
--- a/Modelica/Electrical/Analog/Examples/package.order
+++ b/Modelica/Electrical/Analog/Examples/package.order
@@ -31,6 +31,6 @@ ParallelResonance
 DemonstrateLightning
 DemoPowerSupply
 DemoPowerSupplyWithBuffer
-Lines
 OpAmps
+Lines
 Utilities
--- a/Modelica/Electrical/Analog/Ideal/ImprovedOpAmpLimited.mo
+++ b/Modelica/Electrical/Analog/Ideal/ImprovedOpAmpLimited.mo
@@ -1,16 +1,19 @@
 within Modelica.Electrical.Analog.Ideal;
 model ImprovedOpAmpLimited "Improved operational amplifier with limitation"
   parameter Real V0=15000.0 "No-load amplification";
-  parameter SI.Time Tr=1e-5 "Rise time 10%-90% of output voltage";
   parameter Boolean useSupply=false
     "Use supply pins (otherwise constant supply)" annotation (Evaluate=true);
   parameter SI.Voltage Vps=+15 "Positive supply voltage"
     annotation (Dialog(enable=not useSupply));
   parameter SI.Voltage Vns=-15 "Negative supply voltage"
     annotation (Dialog(enable=not useSupply));
+  parameter Boolean useFirstOrder=false "use firstOrder rise of output voltage?"
+    annotation(Evaluate=true, Dialog(tab="Advanced"));
+  parameter SI.Time Tau=1e-5 "Time constant of firstOrder rise of output voltage"
+    annotation(Evaluate=true, Dialog(tab="Advanced", enable=useFirstOrder));
   parameter Modelica.Electrical.Analog.Types.InitOpAmp initOpAmp=Modelica.Electrical.Analog.Types.InitOpAmp.Linear
-    "Initialization of rise time fristOrder"
-    annotation (Evaluate=true, Dialog(group="Initialization"));
+    "Initialization of firstOrder rise of output voltage"
+    annotation (Evaluate=true, Dialog(tab="Advanced", enable=useFirstOrder));
   Boolean satPos=v_int>=vps "Indicates positive Saturation";
   Boolean satNeg=v_int<=vns "Indicates negative Saturation";
   SI.Voltage vps "Positive supply voltage";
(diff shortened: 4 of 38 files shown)


HansOlsson:
I'll look into it.
Just a quick style-question - is there a particular reason you switched from `SI.Frequency` (etc) to `Modelica.Units.SI.Frequency`?

AHaumer:
@HansOlsson this was by accident. 
I copied all the OpAmp stuff out of MSL, then developed and tested and copied back in. 
You're right, looks clumsy. Therefore replaced all `Modelica.Units.SI.` by `SI.`

maltelenz:
Without looking at the actual changes, I ran the full MSL tests, and got result differences in `Adder`, `ControlCircuit` and `Subtracter`.

Blue is the reference data, orange is the new result with this PR.

`Adder`:
image

In `ControlCircuit` it looks like an event became a softer transition at `0.5`? This is zoomed in:

image

`Subtracter`:
image

AHaumer:
After merging this PR, I'll prepare another PR containing a remake of Modelica.Electrical.Examples.ChuaCircuit using realistic parameters (mH and pF insteaf of H and F), showing the path from 2 periodic attractors to chaos and an additonal example demonstrating how the characteristic of Chua's diode could be implemented with 2 OpAmp-circuits called NIC (negative impedance converter) - that example relies on the OpAmp-model of this PR.

Modelica/Electrical/Analog/Examples/OpAmps/package.order
```diff
@@ -12,7 +12,7 @@ VoltageFollower
 Comparator
 InvertingSchmittTrigger
 SchmittTrigger
+LCOscillator
```
HansOlsson, line 15: Any reason for the changed order?

Modelica/Electrical/Analog/Examples/package.order
```diff
@@ -31,6 +31,6 @@ ParallelResonance
 DemonstrateLightning
 DemoPowerSupply
 DemoPowerSupplyWithBuffer
-Lines
 OpAmps
+Lines
```
HansOlsson, line 35: Also question about the order.

Modelica/Electrical/Analog/Ideal/ImprovedOpAmpLimited.mo
```diff
@@ -0,0 +1,110 @@
...
+    v_int=vns;
+  else
+    v_int=V0*v_in;
+  end if;
+equation
+  if not useSupply then
+    vps = Vps;
+    vns = Vns;
+  end if;
+  assert(vps > vns, "Supply voltages are not consistent: vps <= vns");
```
HansOlsson, line 57: There's no longer any need to repeat the condition; see https://github.com/modelica/ModelicaSpecification/pull/3641

Modelica/Electrical/Analog/Ideal/ImprovedOpAmpLimited.mo
```diff
@@ -0,0 +1,110 @@
...
+  //input currents
+  in_p.i = 0;
+  in_n.i = 0;
+  //firstOrder to model Rise time
+  der(v_int) = (V0*v_in - v_int)/(Tr/log(9));
```
HansOlsson, line 62: I don't fully get the log(9), and I think it should be explained/replaced by some final parameter.
I assume it is related to 10-90% rise time.

I also wonder if it would be possible to have an option to de-activate the firstOrder part.

Modelica/Electrical/Analog/Examples/OpAmps/Differentiator.mo
```diff
@@ -42,7 +41,6 @@ equation
     annotation (Line(points={{10,-10},{40,-10}}, color={0,0,255}));
   annotation (Documentation(info="<html>
 <p>This is a (inverting) differentiating amplifier. Resistance R can be chosen, capacitance C is defined by the desired time constant resp. frequency.</p>
-<p>Note: <code>vOut</code> measure the negative output voltage.</p>
```
HansOlsson, line 45: I noticed this comment has been erased for all of the examples, even though the component remains.
I assume it is intentional and because it is self-explanatory, right?

HansOlsson (commented):
In general good, but I would need some feedback on the filter-part.

Modelica/Electrical/Analog/Examples/OpAmps/Differentiator.mo
AHaumer: Yes I changed the polarity of measurement of the output voltage, therefore removed the line in the documentation. 
I saw that it is much easier to understand: This is an inverting circuit, and I see this fact looking at the output voltage without thinking about the polarity of the voltage sensor.

Modelica/Electrical/Analog/Ideal/ImprovedOpAmpLimited.mo
AHaumer: Yes we could use a choice between fristOrder and simple gain - I'll try that,
Maybe dependent on the size of the time constant (too large is bad, too small is bad).
And I'll name the time constant differently and remove the log(9), this is not exact.
The log(9) stems from:
0.1 = (1 - exp(-t10/Tau)
0.9 = (1 - exp(-t90/Tau)
Tr = t90 - t10 = Tau*ln(9)

Modelica/Electrical/Analog/Examples/OpAmps/package.order
AHaumer: Yes I tried to order the examples according to the problems they did rise during the last regression tests ;-)
I see no big harm in changing the order ...

Modelica/Electrical/Analog/Examples/package.order
AHaumer: Yes I tried to order the examples according to the problems they did rise during the last regression tests ;-)
I see no big harm in changing the order ...

Modelica/Electrical/Analog/Ideal/ImprovedOpAmpLimited.mo
AHaumer: Sorry I missed that discussion. Accepted.

Modelica/Electrical/Analog/Ideal/ImprovedOpAmpLimited.mo
AHaumer: @HansOlsson I think it is a good idea to have an option to deavtivate the firstOrder part. 
On the other hand this makes initialization much more complicated. 
With firstOrder it's easy: Just initialize the output of the firstOrder.
With simple gain I'd have to fiddle around with homotopy (to be able to fix the output at positive or negative limit) which isn't that easy when limits are not constant (parameters).

AHaumer:
> In general good, but I would need some feedback on the filter-part.

You mean the firstOrder? Give me some time over the weekend to design initialization.

AHaumer:
@HansOlsson I hope I addressed all your concerns, please re-review.
Now the firstOrder is optional (and therefore the initialization for it). It is needed only in 2 examples: MultiVibrator and SignalGenerator. The results seem to be the same in Dymola, OpenModelica and SystemModeler. 
@casella @maltelenz @henrikt-ma may I kindly ask for your review?

AHaumer:
Ok there a some cosmetic changes (my fault I know) in the examples.
The main work is a nearly complete redesign of the OpAmp model, some more explanations:

- The differential input voltage is either just amplified by V0 or (optional) additionally delayed with a firstOrder, this gives the fictional intermediate voltage
-  If the optional fristOrder is activated this intermediate voltage can be intialized as Linear, UpperLimit, LowerLimit.
- This intermediate voltage gets limited between positive and negative supply without triggering events to get output voltage.
- Events are triggered by 2 Boolean indicators of saturation, so saturation should be documented exactly.

Should I include a test model (meaybe in ModelicaTest?) measuring the rise time?
To meet the RiseTime stated in data sheet (e.g. LM741 0.3 µs), the time constant has to be set to V0/log(9)*RiseTime. 
RiseTime is measured in a voltageFollower circuit and defined by `Tr = t90 - t10` where 
`0.1 = 1 - exp(-t10/Tau)` and `0.9 = 1 - exp(-t90/Tau)`, using a step input. The differential equation for the voltageFollower is: 
`Tau*(der(vOut)) = (V0*(vStep - vOut)) - vOut` taking the firstOrder into account.
Many applications are nearly not affected by this firstOrder, so it is deactivated. 
For some applications the user would have to carefully compare time constant and transient behaviour of his input. 
For some applications (MultiVibrator and SignalGenerator) the firstOrder is essential to eases numerical solution.

AHaumer:
wait wait wait ... silly me: I forgot to check / adapt the comparisonSignals.

Then I want another change of the order of the examples. After that we get:

- 10 examples using the ready-to-use models from OpAmpCircuits
- 4 non-autonomous circuits and 1 autonomous (up to now uncritical) switching circuit
- 2 autonomous critical circuits

And I'll implement the correct measurement of rise time in ModelicaTest.

All done after the following commit (lastest this evening).

Modelica/Electrical/Analog/Ideal/ImprovedOpAmpLimited.mo
```diff
@@ -0,0 +1,128 @@
...
+<p>Idealized operational amplifier with saturation:</p>
+<ul>
+<li>Input currents are zero.</li>
+<li>No-load amplification is high (but not infinite).</li>
+<li>Output voltage is limited between positive and negative supply.</li>
+</ul>
+<p>Supply voltage is either defined by parameter <code>Vps</code> and <code>Vns</code> or by (optional) pins <code>s_p</code> and <code>s_n</code>.<br>
+In the first case the necessary power is drawn from an implicit internal supply, in the second case from the external supply.</p>
+<p>
+For most applications it is sufficient ot use default settings <code>Advanced.useFirstOrder=false</code>.
```
christiankral, line 104:
```suggestion
For most applications it is sufficient ot use default settings <code>useFirstOrder=false</code>. 
```

Using `Advanced.` somehow suggests that `Advanced` is a Modelica class.

Modelica/Electrical/Analog/Ideal/ImprovedOpAmpLimited.mo
```diff
@@ -0,0 +1,128 @@
...
+<li>Input currents are zero.</li>
+<li>No-load amplification is high (but not infinite).</li>
+<li>Output voltage is limited between positive and negative supply.</li>
+</ul>
+<p>Supply voltage is either defined by parameter <code>Vps</code> and <code>Vns</code> or by (optional) pins <code>s_p</code> and <code>s_n</code>.<br>
+In the first case the necessary power is drawn from an implicit internal supply, in the second case from the external supply.</p>
+<p>
+For most applications it is sufficient ot use default settings <code>Advanced.useFirstOrder=false</code>. 
+In this case the intermediate voltage <code>v_int</code> is simply <code>V0*v_in</code>.<br>
+In some applications it might be necessary to set <code>Advanced.useFirstOrder=true</code>
```
christiankral, line 106:
```suggestion
In some applications it might be necessary to set <code>useFirstOrder=true</code> 
```

Modelica/Electrical/Analog/Ideal/ImprovedOpAmpLimited.mo
```diff
@@ -0,0 +1,128 @@
...
+If this intermediate voltage <code>v_int</code> gets lower  than negative supply, Boolean <code>satNeg</code> inidicates negative saturation.<br>
+Intermediate voltage <code>v_int</code> is limited between positive supply and negative supply to achieve output voltage <code>v_out</code> without further events.
+</p>
+<h4>Initialization</h4>
+<p>
+<code>v_in.start</code> and <code>i_out.start</code> are just helpful starting values during initialization, 
+it is unlikely that it is necessary to change the start values and / or to declare them as fixed.
+</p>
+<p>
+If it is necessary to use <code>Advanced.useFirstOrder=true</code>, it is essential to initialize intermediate voltage <code>v_int</code> correctly.<br>
```
christiankral, line 121:
```suggestion
If it is necessary to use <code>useFirstOrder=true</code>, it is essential to initialize the intermediate 
```

Modelica/Electrical/Analog/Types/InitOpAmp.mo
```diff
@@ -0,0 +1,6 @@
+within Modelica.Electrical.Analog.Types;
+type InitOpAmp = enumeration(
+    Linear "Initialization of firstOrder linearly dependent on input",
+    UpperLimit "Initialization of firstOrder at positive supply",
+    LowerLimit "Initialization of firstOrder at negative supply")
+  "Enumeration defining initialization of the IdealizedOpAmpLimited" annotation (Evaluate=true);
```
christiankral, line 6:
```suggestion
  "Enumeration defining initialization of the ImprovedOpAmpLimited" annotation (Evaluate=true);
```

The type `InitOpAmp` is not used in `IdealizedOpAmpLimited`

christiankral (commented):
I wonder if `ImprovedOpAmpLimited` is a good choice for a component name. One day `IdealizedOpAmpLimited` will be removed and we might not remember exactly what was improved ...

christiankral:
> Sorry @maltelenz I forgot to mention:
> 
> * In some examples I flipped the output voltage sensor vertically. That explains the drastic differences.

We might want to add a comment in `Modelica.Electrical.Analog.UsersGuide.ReleaseNotes` in case someone is wondering why the results appear differently in MSL 4.1.0.

AHaumer:
@christiankral I accepted all your valuable suggestions with [9209342](https://github.com/modelica/ModelicaStandardLibrary/pull/4643/commits/92093421f165eb2446075d7ed9647ee8869586fb).
Regarding the naming, in Modelica.Electrical.Analog.Ideal we have;

- IdealOpAmp (infinite gain, no saturation)
- IdealOpAmp3Pin (infinite gain, no saturation)
- IdealOpAmpLimited (infinite gain, saturation, compulsory supply)
- deprecated IdealizedOpAmpLimited (finite gain, saturation, optional supply)
- new ImprovedOpAmpLimited (finite gain, saturation, optional supply, optional firstOrder rise of output, stability)

The new one is not only improved comparing wih the deprecated one, but also with the older three.

HansOlsson (approved):

christiankral (approved):
OK then, so we stick with the name `ImprovedOpAmpLimited`.

If I everybody else agrees with the name, it's fine with me, too...

AHaumer:
Text for release notes:
ImprovedOpAmpLimited replaces IdealizedOpAmpLimited (which is deprecated), with finite gain, saturation, optional supply and optional firstOrder rise of output which improves stability.
The examples in OpAmps are adapted to use the new ImprovedOpAmpLimited and harmonized polarity of output voltage sensor.
@GallLeo @MatthiasBSchaefer please provide new reference results for all examples in Modelica.Electrical.Analog.Examples.OpAmps after this PR has been merged.

casella:
Unfortunately the discussion on the naming was too quick for me to react. I don't really think that "Improved" is a good attribute for any Modelica Standard Library model. Same as "New". It is new or improved now, and then it will be soon obsolete, as all things human.

The name should describe **what the model represents**, not **how its development history went**.

Why don't we just call it OpAmpLimited? It's not idealized, because it has a finite gain. So, it's just an op amp. @AHaumer do you agree?

PR #4644, opened 2025-05-19
Redesign ChuaCircuit and add ChuaDiode

AHaumer:
I wanted to redesign the parameterization of Chua's Circuit to realistic values (mH instead of H, nF instead of F) and explain how the nonlinear resistor called Chua's diode can be implemented using two OpAmp circuits called negative impedance converter.
With Chua's circuit you may see that changing initialization switches between two attractors.
Furthermore, changing a resistance you may see the path to chaos. 
Before we start a discussion about perodicity and chaos: Please have a look at the references. 
Default parameterization delivers periodic results. 
Results obtained with Dymola, Open Modelica and Wolfram System Modeler are (nearly) the same:
ChuaCircuitv1
ChuaCircuitv2

### DIFF
--- a/Modelica/Electrical/Analog/Examples/ChuaCircuit.mo
+++ b/Modelica/Electrical/Analog/Examples/ChuaCircuit.mo
@@ -1,58 +1,94 @@
 within Modelica.Electrical.Analog.Examples;
-model ChuaCircuit "Chua's circuit, ns, V, A"
+model ChuaCircuit "Chua's circuit"
   extends Modelica.Icons.Example;
-
-  Modelica.Electrical.Analog.Basic.Inductor L(L=18, i(start=0, fixed=true)) annotation (Placement(transformation(
-        origin={-75,38},
-        extent={{-25,-25},{25,25}},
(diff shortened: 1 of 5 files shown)


christiankral (commented):
Is there a good reason, why the components in the Diagram View have double the size than usual?

image

AHaumer:
This ist the original layout, just adapted the parameters.

christiankral:
OK, I then very much recommend to switch to the standard size for the sake of a unified layout and appearance of the MSL.

AHaumer:
converted to draft: Has to be updated after #4653 has been merged.

christiankral:
I think we shall avoid redundant information in the diagram layer which is also stated in the documentation. 

image

AHaumer:
Regarding information on the diagram layer:
I believe that many users look at the diagram layer, and after that at the text layer. 
They don't read the documentation layer unless is pops up due to "preferredView".
But ok.

casella:
Maybe we should write "RTFD" in block capitals in the diagram 😉

PR #4658, opened 2025-06-05
Remove connection that was already defined in parent class. It genera…

fzenith:
There is a redundant connection in Enable1 that is already defined (identical down to annotation and indenting) in parent class Enable.
Models would still compile as the specification allows it, but OMedit reports irritating "scripting errors".


AHaumer:
`Modelica.Electrical.PowerConverters.Interfaces.Enable.Enable1` extends from `Modelica.Electrical.PowerConverters.Interfaces.Enable.Enable`
Both contain the equation `connect(enableLogic.enable, enable);`

AHaumer (approved):
IMHO correct

PR #4659, opened 2025-06-10
Setting master version to 4.2.0 dev

AHaumer:
Setting version to 4.2.0 (also ModelicaTest, Complex, etc. - hope I didn't forget one) 
and provide ReleaseNotes.mo for 4.2.0, including the already merged OpAmp (#4643 and #4653)


maltelenz (commented):
Seems fine to me, but some unnecessary whitespace changes, see comments.

AHaumer:
Thanks @maltelenz I wasn't aware of these tiny whitespace changes.
Without discussing tool issues in public: I didn't touch these parts of MSL and I don't know how to avoid these changes.
I hope I have addressed them all. Sorry I did this myself because I wasn't logged in at github and therefore I didn't see that I could accept your suggestions easily ...

maltelenz:
> Without discussing tool issues in public: I didn't touch these parts of MSL and I don't know how to avoid these changes.

Perfectly preserving whitespace when editing is a very hard problem for a tool, so whichever tool you used I don't blame it :)

I try to discard small changes like these before committing, but they can slip through the cracks easily.

Modelica/UsersGuide/ReleaseNotes.mo
```diff
@@ -126,6 +126,84 @@ more of the following changes.
 </html>"));
 end VersionManagement;
 
+class Version_4_2_0 "Version 4.2.0 (mmm dd, yyyy)"
+  extends Modelica.Icons.ReleaseNotes;
+
+  annotation (Documentation(info="<html>
+<p>Version 4.2.0 is backward compatible to version 4.1.0. This means, there is no need to convert user models, because no renaming or moving has been done on the library. Please note: simulation results might be affected by necessary changes and bug fixes that were applied since the 4.1.0 release, see below for more details.</p>
```
maltelenz:
```suggestion
<p>Version 4.2.0 is backward compatible to version 4.1.0. This means, there is no need to convert user models, because no renaming or moving has been done on the library.</p>
```

maltelenz (dismissed):
The rest of the placeholder things are probably OK (so it is remembered to fill it out when the time comes), but this sentence seems specific to 4.1.0 and I don't think we need it.

Otherwise fine.

HansOlsson (commented):
Clearly we are aiming for 4.2.0 at some point in time, but calling it 4.2.0 already will be a bit confusing. Wouldn't it be better to call it "4.2.0 Development" or similarly?

That syntax is specified in Modelica; and intended exactly for such scenarios.

beutlich (changes requested):
I would not change the version now. See also #4661 as my alternate PR.

maltelenz:
> To be honest, I would prefer "4.2.0 dev" - you see even later it had been taken from master.

If that's the consensus, I'm OK with that, even if I prefer to keep to cleaner version numbers.

PR #4660, opened 2025-06-10
Avoid wildcard effect in default unity gains

henrikt-ma:
Too often, I have been bitten by the hidden gains in blocks such as `Integrator` and `Add`, where unit errors are absorbed by inferring unexpected units for these gains with value 1.  This PR comes in three parts corresponding to the initial three commits:
- Introduce a constant to use instead of a literal `1` when one wants a 1 with unit `"1"`.
- Use the new constant for the default gains where one doesn't want the wildcard effect.
- Fix detected unit errors where the affected blocks are used.

I must warn that I will soon be off for vacation, and may not be able to respond to feedback until July 7.

### DIFF
--- a/Modelica/Blocks/Continuous.mo
+++ b/Modelica/Blocks/Continuous.mo
@@ -622,18 +622,17 @@ This is discussed in the description of package
     parameter Real y_start=0 "Initial value of output"
       annotation(Dialog(enable=initType == Init.InitialOutput, group=
             "Initialization"));
-    constant SI.Time unitTime=1 annotation(HideResult=true);
 
     Blocks.Math.Gain P(k = Modelica.Constants.one) "Proportional part of PID controller"
       annotation (Placement(transformation(extent={{-60,60},{-20,100}})));
-    Blocks.Continuous.Integrator I(k=unitTime/Ti, y_start=xi_start,
+    Blocks.Continuous.Integrator I(k=1/Ti, y_start=xi_start,
       initType=if initType==Init.SteadyState then
                   Init.SteadyState else
                if initType==Init.InitialState then
                   Init.InitialState else Init.NoInit)
       "Integral part of PID controller"
       annotation (Placement(transformation(extent={{-60,-20},{-20,20}})));
-    Blocks.Continuous.Derivative D(k=Td/unitTime, T=max([Td/Nd, 100*Modelica.
+    Blocks.Continuous.Derivative D(k=Td, T=max([Td/Nd, 100*Modelica.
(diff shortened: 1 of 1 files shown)


AHaumer:
I agree with you on the fact of getting weird units ...
Fine with me to have a constant "one" (1 with unit "1"), but rather dangerous or misleading for integrator and some other. 
Let's look at the integrator: `der(y) = k*u;`
So if u has unit "1" and you expect y to have unit "1" k has to have unit "1/s". 
That's the drawback of having signals without units (or having unit "1"). 
In other cases unit "1" for k is ok:
If u has unit "m/s" and you expect y to have unit "m" k has to have unit "1".
I need some more time to check the usage in control applications where normally we use k = 1/(integral time constant).

henrikt-ma:
> So if u has unit "1" and you expect y to have unit "1" k has to have unit "1/s".
> That's the drawback of having signals without units (or having unit "1").

To me this is not a drawback, it is precisely how it should be!

AHaumer:
> > So if u has unit "1" and you expect y to have unit "1" k has to have unit "1/s".
> > That's the drawback of having signals without units (or having unit "1").
> 
> To me this is not a drawback, it is precisely how it should be!

That's ok (Sometimes I prefer having signals with units).
But in that case integrator.k should have unit "1/s" instead of "1".

henrikt-ma:
I should also mention the relation to https://github.com/modelica/ModelicaSpecification/pull/3688.  With that syntax, the definition of `Modelica.Constants.one` would become obsolete, as one could just write `1'1'` instead of referring to the constant.  There would also be additional convenient alternatives for how to parameterize the `integrator` in my example above:
```
  Modelica.Blocks.Continuous.Integrator integrator(k = 10'1/s');
```
or
```
  Modelica.Blocks.Continuous.Integrator integrator(k = 1 / 0.1's');
```

(Unfortunately, I am not allowed to explain how the syntax can be enabled in Wolfram System Modeler, but let me know if you would like a demonstration.)

AHaumer:
> ```
>   Modelica.Blocks.Continuous.Integrator integrator(k = 10'1/s');
> ```
I like that idea! Additionally, it should be possible to adapt the unit not only textually but also in the graphical menu.

henrikt-ma:
> I like that idea! Additionally, it should be possible to adapt the unit not only textually but also in the graphical menu.

So far so good, but now we must make sure that this side-topic is not growing too big in the discussion of this PR.  To be continued in https://github.com/modelica/ModelicaSpecification/pull/3688.

PR #4661, opened 2025-06-10
Update latest version to v4.1.0

beutlich:


AHaumer:
ok so far ... but shouldn't 4.1.0 be tagged (and not touched any more) with these dates, and then changing the version number in master to 4.2.0 (#4659)?

AHaumer (changes requested):
In MAP-Lib we decided the next relase will be 4.2.0 in about 1 year.
A patch 4.1.1 will be released only if severe bugs are reported.
Naming the development version same as the latest release is misleading.
The dates are ok of course.
"4.2.0 dev" would be easy to understand (this is my PR #4659)
And: For heavens sake include an empty table for the next release notes (also in #4659)!

AHaumer:
If @casella overrides me, the master version shall be named 4.1.0.
But: I'll never accept this PR without an empty release notes table!!!
If the name "4.2.0 dev" is acceptec within MAP-Lib, we could use #4659

PR #4672, opened 2025-07-10
Update README.md with the correct current release

MartinOtter:
Refer to the correct current release (4.1.0 not 4.0.0)


beutlich:
Already handled by #4661.

PR #4674, opened 2025-07-16
Add starting point for figures in Blocks and ComplexBlocks.

maltelenz:
These originate from figures created for Wolfram System Modeler, and are likely to need cleanup and improvement.

Our one existing figure in MSL is probably a good style to follow:

https://github.com/modelica/ModelicaStandardLibrary/blob/3f72865f0764cad0f2fb2d47d1e3fc0bf96ec4e6/Modelica/Blocks/package.mo#L159-L180

Creating as Draft to indicate that library officer(s) probably want/should improve them before merging.

PR #4675, opened 2025-07-16
Add starting point for figures in Mechanics.Translational.

maltelenz:


christiankral:
See also #4674: Many models work fine. Yet there are some models which fail in Dymola2025x under Linux. The failing models are:

- Modelica.Mechanics.Translational.Examples.Brake
- Modelica.Mechanics.Translational.Examples.Friction
- Modelica.Mechanics.Translational.Examples.HeatLosses

Example error message caused by Dymola2025x:

```
Compilation of [Modelica.Mechanics.Translational.Examples.Brake](modelica://Modelica.Mechanics.Translational.Examples.Brake) failed.

Note: There were translation errors that might explain the problem.

Compiler message:
Using CC=cc -Wl,--no-as-needed
dsmodel.c: In function ‘Modelica_Blocks_Types_ExternalCombiTable1D_M’:
dsmodel.c:181:9: error: ‘ModelicaStandardTables_CombiTable1D_init3’ undeclared (first use in this function); did you mean ‘ModelicaStandardTables_CombiTable1D_init2’?
181 | ModelicaStandardTables_CombiTable1D_init3)(fileName0_0, tableName0_0,
| ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
| ModelicaStandardTables_CombiTable1D_init2
dsmodel.c:181:9: note: each undeclared identifier is reported only once for each function it appears in
```

This somehow seems to be a table issue.

tobolar:
> Yet there are some models which fail in Dymola2025x under Linux.

Shall this not be an another issue?

Modelica/Mechanics/Translational/Examples/Accelerate.mo
```diff
@@ -19,6 +19,23 @@ Demonstrate usage of component
...
+</html>",
+      figures = {
+        Figure(
+          title = "Mass variables",
+          identifier = "79d80",
+          preferred = true,
+          plots = {
+            Plot(
+              curves = {
+                Curve(y = mass.a, legend = "Accelleration of mass"),
```
HansOlsson, line 31:
```suggestion
                Curve(y = mass.a, legend = "Acceleration of mass"),
```
Spelling

christiankral:
> Some ideas for initial content of a figure style guide for the MSL:
> 
> * Don't mix variables with incompatible units in the same plot. (The use of a second y-axis can be a nice way to relax this requirement, but this feature is not in the standard.)
> * Each figure shall have a caption.
> * Some style to be used when describing the different plots in the caption of a multi-plot figure.

I think we shall not manually add units to the curve legend, as the tools are free to chose for example MJ (Megajoule) instead of 1E6 in scaling. The unit shall be added by the tool automatically. So not mixing variables makes a lot more sense to me.

christiankral:
One case where units are yet not derived correctly in Dymola2025x is described in https://github.com/modelica/ModelicaStandardLibrary/pull/4676#issuecomment-3317459242 referring to example Modelica.Electrical.PowerConverters.Examples.DCAC.PolyphaseTwoLevel.PolyphaseTwoLevel_RL:

image

I tested this case in Dymola2025x. For some reason the unit of the rms current and voltage are not determine correctly in `Modelica.Blocks.Math.Harmonic`. Even though the units match in the plots, the units are not displayed.

AHaumer:
> I assume that either the table-functions library wasn't recompiled for MSL 4.1.0 on Linux, or the new library wasn't used (the Advanced.Translation.PlaceDymolaSourceFirst setting). Dymola 2025x Refresh 1 should have reduced the second issue, and it will work without any problems on Dymola 2026x

Same issue with Dymola 2025x Refresh 1 under Windows 11 24H2:
Compiling and linking the model (Visual C++).
dsmodel.c
LINK : fatal error LNK1181: Eingabedatei "ModelicaIO.lib" kann nicht geffnet werden.
Error generating Dymosim.

PR #4676, opened 2025-07-16
Add starting point for figures in Electrical Machines, PowerConvertes, QuasiStatic

maltelenz:


AHaumer:
I agree with @christiankral that the issue (showing units) has to be resolved. 
A workaround could be: Split the plots into subplots.

PR #4681, opened 2025-07-16
Add starting point for figures in Mechanics.Rotational.

maltelenz:


tobolar:
@maltelenz First of all, thanks for reaching out and opening this and related "mechanical" issues.

Generally, I would prefer to reduce the usage of titles.
IMO the people use to understand texts like "inertial.w" or "torque.tau", so my preference is to reduce the usage of `legend` in `Curve` where possible. Where curves of the same quantity are ploted in one diagram, something like  `Plot.title="Angular velocities"` could be sufficient.

I will review your PRs regarding "mechnical" examples having abovementioned in mind. Starting probably end of September due to vacations.

henrikt-ma:
> Generally, I would prefer to reduce the usage of titles. IMO the people use to understand texts like "inertial.w" or "torque.tau", so my preference is to reduce the usage of `legend` in `Curve` where possible. Where curves of the same quantity are ploted in one diagram, something like `Plot.title="Angular velocities"` could be sufficient.

I think most of these figures were created before we even had support for captions in System Modeler, which much also have contributed to titles trying to convey too much information.  Now that a figure can have captions, I am also strongly in favor of having concise titles (but not to leave empty and for the tool to decide), and then make use of captions whenever this is needed to make the figure speak for itself.

PR #4682, opened 2025-07-16
Add starting point for figures in StateGraph.

maltelenz:

### DIFF
--- a/Modelica/StateGraph.mo
+++ b/Modelica/StateGraph.mo
@@ -809,7 +809,31 @@ package Examples
       annotation (Line(points={{30.5,10},{46,10}}));
     connect(transition2.outPort, initialStep.inPort[1]) annotation (Line(points=
              {{51.5,10},{70,10},{70,32},{-62,32},{-62,10},{-49,10}}));
-    annotation (experiment(StopTime=5.5));
+    annotation (
+      Documentation(
+        figures = {
+          Figure(
+            title = "Boolean signal in system",
+            identifier = "ef029",
+            preferred = true,
+            plots = {
+              Plot(
+                curves = {
+                  Curve(y = initialStep.active, legend = "Output from initialStep"),
+                  Curve(y = transition1.t, legend = "transition1 wait time")
+                }
+              ),
+              Plot(
+                curves = {
+                  Curve(y = step.active, legend = "Output from step"),
+                  Curve(y = transition2.t, legend = "transition2 wait time")
+                }
+              )
+            }
+          )
+        }
+      ),
+      experiment(StopTime=5.5));
   end FirstExample;
 
   model FirstExample_Variant2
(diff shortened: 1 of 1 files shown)


henrikt-ma:
> Hmm... Checking for Modelica.StateGraph, it's @HansOlsson and @MartinOtter !
> I'll try formulate some caption at least for the FirstExample(s) as a first step.

Great!  In my opinion, having a caption should be a base requirement that all figures should live up to before a PR like this is accepted.  How about writing down this sort of rule in https://github.com/modelica/ModelicaStandardLibrary/blob/master/Modelica/UsersGuide/Conventions.mo?  With guidelines to follow, it should be easier to recognize when these PRs with the first round of figures are ready to be merged, and we would get more consistent looks across sub-libraries.

Modelica/StateGraph.mo
```diff
@@ -1076,7 +1150,46 @@ has a higher priority to fire as alternative.split[2]).
...
+              Plot(
+                curves = {
+                  Curve(y = step6.active, legend = "Output from step6"),
+                  Curve(y = transition7.condition, legend = "transition7 condition for firing status")
+                }
+              )
+            }
+          )
+        }
+      ), experiment(StopTime=15));
```
HansOlsson:
```suggestion
     figures = {
          Figure(
          title = "Cyclic activation of a parallel and alternative steps.",
          caption="This shows that steps can be in parallel (synchronized at the end).\nAdditionally it contains alternatives, that are selected based on priority.\nThe compositeStep shows how to build structure"
            identifier = "22111",
            preferred = true,
            plots = {
              Plot(
                curves = {
                  Curve(y = step0.active, legend = "Active step0"),
                Curve(y = compositeStep.active, legend = "Active compositeStep - parallel to step1"),
                  Curve(y = step1.active, legend = "Active step1"),
                  Curve(y = step6.active, legend = "Active step6")}),
              Plot(
                curves={
                  Curve(y = compositeStep.active, legend = "Active compositeStep - where step3, step4, step4a are alternatives"),
                  Curve(y = compositeStep.initStep.active, legend = "Active compositeStep.initStep"),
                  Curve(y = compositeStep.step3.active, legend = "Active compositeStep.step3"),
                  Curve(y = compositeStep.step4.active, legend = "Active compositeStep.step4"),
                  Curve(y = compositeStep.step4a.active, legend = "Active compositeStep.step4a"),
                  Curve(y = compositeStep.exitStep.active, legend = "Active compositeStep.exitStep")
                }
              )
            }
          )
        }
      ),
  experiment(StopTime=15));
```
As I understand it, this illustrates the main point of the example as:
- step1 and compositeStep are in parallel; so both are active at the same time (there's a synchronization when ending)
- inside compositeStep there are alternatives (step3, step4, step4a) - where only one is active.

The actual timings etc aren't the focus, so don't plot them.
(Should have a similar figure for ExecutionPaths.)

I believe the example would be clearer if different alternatives were selected for the different cycles.

Issue #4688, opened 2025-08-21
Problem with the test of Modelica/Electrical/Machines/Examples/InductionMachines/IMC_DCBraking


uschna:
There is a problem with the test of `Modelica/Electrical/Machines/Examples/InductionMachines/IMC_DCBraking`. If we reduce the tolerances the results will change. According to `Modelica/Electrical/Machines/Examples/InductionMachines/IMC_DCBraking/creation.txt` (in branches v4.0.0 and v4.1.0) the used tolerance is `1e-7`.
With this tolerance the last value of "`imc.idq_sr[1]`" in `IMC_DCBraking.csv` is `-15.240146546765066`.
But with smaller tolerances (for example `1e-10`) this value is approximately `-14.1`.
Could you please compute and save the reference results with `Tolerance=1e-10`?
Thank you.

henrikt-ma:
I made a quick experiment with System Modeler.  For a variable with dynamic range of approximately 280, I aimed for an absolute error tolerance of 1e-3 * 280 = 0.28.  With tighter tolerance and when using DASSL for integration, the final result appears to approach −14.125 from below.  When using CVODES for integration I get somewhat different results, but both methods agree on the final value -14.125 at tolerance 1e-10.  At 1e-8 I was above −14.125 - 0.28 with DASSL (and as usual CVODES was consistently performing better), so based on this I would have set `Tolerance = 1e-8` in the `experiment`-annotation, and used 1e-10 for reference result creation.

AHaumer:
Let me explain what happens:
The current in the windings and therefore the Clark-transformed space phasor w.r.t. the stator-fixed coordinate system is impressed. 
The transformation to the rotor-fixed coordinate system depends on the mechanical angle imc.phiMechanical. 
The dependency of imc.phiMechanical on Tolerance is rather weak. 
But if the angle is either near n*pi or pi/2+n*pi either sine or cosine change significantly even with small changes of the angle. 
Therefore currents w.r.t. the rotor-fixed coordinate system shouldn't be used for comparison, especially only comparing final values.
I think we should't change the Tolerance (with could be dependent on the algorithm, the tool, ...)
@uschna do you agree?

Thanks to @johhell another problem was detected - no idea why this didn't some uo during the last tests:
This example won't simulate in OpenModelica. 
I have prepared an adapted version - see enclosed. @uschna could you please test?
Then I would prepare a PR. 
[...]

PR #4691, opened 2025-09-12
Examples/InductionMachines/IMC_DCBraking.mo 

AHaumer:
enhances the mentioned example to avoid problems
changes the comparisonSignals (old ones are bad - see discussion in #4688)
@uschna pls. test
@GallLeo @MatthiasBSchaefer pls. produce new reference result - thx!
@casella should we backport to 4.1.1? I suppose "no".


uschna:
Dear @AHaumer,
I tested it again with this version and your results. It is also valid. Thank you.

christiankral:
Instead of comparing stator and rotor current components, we could for example compare the magnitudes of the current phasor components. In case of the stator current this is equivalent to the quasi rms current of the phase currents sensed by `Polyphase.Sensors.CurrentQuasiRMSSensor`:

image

Modelica/Electrical/Machines/Examples/InductionMachines/IMC_DCBraking.mo
```diff
@@ -49,50 +56,32 @@ model IMC_DCBraking "Induction machine with DC current braking"
     annotation (Placement(transformation(extent={{-20,70},{-40,90}})));
   Modelica.Electrical.Polyphase.Basic.PlugToPin_p plugToPin2(m=m, k=2)
     annotation (Placement(transformation(extent={{-20,30},{-40,50}})));
-  Modelica.Electrical.Polyphase.Basic.PlugToPin_p plugToPin3(m=m, k=3) if  settings.connect3
+  Modelica.Electrical.Polyphase.Basic.PlugToPin_p plugToPin3(m=m, k=3)
     annotation (Placement(transformation(extent={{-20,-10},{-40,10}})));
-  Modelica.Electrical.Analog.Sources.ConstantCurrent
-    constantCurrent(I=settings.Idc)
+  Analog.Sources.ConstantVoltage
```
HansOlsson:
```suggestion
  Modelica.Electrical.Analog.Sources.ConstantVoltage
```
I know that both variants work, but I find it confusing that just a few declarations use a shorter name.

PR #4705, opened 2025-11-06
Fix descriptions in ComplexElectric types

GallLeo:
Description for `re` and `im` were swapped?

Modelica.Units.SI.ComplexElectricPotential


HansOlsson:
I agree with the change, but I haven't used complex numbers for electrical things that much so I might miss some convention.

As far as I understand it was introduced by https://github.com/modelica/ModelicaStandardLibrary/commit/2a86acdc9d427ed006a452d7a1f9ebfb9612d386
and then updated several times without anyone noticing.

AHaumer (approved):
correct

PR #4708, opened 2025-11-13
Remove Dymola_editText annotation

maltelenz:
Vendor specific annotations should be avoided in the standard library.


HansOlsson:
BTW: 
I don't see that the annotation brings that much benefit anyway in Dymola (will internally investigate), and it has been there at least since Modelica_Synchronous was added to MSL.

PR #4726, opened 2026-01-28
Clarify file extension and formats

### LINKED ISSUE #4704: The CombiTables documentation does not report that the .csv file type doesn't requre the header
The documentation of the CombiTables in the MSL reports that:

```
The table matrix can be defined in the following ways:
...
2. the data table can be Read from a file "fileName" where the matrix is stored
    as "tableName". Both text and MATLAB MAT-file format is possible. (The text format is described below)._
...
```
and that:

```
If tables are read from a text file, the file needs to have the following structure ("-----" is not part of the file content):

-----------------------------------------------------
#1
double tab1(5,2)   # comment line
  0   0
  1   1
  2   4
  3   9
  4  16
double tab2(5,2)   # another comment line
  0   0
  2   2
  4   8
  6  18
  8  32
----------------------------------------------------
```

This  works fine if the file in which the data matrix is stored has the extension different from **".csv"**. If the extension of the file is **".csv"** the above listed format generates an error in OpenModelica (not in Dymola).

If the first two header lines are removed and the file contains one only matrix then the data matrix file is properly read by both OpenModelica and Dymola.

If this different behavior of the **".csv"** files has been intentionally implemented, then it should be described in the MSL documentation.

beutlich:
Closes #4704 

See also https://github.com/tbeu/Modelica2021_ExternData/blob/02117c93ee96e6d5a5b1115ded46a83fbaa91c5e/modelica2021_ExternData.tex#L132 on the name "Dymola MOS".

### DIFF
--- a/Modelica/Blocks/Sources.mo
+++ b/Modelica/Blocks/Sources.mo
@@ -1839,8 +1839,11 @@ tableName is \"NoName\" or has only blanks,
 fileName  is \"NoName\" or has only blanks.
 </pre></blockquote></li>
 <li><strong>Read</strong> from a <strong>file</strong> \"fileName\" where the matrix is stored as
-    \"tableName\". CSV, text and MATLAB MAT-file format is possible.
-    (Both the limitations on the CSV format and the text format are described below).
+    \"tableName\". CSV, DLR matrix text and MATLAB MAT-file formats are possible.
+    (Both the limitations on the CSV format and the DLR matrix text format are described below).
+    Files with extension .csv are read as CSV format.
+    Files with extension .mat are read as MAT-file format.
+    Files with any other extension are read as DLR matrix text format.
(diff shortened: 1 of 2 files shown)


maltelenz (changes requested):
I object to calling the text file format a Dymola MOS file. Even if Dymola was the first to use it (I assume it was), it has clearly become something tool independent in the decade(s) since then.

beutlich:
@maltelenz In a constructive manner, please do not state how to not call 
it but how to call it instead. You can directly edit this PR.

Am 2026-01-29 09:06, schrieb Malte Lenz:
> @maltelenz requested changes on this pull request.
> 
> I object to calling the text file format a Dymola MOS file. Even if 
> Dymola was the first to use it (I assume it was), it has clearly become 
> something tool independent in the decade(s) since then.

PR #4739, opened 2026-02-16
FluxTubes.Basic.ElectroMagneticConverter: L_stat

AHaumer:
Both the ElectrocMagenticConverter and the ElectroMagneticConverterWithLeakageInductance lack a warning that L_stat is only valid if the coil is the only source in the magnetic circuit, Otherwise the magnetc flux is at least partly driven by other sources, too, and the calculation of L_stat is misleading.
Additonally, in the equation for L_stat the usage of noEvent has been corrected.


dietmarw (approved):

christiankral (approved):
Important improvement

PR #4741, opened 2026-02-16
FluxTubes: Additonal approximation

AHaumer:
When investigating #4740, trying to implement a new material, I found it really hard to calculate correct parameters for the used approximation according to T.Roschke. I have implemented a backwards compatible solution to utilize additional approximation formulae. For now, I've implemented a (slightly modified) formula suggested by W.Macfadyen, and created two parameter records (M330-50A and M400-50A), as well as two records with raw data (magnetization characteristic from manufacturer) to document the source of the approximation parameters. If I manage to get more manufacturers' datasheet, I would implement more parameter records.
Additionally, I added 3 examples demonstrating the usage of the new approximation function., 
and some new pages in the User's Guide explaining the two approximation functions 
and the workflow to fit parameters to given characteristics.
Feedback is highly welcome.


AHaumer:
Closing: will be replaced by a new PR.

PR #4753, opened 2026-02-27
Modelica.Math.Special.{der(erf), der(erfc)}

AHaumer:
I noticed that the derivatives of erf and erfc are missing but can be easily defined. 
Important if using erf in model equations or for index reduction, since erf is non-trivial.
Although it's a bug if one needs the derivative this is an enhancement.


casella (commented):
@AHaumer, in the testing, we need to prove that the derivatives provided by the derivative annotation are very precise. Reference signals are not good for that, because CSV-compare if rather tolerant.

I would suggest to remove the comparison signals entirely, and instead
- select a much tighter tolerance, e.g 1e-9, for the simulation of TestErf
- add two asserts in the equation section of the TestErf such as

```modelica
assert(abs(y1i - y1) < 1e-7, "The derivative function for erf is not correct");
assert(abs(y2i - y2) < 1e-7, "The derivative function for erfc is not correct");
```
If the test model runs successfully, then by construction it means the derivative is computed with high accuracy.

HansOlsson:
And also add `annotation(smoothOrder=100)` or similarly for these derivative-functions.
That way all tools can differentiate them as much as needed.

HansOlsson (approved):

PR #4754, opened 2026-02-28
Simplify cases for `erf`

phannebohm:
The local variable `z` is by construction non-negative and the linear case gives `y = 0` for `z = 0`.

~This makes the symbolic derivative of `erf` continuous around zero.~ (No longer relevant as soon as #4753 is merged.)

This simplifies the algorithm and clarifies the constant values.

### DIFF
--- a/Modelica/Math/Special.mo
+++ b/Modelica/Math/Special.mo
@@ -30,7 +30,7 @@ package Special "Library of special mathematical functions"
       inv := true;
     end if;
 
-    if z < sqrt(3.0*Modelica.Constants.eps/2.0) then
+    if z < sqrt(1.5*Modelica.Constants.eps) then
       // Maclaurin series to first order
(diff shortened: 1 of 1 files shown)


Modelica/Math/Special.mo
```diff
@@ -30,10 +30,10 @@ package Special "Library of special mathematical functions"
       inv := true;
     end if;
 
-    if z < sqrt(3*2^(-53)) then
+    if z < sqrt(3.0*Modelica.Constants.eps) then
```
HansOlsson: Machine epsilon in Modelica is the "Interval machine epsilon", so `2^(-52)` -i.e., the distance between 1 and the next floating point number, not `2^(-53)` (the rounding machine epsilon).

But here you switched from  `2^(-53)` to eps.

HansOlsson (approved):
Looks good

PR #4767, opened 2026-03-19
Initialize CI for PullRequest Testing

MatthiasBSchaefer:
This PR adds the yml file to .github/workflows in order to start the CI for testing PullRequests with the ltx-resim tool.
infrastructure to run the CI-tests is currently hosted and maintained by LTX

see PRs #4725 #4763 #4764 #4765 , which are tests of the CI on a separate branch


.github/workflows/PRTesting.yml
```diff
@@ -0,0 +1,35 @@
+name: PR_testing
+on:
+  # Triggers the workflow on push or pull request events but only for the "main" branch
+  pull_request:
+  
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
```
beutlich:
```suggestion
name: CI ReSim

on:
  pull_request:
  workflow_dispatch:

```

.github/workflows/PRTesting.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  prepare:
+    runs-on: self-hosted
```
beutlich:
```suggestion
    runs-on: [ self-hosted, linux, regression-testing ]
```

.github/workflows/PRTesting.yml
```diff
@@ -0,0 +1,35 @@
...
+  
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  prepare:
+    runs-on: self-hosted   
+    steps: 
+      - name: checkout_pr_repo
```
beutlich:
```suggestion
      - name: Checkout code
```

.github/workflows/PRTesting.yml
```diff
@@ -0,0 +1,35 @@
...
+        run: /home/resimuser/regression_testing/docker/work/run_scripts/prepare_pr.sh ${{ github.event.number }}
+  
+  testrun_modelica:
+    needs: prepare
+    runs-on: self-hosted
```
beutlich:
```suggestion
    runs-on: [ self-hosted, linux, regression-testing ]
```

.github/workflows/PRTesting.yml
```diff
@@ -0,0 +1,35 @@
...
+    environment: 
+        name: link_modelica
```
beutlich:
```suggestion
        name: Build libraries
```

Not sure, what is meant by "link".

.github/workflows/PRTesting.yml
```diff
@@ -0,0 +1,35 @@
...
+        url: https://serv.ltx.de/prs/${{ github.event.number }}/Modelica/report/PR_comparison_report.html
+    steps:         
+      - name: run modelica      
+        run: /home/resimuser/regression_testing/docker/work/run_scripts/run_pr.sh ${{ github.event.number }} ${{ github.event.pull_request.base.sha }} Modelica
+        
+  testrun_modelicatest:
+    needs: prepare
+    runs-on: self-hosted
```
beutlich:
```suggestion
    runs-on: [ self-hosted, linux, regression-testing ]
```

.github/workflows/PRTesting.yml
```diff
@@ -0,0 +1,35 @@
...
+        run: /home/resimuser/regression_testing/docker/work/run_scripts/prepare_pr.sh ${{ github.event.number }}
+  
+  testrun_modelica:
+    needs: prepare
+    runs-on: self-hosted
+    environment: 
+        name: link_modelica
+        url: https://serv.ltx.de/prs/${{ github.event.number }}/Modelica/report/PR_comparison_report.html
+    steps:         
+      - name: run modelica
```
beutlich:
```suggestion
      - name: Run tests
```

.github/workflows/PRTesting.yml
```diff
@@ -0,0 +1,35 @@
...
+        url: https://serv.ltx.de/prs/${{ github.event.number }}/Modelica/report/PR_comparison_report.html
+    steps:         
+      - name: run modelica      
+        run: /home/resimuser/regression_testing/docker/work/run_scripts/run_pr.sh ${{ github.event.number }} ${{ github.event.pull_request.base.sha }} Modelica
+        
+  testrun_modelicatest:
+    needs: prepare
+    runs-on: self-hosted
+    environment: 
+        name: link_modelicatest
```
beutlich:
```suggestion
        name: Build libraries
```

.github/workflows/PRTesting.yml
```diff
@@ -0,0 +1,35 @@
...
+        url: https://serv.ltx.de/prs/${{ github.event.number }}/ModelicaTest/report/PR_comparison_report.html
+    steps:    
+      - name: run modelicatest
```
beutlich:
```suggestion
      - name: Run tests
```

.github/workflows/PRTesting.yml
```diff
@@ -0,0 +1,35 @@
...
+    steps: 
+      - name: checkout_pr_repo
+        run: /home/resimuser/regression_testing/docker/work/run_scripts/prepare_pr.sh ${{ github.event.number }}
+  
+  testrun_modelica:
+    needs: prepare
+    runs-on: self-hosted
+    environment: 
+        name: link_modelica
+        url: https://serv.ltx.de/prs/${{ github.event.number }}/Modelica/report/PR_comparison_report.html
```
beutlich: Do not use hard-coded URIs for https://serv.ltx.de (here and in other places), instead prefer variables via https://github.com/modelica/ModelicaStandardLibrary/settings/variables/actions. This way LTX can move its server while workflow keeps untouched.

.github/workflows/PRTesting.yml
```diff
@@ -0,0 +1,35 @@
...
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  prepare:
+    runs-on: self-hosted   
+    steps: 
+      - name: checkout_pr_repo
+        run: /home/resimuser/regression_testing/docker/work/run_scripts/prepare_pr.sh ${{ github.event.number }}
```
beutlich: For the sake of transparancy, can we add all the CI bash scripts to a new folder ReSim in [.CI](https://github.com/modelica/ModelicaStandardLibrary/tree/master/.CI)? Otherwise it will be pure magic for library officers. Usually and as best practice the CI scripts are maintained along with the repo under test.

beutlich (changes requested):
The log is pretty verbose and it's not obvious what might be the reason for failure at first sight.

I have saome ideas:

1. Can we reduce the log verbosity. Can we have a configurable log verbosity.
2. Can we add problem matchers? See for example https://github.com/modelica/ModelicaStandardLibrary/blob/master/.github/moparser.json or https://oneuptime.com/blog/post/2026-01-30-github-actions-problem-matchers/view.
3. Can we upload the comparison reports using [actions/upload-artifact](https://github.com/actions/upload-artifact)?

.github/workflows/PRTesting.yml
MatthiasBSchaefer: @beutlich 
Thank you for your detailed review !

1.) I know, log is pretty verbose. But it is not really meant to be analyzed. Finding errors (like example Modelica.Blocks.Examples.xxx is not running anymore due to the changes in the PR)  can be done in the published reports. The log is more or less to analyze possible errors in the workflow itself by us (LTX). We can reduce the logging once we are sure the workflow is stable. (without any changes inside this repository)

2.) probably we could introduce problem matchers, but may not be necessary due to 1.)

3.) We discussed this in the MAP-LIB meetings. Upload artifacts would induce the need to download the reports (which are not that small) , unzip and then open in a browser. These steps, we want to avoid, making the reports directly clickable by publishing on our server.

4.) Thank you for the very good hint with the variables instead of hard links!

5.)We can publish the main shell scripts called by the CI. But these are mainly running the Modelica tools inside docker containers. And we wont publish our regression testing tool itself, nor the docker files, due to intellectual property.  So publishing the scripts is only partially helpful concerning transparency, but would induce a change of these scripts (and such a change in the MSL) in order to add a further Modelica tool (e.g. Simulation X) to the regression testing. 

6.) Your proposed naming of the jobs is fine for me.

.github/workflows/PRTesting.yml
MatthiasBSchaefer: The enviroment is meant to create\show the links to the reports.
So "Report modelicatest" might be a good name

MatthiasBSchaefer:
As requested, i am now:
-  using the "secrets and variables" feature of github
- reduced the logging in order to have a quick overview there 
- published the principal scripts 

Please have a look at
https://github.com/MatthiasBSchaefer/ModelicaStandardLibrary/actions/runs/25574891439
to see an example run. 
(In this case i only added some whitespaces, so not differences in simulation success and trajectory comparison is expected)

.github/workflows/ReSim_CI.yml
```diff
@@ -0,0 +1,43 @@
...
+    
+  testrun_modelica:
+    needs: prepare
+    runs-on: [ self-hosted, Linux, regression_testing ]
+    environment: 
+        name: Report modelica
+        url: ${{ env.PR_SERVER }}/${{ env.EVENT_NUMBER }}/Modelica/report/PR_comparison_report.html
+    steps:
+      - name: Run tests modelica      
+        run: ./.CI/ReSim/run_pr.sh ${{ env.EVENT_NUMBER }} ${{ env.PR_SHA }} Modelica ${{ env.TESTING_TOOLS }}
```
AnHeuermann, line 32: Test job is missing a checkout step.
This would only work on self-hosted runners where workspace files persist (generally a bad thing).  There seems to be no sanitation of the VM / sandbox in between jobs. On the GitHub runners you would need to use artifacts between jobs.

.github/workflows/ReSim_CI.yml
```diff
@@ -0,0 +1,43 @@
...
+
+            
+  testrun_modelicatest:
+    needs: prepare
+    runs-on: [ self-hosted, Linux, regression_testing ]
+    environment: 
+        name: Report modelicatest
+        url: ${{ env.PR_SERVER }}/${{ env.EVENT_NUMBER }}/ModelicaTest/report/PR_comparison_report.html
+    steps:
```
AnHeuermann, line 41: Test job is also missing a checkout step, see comment above.

.github/workflows/ReSim_CI.yml
```diff
@@ -0,0 +1,43 @@
+name: CI ReSim
+
+on:
+  pull_request_target:
+  workflow_dispatch:
```
AnHeuermann, line 5: `workflow_dispatch` will probably fail due to missing context.

When triggered via `workflow_dispatch`, `github.event.number` and `github.event.pull_request.base.sha` are undefined, causing `EVENT_NUMBER` and `PR_SHA` to be empty. 

```diff
on:
   pull_request_target:
   workflow_dispatch:
+    inputs:
+      pr_number:
+        description: 'PR number'
+        required: true
+        type: number
+      base_sha:
+        description: 'Base commit SHA'
+        required: true
+        type: string
```

This also needs an update on the next lines:

```diff
 env:
     PR_SERVER: ${{ vars.LTX_PR_SERVER }}
     TESTING_TOOLS: ${{ vars.LTX_TESTING_TOOLS }}
-    EVENT_NUMBER:  ${{ github.event.number }}
-    PR_SHA: ${{ github.event.pull_request.base.sha }}
+    EVENT_NUMBER:  ${{ github.event.number || inputs.pr_number }}
+    PR_SHA: ${{ github.event.pull_request.base.sha || inputs.base_sha }}
```

.CI/ReSim/prepare_pr.sh
```diff
@@ -0,0 +1,7 @@
+event_number=$1
```
AnHeuermann, line 1: It's a common practice to add a shebang and to validate arguments in all shell scripts. Also using `set -euo pipefail` is a good way to ensure that your script fails loud, see https://gist.github.com/imrekoszo/eee95e94389e168a2dffe42bbc058a47.

```bash
#!/bin/bash
set -euo pipefail

if [ -z "${1:-}" ]; then
  echo "Error: event_number argument is required" >&2
  exit 1
fi
```

Personally I would also add `set -x` to echo the commands. Makes debugging later on way easier.
And I'm a fan of self documenting scripts, so I would suggest to add a help functions that also severs as the documentation for this script.

.CI/ReSim/run_pr.sh
```diff
@@ -0,0 +1,33 @@
...
+fi
+if [ ! -d "/var/www/html/prs/$event_number" ]; then
+    mkdir /var/www/html/prs/$event_number
+fi
+mkdir /var/www/html/prs/$event_number/$pkg_name
+
+# copy data to webserver directory
+docker cp create_overview_$event_number\_$pkg_name:/shared_data/resim_output/PRs/PR_$event_number/$pkg_name/report  /var/www/html/prs/$event_number/$pkg_name
+docker rm create_overview_$event_number\_$pkg_name
+exit $(head -n 1 /var/www/html/prs/$event_number/$pkg_name/report/status.txt)
```
AnHeuermann, line 33: From @CodeRabbitAI:

Add error handling and validate exit status.

The script assumes status.txt exists and contains a valid numeric exit code. If the file is missing, empty, or contains non-numeric content, the exit command will fail or produce incorrect results. Additionally, if `docker cp` or `docker rm` fails, the script continues without reporting the error.

```suggestion
docker cp create_overview_${event_number}_${pkg_name}:/shared_data/resim_output/PRs/PR_${event_number}/${pkg_name}/report "/var/www/html/prs/${event_number}/${pkg_name}" || {
  echo "Error: Failed to copy report from container" >&2
  exit 1
}
docker rm create_overview_${event_number}_${pkg_name}

status_file="/var/www/html/prs/${event_number}/${pkg_name}/report/status.txt"
if [ -f "$status_file" ]; then
  status=$(head -n 1 "$status_file")
  if [[ "$status" =~ ^[0-9]+$ ]]; then
    exit "$status"
  else
    echo "Error: Invalid status in $status_file: $status" >&2
    exit 1
  fi
else
  echo "Error: Status file not found: $status_file" >&2
  exit 1
fi
    mkdir /var/www/html/prs/$event_number
fi
mkdir /var/www/html/prs/$event_number/$pkg_name
```

.CI/ReSim/run_pr.sh
```diff
@@ -0,0 +1,33 @@
+event_number=$1
```
AnHeuermann, line 1: See previous comment.

```suggestion
#!/bin/bash
set -euo pipefail

if [ $# -lt 4 ]; then
  echo "Error: requires 4 arguments: event_number master_hash pkg_name tools" >&2
  exit 1
fi
event_number=$1
```

AnHeuermann (changes requested):
I did run [CodeRabbit](https://www.coderabbit.ai/) on this PR and it pointed out multiple critical issues.

I would suggest to:

1. See my mail.
2. Replace shared volumes
    * Use artifacts (+ `docker save` / `docker load)` to pass data from one job to the next.
3. Use GitHub runners instead of self-hosted runners wherever possible to minimize attack surface.
    * Split all tools into separate jobs to enable parallelization (some use GitHub runners, some self-hosted runners)
    * Make Docker images available to GitHub runners (use GHCR or a private registry and authenticate with secrets)
4. Use artifacts to publish results
    * Use artifacts to upload to a self-hosted web-server or use something like GitHub pages

beutlich (changes requested):
Image

Image

This needs fixing.

PR #4769, opened 2026-03-23
added debye3 function for eventual T-dependent heat capacities

tburch-LSS:
3rd order Debye function has been included to the Special functions. This will be useful for expanding the Thermal library to include temperature-dependent heat capacities based upon the Debye model.

### DIFF
--- a/Modelica/Math/Special.mo
+++ b/Modelica/Math/Special.mo
@@ -426,6 +426,89 @@ sinc(3)   // = 0.0470400026866224
 </pre></blockquote>
 </html>"));
   end sinc;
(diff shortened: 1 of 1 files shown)


tburch-LSS:
i should add that i tested the functionality against the same function from the Gnu Scientific Library:
Debye3_fracerr
Debye3_test

tburch-LSS:
another thought: i recognize that this may be a special function with a rather limited scope and that it may not be an optimal implementation. it could therefore be moved to a section like Thermal.HeatTransfer.Utilities.

PR #4779, opened 2026-04-26
Output interval less dense to reduce size of ReferenceResult

AHaumer:
Factor 100 is ok for the trajectories and should reduce size of ReferenceResults sufficiently. 
I suppose the dense output slipped in during development starting from an example with more intense transients.


christiankral (approved):
This makes a lot of sense

maltelenz (approved):
Looks reasonable.

PR #4787, opened 2026-05-03
Prepare for CI: Magnetic.QuasiStatic.FundamentalWave

AHaumer:
same procedure as #4775, except Modelica.Magnetic.QuasiStatic.FundamentalWave.Examples.BasicMachines.InductionMachines.IMC_withLosses where IMHO a longer interval (factor 100) is sufficient to reduce the size of the reference results.


christiankral:
@AHaumer Some "new" models have been added to ModelicaTest. Why can't we just extend these models from the MSL and add a proper annotation? They look quite identical to me.

christiankral:
Extending from the Modelica Standard Library make the code very maintainable. One more aspect here:

- The annotations in ModelicaTest now show simulation start and stop time and indicate the time window for comparison. This is OK.
- The same annotation also shows the documentation of original model stored in the MSL. This causes some obvious conflicts. Example:

```modelica
 extends Modelica.Magnetic.QuasiStatic.FundamentalWave.Examples.BasicMachines.SynchronousMachines.SMPM_CurrentSource;
  annotation (
    experiment(StopTime=0.20, Interval=1E-4, Tolerance=1E-6),
    TestCase(shouldPass = true,
      __ModelicaAssociation(Comparison(TimeWindows={TimeSlot(0.00, 0.20)}))),
    Documentation(info="<html>
<p>
This example compares a time transient and a quasi-static model of a permanent magnet synchronous machine. The machines are fed by a current source. The current components are oriented at the magnetic field orientation and transformed to the stator fixed reference frame. This way the machines are operated at constant torque. The machines start to accelerate from standstill.</p>

<p>
Simulate for 2 seconds and plot (versus time):
</p>
```

The stop time is set to 0.20 seconds and the documentation states "Simulate for 2 seconds".

I wonder if it makes more sense to just state something like:

```
    Documentation(info="<html>
<p>
The original documentation is available at the model from which this one is extended.
</p>
```

Reasons: 

- Who is maintaining the documentation of the ModelicaTest models extended from the MSL? I guess no one. So whenever we update the documentation of the MSL model, it is entirely forgotten about the documentation of the ModelicaTest models.
- However, we could add a hyperlink to the original model of the MSL. But this is not required from my point of view, as the `extends` clearly shows which model we refer to.

henrikt-ma:
> * The annotations in ModelicaTest now show simulation start and stop time and indicate the time window for comparison. This is OK.

I don't see this being OK regarding the `StartTime`, as this may have all sorts of unwanted effects compared to just trimming the first part of the main model's simulation result.

To make time windowed comparison beyond just taking the initial part of the result of a chaotic system, we need this `TestCase` annotation that we have talked about, where the time intervals have both beginning and end.

AHaumer:
@christiankral I adreesed your concerns. Sorry for not using extends, my fault.
@henrikt-ma I don't understand your comment. 
All the "extended" examples in ModelicaTest use the same default StartTime=0 as the original models.

ModelicaTest/Resources/Reference/ModelicaTest/Magnetic/QuasiStatic/FundamentalWave/Examples/BasicMachines/InductionMachines/IMC_Conveyor/comparisonSignals.txt
```diff
@@ -0,0 +1,12 @@
+time
+imc.rotorCage.electroMagneticConverter.singlePhaseElectroMagneticConverter[1].Phi.re
+imc.rotorCage.electroMagneticConverter.singlePhaseElectroMagneticConverter[2].Phi.im
+imc.stator.electroMagneticConverter.singlePhaseElectroMagneticConverter[1].Phi.re
+imc.stator.electroMagneticConverter.singlePhaseElectroMagneticConverter[2].Phi.im
+// imc.stator.zeroInductor.i0
```
maltelenz, line 6: Maybe we can take the opportunity to get rid of old commented out signals, unless we have a reason to keep them?
```suggestion

```
(and similarly for the other cases of this)

AHaumer: Dear @maltelenz I tend to keep them to have a reminder not to re-introduce these signals.
Originally we wanted to compare these signals but we ran into problems during comparison.
This is the zero-component of the Clarke-transformation and in most cases it should be zero 
but in many cases it shows small numerical noise.

maltelenz: @AHaumer Fine by me, thanks for explaining the reasoning!

henrikt-ma:
> @henrikt-ma I don't understand your comment. All the "extended" examples in ModelicaTest use the same default StartTime=0 as the original models.

I was commenting on the following (emphasis is mine):
> The annotations in ModelicaTest now show simulation **start** and stop time and indicate the time window for comparison. This is OK.

That is, if you remove the _start and_, and maybe stop calling it a _time window_ when the window must be open-eded to the left, then it would seem OK to me as well.

I would rather call what you are doing something like a _short-time comparison model_ to not cause confusion with a proper time-windowed comparison which we seem to need sooner or later anyway.

HansOlsson:
> I wonder if it makes more sense to just state something like:
> 
> ```
>     Documentation(info="<html>
> <p>
> The original documentation is available at the model from which this one is extended.
> </p>
> ```
> 
> Reasons:
> 
> * Who is maintaining the documentation of the ModelicaTest models extended from the MSL? I guess no one. So whenever we update the documentation of the MSL model, it is entirely forgotten about the documentation of the ModelicaTest models.
> * However, we could add a hyperlink to the original model of the MSL. But this is not required from my point of view, as the `extends` clearly shows which model we refer to.

I agree that we shouldn't duplicate the documentation, but to me it seems simplest to as default not have _any_ documentation for these models.

Tools can already show that it extends from another model, and in some way make that accessible.

In Dymola that would currently give something like:

> **Information**

>
>
> Extends from [Modelica.Magnetic.QuasiStatic.FundamentalWave.Examples.BasicMachines.SynchronousMachines.SMPM_CurrentSource](https://doc.modelica.org/Modelica%204.1.0/Resources/helpDymola/Modelica_Magnetic_FundamentalWave_Examples_BasicMachines_SynchronousMachines.html#[Modelica.Magnetic.FundamentalWave.Examples.BasicMachines.SynchronousMachines](https://doc.modelica.org/Modelica%204.1.0/Resources/helpDymola/Modelica_Magnetic_FundamentalWave_Examples_BasicMachines_SynchronousMachines.html#Modelica.Magnetic.FundamentalWave.Examples.BasicMachines.SynchronousMachines.SMPM_CurrentSource).SMPM_CurrentSource)
> (Test example: PermanentMagnetSynchronousMachine fed by current source).

(But with an internal link instead.) 

Thus, I don't see that spending time on writing "The original documentation is available at the model from which this one is extended." really helps.

I could even imagine that tools could be more helpful if the documentation of the derived class is empty, so it might be that the text is counter-productive.

ModelicaTest/Magnetic/QuasiStatic/FundamentalWave/Examples/BasicMachines/InductionMachines/IMC_Initialize.mo
```diff
@@ -0,0 +1,25 @@
...
+      Interval=0.0001,
+      Tolerance=1e-06),
+    TestCase(shouldPass = true,
+      __ModelicaAssociation(Comparison(TimeWindows={TimeSlot(0.45, 0.60)}))),
+    Documentation(
+        info="<html>
+<p>
+The original documentation is available at the model from which this one is extended.
+</p>
+</html>"),
```
HansOlsson:
```suggestion
```
No need for documentation to say that it extends from something.

ModelicaTest/Magnetic/QuasiStatic/FundamentalWave/Examples/BasicMachines/InductionMachines/IMC_Transformer.mo
```diff
@@ -0,0 +1,22 @@
...
+  extends Modelica.Magnetic.QuasiStatic.FundamentalWave.Examples.BasicMachines.InductionMachines.IMC_Transformer;
+  annotation (experiment(StopTime=2.1, Interval=0.0001, Tolerance=1e-06),
+    TestCase(shouldPass = true,
+      __ModelicaAssociation(Comparison(TimeWindows={TimeSlot(2.00, 2.10)}))),
+    Documentation(
+        info="<html>
+<p>
+The original documentation is available at the model from which this one is extended.
+</p>
+</html>"),
```
HansOlsson:
```suggestion
```
No need for documentation to say that it extends from something.

ModelicaTest/Magnetic/QuasiStatic/FundamentalWave/Examples/BasicMachines/InductionMachines/IMC_YD.mo
```diff
@@ -0,0 +1,21 @@
...
+  extends Modelica.Magnetic.QuasiStatic.FundamentalWave.Examples.BasicMachines.InductionMachines.IMC_YD;
+  annotation (experiment(StopTime=2.1,Interval=0.0001,Tolerance=1e-06),
+    TestCase(shouldPass = true,
+      __ModelicaAssociation(Comparison(TimeWindows={TimeSlot(2.00, 2.10)}))),
+    Documentation(
+        info="<html>
+<p>
+The original documentation is available at the model from which this one is extended.
+</p>
+</html>"),
```
HansOlsson:
```suggestion
```
No need for documentation to say that it extends from something.

ModelicaTest/Magnetic/QuasiStatic/FundamentalWave/Examples/BasicMachines/SynchronousMachines/SMEE_Generator.mo
```diff
@@ -0,0 +1,28 @@
...
+      Interval=1E-4,
+      Tolerance=1e-06),
+    TestCase(shouldPass = true,
+      __ModelicaAssociation(Comparison(TimeWindows={TimeSlot(14.9, 15.1)}))),
+    Documentation(
+        info="<html>
+<p>
+The original documentation is available at the model from which this one is extended.
+</p>
+</html>"),
```
HansOlsson:
```suggestion
```
No need for documentation to say that it extends from something.

ModelicaTest/Magnetic/QuasiStatic/FundamentalWave/Examples/BasicMachines/SynchronousMachines/SMPM_CurrentSource.mo
```diff
@@ -0,0 +1,23 @@
...
+  annotation (
+    experiment(StopTime=0.20, Interval=1E-4, Tolerance=1E-6),
+    TestCase(shouldPass = true,
+      __ModelicaAssociation(Comparison(TimeWindows={TimeSlot(0.00, 0.20)}))),
+    Documentation(
+        info="<html>
+<p>
+The original documentation is available at the model from which this one is extended.
+</p>
+</html>"),
```
HansOlsson:
```suggestion
```
No need for documentation to say that it extends from something.

ModelicaTest/package.order
```diff
@@ -3,11 +3,11 @@ Blocks
 Tables
 Electrical
 Fluid
+Magnetic
 Media
```
HansOlsson, line 7:
```suggestion
Electrical
Magnetic
Fluid
Media
```
I understand that it looked odd to have "Magnetic" last, but I assume it is an expanded view of the Modelica-structure not a pure alphabetic order (that's why Media is before Math and Tables after Blocks), so I think Magnetic should be grouped with Electrical, not between Fluid and Media.

It could also be that it is just based on when the packages were added, but even in that case I think we should start working towards a logical order based on MSL - not an alphabetical one.

HansOlsson (changes requested):
Some minor issues, but overall it looks good.

christiankral:
> Thus, I don't see that spending time on writing "The original documentation is available at the model from which this one is extended." really helps.

> I could even imagine that tools could be more helpful if the documentation of the derived class is empty, so it might be that the text is counter-productive.

@HansOlsson @AHaumer I totally agree. The simplest and most transparent solution is just to not add any documentation and leave it up to the tool.

HansOlsson (approved):
LGTM - except for the package.order comment.

christiankral (approved):
Looks good

MatthiasBSchaefer:
Please have a look of the comparison of the adapted examples of Modelica.Magnetic.QuasiStatic.FundamentalWave with the existing MAP-LIB-ReferenceResults here: 

https://www.ltx.de/download/MA/PR_Testing/PR4787/Modelica/testrun_report.html

and the comparison of the new examples in ModelicaTest.Magnetic.QuasiStatic.FundamentalWave with the existing MAP-LIB-ReferenceResults (inside Modelica !!!)  here:
https://www.ltx.de/download/MA/PR_Testing/PR4787/ModelicaTest/testrun_report.html
 
HansOlsson:
> Please have a look of the comparison of the adapted examples of Modelica.Magnetic.QuasiStatic.FundamentalWave with the existing MAP-LIB-ReferenceResults here:
> 
> https://www.ltx.de/download/MA/PR_Testing/PR4787/Modelica/testrun_report.html
> 
> and the comparison of the new examples in ModelicaTest.Magnetic.QuasiStatic.FundamentalWave with the existing MAP-LIB-ReferenceResults (inside Modelica !!!) here: https://www.ltx.de/download/MA/PR_Testing/PR4787/ModelicaTest/testrun_report.html
> 
> Therefore i used a modified CSV-Comparison Tool ([modelica-tools/csv-compare#80](https://github.com/modelica-tools/csv-compare/pull/80), [modelica-tools/csv-compare#81](https://github.com/modelica-tools/csv-compare/pull/81))

Was there something specific we should look at, or did you just mean that we can compare and see that it worked (the test-results are mostly green and the main difference are missing variables which is intentional)?

PR #4788, opened 2026-05-20
Prepare for CI: Electrical.Machines

AHaumer:
Reducing the size of Reference Results
Similar procedure as #4775 and #4787


HansOlsson (commented):
I believe this is correct, but I would prefer input from the ones doing the CI-testing.

MatthiasBSchaefer:
The comparison of the adapted examples of Modelica.Electrical.Machines with the existing MAP-LIB-ReferenceResults here:

https://www.ltx.de/download/MA/PR_Testing/PR4788/Modelica/testrun_report.html

and the comparison of the new examples of ModelicaTest.Electrical.Machines with the existing MAP-LIB-ReferenceResults here:
https://www.ltx.de/download/MA/PR_Testing/PR4788/ModelicaTest/testrun_report.html
Please consider that this is a comparison of new examples inside ModelicaTest with the ReferenceResults of the corresponding MODELICA examples, they are extended from.

MatthiasBSchaefer (approved):
looks good to me

PR #4796, opened 2026-07-30
Fixes invalid specification of significantDigits on a String with integer argument.

modlfo:
I found a few cases inside `DynamicSelect` where the `String()` is called with an integer argument and the `significantDigits` are provided. That's invalid according to the specification.

> The standard type coercion described in [section 10.6.13](https://specification.modelica.org/master/arrays.html#standard-type-coercion) shall not be applied for the first argument of String. Hence, specifying significantDigits is an error when the first argument of String is an Integer expression.


maltelenz (approved):

HansOlsson:
Note that this restriction was added/clarified in https://github.com/modelica/ModelicaSpecification/issues/3262#issuecomment-1323889913

MatthiasBSchaefer:
What do you think about converting the integer into a real (by multiply with 1.0 ) and set significantDigits=3 ?
String( y*1.0, minimumLength=1, significantDigits=3)

Such, large integers are displayed in scientific notation
y=123456789 ==> "1.23e8"

HansOlsson:
> What do you think about converting the integer into a real (by multiply with 1.0 ) and set significantDigits=3 ? String( y*1.0, minimumLength=1, significantDigits=3)
> 
> Such, large integers are displayed in scientific notation y=123456789 ==> "1.23e8"

Not sure. I don't know how common such large integers are for these blocks, and what people want to see for such values.

However, it also impacts smaller numbers so 1234 will be written as "1.23e+03". That uses more space, and seems less readable if you work with integers than plain "1234", but I don't know how relevant that is either.

HansOlsson (approved):
Seems ok.

HansOlsson:
Looking closer I realized that the removed setting is `significantDigits=0`. That seems bad even for two-digit integers, so I think the change is good.

PR #4798, opened 2026-08-06
FluxTubes: Alternative approximation

AHaumer:
When working with FluxTubes, I found it nearly impossible to implement new materials using the approximation formula of Roschke. Additionally, when the model drives the material very high into saturation ("overfluxed region"), the behaviour is not always as expected (e.g. polarization J doesnt show a horizontal asymptote but decreases with increasing field strength). Unfortunately the original authors are no longer available. 
So I decided to implement an alternative approximation: Smoothing Splines with Exponential Extrapolation.
Everything is explained in the documentation, and there will be a contribution to the Modelica 2027 conference. 
Additionally, VoestAlpine allowed me to use their measured data of fully processed electric sheet (as used in rotating electric machines). So I was able to add 10 material records. The new parameter records contain the used raw data, so the approximation is traceable. It should be easier for users to implement their own material. 
Additonally, 4 new examples have been implemented to demonstrate and test the new approximation thoroughly. 
You might call it an enhancement, but it solves the users' problem of implementing new material records. 
The solution is 100% backwards-compatible, and has been tested with (in alphabetical order) Dymola, OpenModelica and WolframSystemModeler.


AHaumer:
One of the original authors, Thomas Boedrich, wrote to me:
Unfortunately we (he and Johannes Ziske) do not have the ressources to review that PR, but we fully agree with you that this is a valuable and important enhancement.

AHaumer:
I have a set of functions to determine the approximation parameters from raw data.
These functions run both in Octave and Matlab. 
Should I provide the code (and the raw data files)? In Ressources/xxx?

Since I get no answer:
I provided the functions in the [AppMag-repository](https://github.com/AHaumer/AppMag) together with the used raw data RD.ods in OpenDocument format as well as the BaseData Modelica record and stated that fact in the documentation.

tobolar:
> I provided the functions in the [SSEE-repository](https://github.com/AHaumer/SSEE) ...

I can not reach that repository.

AHaumer:
> > I provided the functions in the [SSEE-repository](https://github.com/AHaumer/SSEE) ...
> 
> I can not reach that repository.

Sorry @tobolar I had to change the repo: [AppMag-repository](https://github.com/AHaumer/AppMag)
But this PR is mainly about the FluxTubes, the preprocessing is only an additional feature.

tobolar:
@AHaumer No problem, I just wanted to point out the repo is not accessible.

Modelica/Magnetic/FluxTubes/BaseClasses/FixedShape.mo
```diff
@@ -1,42 +1,60 @@
...
+  Modelica.Units.SI.MagneticFluxDensity B(start=0) "Magnetic flux density";
+  Modelica.Units.SI.MagneticPolarization J "Magnetic polarization";
+  Modelica.Units.SI.MagneticFieldStrength H(start=0) "Magnetic field strength";
+  Modelica.Units.SI.RelativePermeability mu_r(start=mu_rConst)
+    "Relative magnetic permeability";
 protected
-  Real B_N "Absolute value of normalized B";
-
+  Real B_N=abs(B/material.B_myMax) "Absolute value of normalized B";
+  constant Real epsilon=1e3*Modelica.Constants.eps;
```
HansOlsson:
```suggestion
```
As far I could see epsilon was newly introduced and not used, I would thus recommend to not introduce it.

(Slowly going through mail-back-log.)

Modelica/Magnetic/FluxTubes/BaseClasses/FixedShape.mo
```diff
@@ -1,42 +1,60 @@
 within Modelica.Magnetic.FluxTubes.BaseClasses;
 partial model FixedShape "Base class for flux tubes with fixed shape during simulation; linear or non-linear material characteristics"
+  import Modelica.Magnetic.FluxTubes.Types.MagType;
+  import Modelica.Magnetic.FluxTubes.Material.SoftMagnetic.SSEE.Functions.app_J;
+  import Modelica.Magnetic.FluxTubes.Material.SoftMagnetic.SSEE.Functions.app_mu_r;
 
-  extends Interfaces.TwoPort;
-
+  // Material
+  // This parameter is kept for backwards compatibility reasons, might be replaced by a member "Linear" in the enumeration Magnetization
```
HansOlsson: I don't like this "might be replaced", and I see two possibilities:
1. State that it is kept for backwards compatibility, and will be replaced by member "Linear" in next major version.
2. Additionally already _now_ add "Linear" in the enumeration and change the if-statement to use something like: `nonLinearPermeability and not magType==MagType.Linear`

AHaumer: I prefer 1. - done and resolved

HerbertSchmidt75:
This method appears to be more flexible than the one used before and has the potential to yield better fits than the one used before (since it does not force a simple equation upon the data). Neither of them is physical in the way that first-principles functions are used. Both "only" try to fit the data while not contradicting physical constraints (like a constant saturation in J).

One constraint that I have seen enforced for ferromagnetics (and the Roschke-equation satisfies that to my eye) is mu_r>1 for all H. The SSEE method does not consider this to my understanding, it just follows the data. Now, physical data for ferromagnetics will respect this constraint, however, by not enforcing it, a situation with mu<1 might theoretically occur from fitting in an area, where mu is close to one (but physically positive).

Physically, this means the behaviour of an air gap is reversed (reluctance in the material is higher than in the surrounding medium, rather than other way round). This may lead to interesting artefacts.

This is a rather remote scenario, since most likely the area with mu close to one will be covered by the exponential bit in the model. My question is: should we make absolutely certain, that this problem cannot occur, or are we happy with very most likely it will not occur?

AHaumer:
@HerbertSchmidt75 thanks for your advice - just some remarks from my point of view:
The "old" approximation will still be available, at least for backwards compatibility reasons.
The scenario mu_r < 1 might happen for *really* buggy measurements, even in the region of high field strength if the measured polarization decreases with increasing field strength instead of approaching saturation asymptotically. 
But to use such a material, it is necessary to pre-process the data and determine the parameters. 
During this process [AppMag](https://github.com/AHaumer/AppMag) figures are shown which reveal such a bad behaviour.

AHaumer:
pinging potential reviewers @christiankral @casella @HansOlsson @henrikt-ma @maltelenz @MartinOtter @ThomasBoedrich @HerbertSchmidt75 
If we want to store the [preprocessing tool](https://github.com/AHaumer/AppMag) in another repo belonging to MA we can decide that later.

HansOlsson:
Just a minor note: Modelica.Magnetic.FluxTubes.BaseClasses.FixedShape now has an input `A` instead of a normal variable `A`. I believe that makes sense, but in theory it breaks backwards compatibility (in case someone had their own models inheriting from FixedShape).

However, I believe it is ok - as requiring a user to set a non-input `A` was just confusing and I don't think any actual models are impacted; so we should just document it.

I'll leave the actual magnetics part to experts.

AHaumer:
@HansOlsson you're right.
But it's the same with non-input `G_m` which is also calculated in models extending from FixedShape.
Therefore for backwards compatibility reasons I reverted `A` to a non-input.

PR #4803, opened 2026-09-07
Remove duplicate entry "Magnetic" in package.order of ModelicaTest

MatthiasBSchaefer:
Removed duplicate 'Magnetic' entry from package order.


maltelenz (approved):

AHaumer (approved):
good.

PR #4804, opened 2026-09-07
Remove duplicate from package.order

sjoelund:


AHaumer:
Already solved by #4803 
Anyway - thanks for being careful @sjoelund !

PR #4805, opened 2026-09-08
Added workflow file to trigger CI for testing pull requests 

MatthiasBSchaefer:
This workflow acts only as a trigger.
The regression testing for pull requests itself will be performed in a private repository.
This is neccessary to avoid potential security issues and keep the use of commercial tools private.
The workflow is described in https://github.com/modelica/MA-Internal/blob/master/MA-Projects/MAP-LIB/MeetingMinutes/20260714-Monthly/20260714_RegressionTestingCI_Workflow.pdf
As a result, the workflow will approve the pull request or request changes (including links to reports)

### DIFF
--- a/.github/workflows/CI_ReSim_trigger.yml
+++ b/.github/workflows/CI_ReSim_trigger.yml
@@ -1,0 +1,38 @@
+name: Sync_CI_Repo
+
+on: pull_request_target
+
+permissions:
+  contents: read
+  
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    steps: 
+      - uses: actions/github-script@v9
+        with:
+          github-token: ${{ secrets.LTX_SIMULATION_TOKEN }}
+          script: |
+            const prNumber = context.payload.pull_request?.number ?? context.payload.number;
+
+            const { data: pr } = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: prNumber,
+            });
+
+            const labels = (pr.labels || []).map(l => typeof l === 'string' ? l : l.name);
+            const hasLabel = labels.includes('c-code-checked');
+
+            await github.rest.repos.createDispatchEvent({
+              owner: '${{ secrets.PRIVATE_CI_OWNER }}',
+              repo: '${{ secrets.PRIVATE_CI_REPO }}',
+              event_type: 'CI_trigger_type',
+              client_payload: {
+                number: prNumber,
+                head_sha: pr.head.sha,
+                base_sha: pr.base.sha,
+                has_label: hasLabel,
+                github_repo: context.repo.full_name
+              }
+            });
--- a/.github/workflows/CI_ReSim_trigger.yml
+++ b/.github/workflows/CI_ReSim_trigger.yml
@@ -22,8 +22,9 @@ jobs:
             });
 
             const labels = (pr.labels || []).map(l => typeof l === 'string' ? l : l.name);
-            const hasLabel = labels.includes('c-code-checked');
-
+            const hasLabel = labels.includes('ready_for_ci');
+            const author = context.payload.pull_request.user.login
+            
             await github.rest.repos.createDispatchEvent({
               owner: '${{ secrets.PRIVATE_CI_OWNER }}',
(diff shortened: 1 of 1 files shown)


AHaumer:
Since I have no experience with git workflows I'd be glad about reviews from colleagues that have deeper knowledge.
Really a pity that @beutlich refuses to review. Feel free to invite other colleagues for review.

HansOlsson:
I have tried to look at it based on the GitHub documentation, and to me it seems good with some minor comments:
- It runs for all pull requests, should we limit it to pull requests targeting master and specific maintenance-branches (as the CI cannot run for earlier maintenance levels, right?)
- Regarding hasLabel, there seems to be two completely different styles for checking labels, this is one of them, and it should work.

However, I'm not an expert regarding the security implications (even if seems they are more of an issue for the LTX-machines).
So, either:
* I click "accept" with that caveat
* Since no-one here seems willing to say that they are an expert we hire someone externally to evaluate it

.github/workflows/CI_ReSim_trigger.yml
```diff
@@ -0,0 +1,40 @@
+name: Sync_CI_Repo
+
+on: pull_request_target
```
AnHeuermann, line 3: Add a `types` filter so adding `ready_for_ci` lable is also picked up. Otherwise developers need to push a new commit to trigger this.

Default events: `opened`, `synchronize`, `reopened`
It does not include `labeled`/`unlabeled`

.github/workflows/CI_ReSim_trigger.yml
```diff
@@ -0,0 +1,40 @@
...
+
+permissions:
+  contents: read
+  
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    steps: 
+      - uses: actions/github-script@v9
```
AnHeuermann, line 12: Nitpick: I prefer to fix actions with a full-length commit SHA.

There is also an option in the organization/repo settings to enforce this. See https://docs.github.com/en/actions/reference/security/secure-use?learn=getting_started#using-third-party-actions for more details.

AnHeuermann, line 12: Add a [dependabot](https://docs.github.com/en/enterprise-cloud@latest/code-security/tutorials/secure-your-dependencies/dependabot-quickstart) to keep the used actions up to date.

AnHeuermann (approved):
LGTM!
No obvious unsafe stuff like leaking secrets to a fork PR.

Dependabot is worth a look. Otherwise someone needs to keep actions up to date manually.

.github/workflows/CI_ReSim_trigger.yml
HansOlsson: After reviewing some security guidelines, I agree with this nitpick.
(We could add the readable version as a comment.)

AnHeuermann: Yes, usually you write it like

```yml
    - name: checkout
      uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
```

This way Dependabot will pick them up as well and update both the SHA and the comment.

MatthiasBSchaefer: I think we should generalize this to all workflows and then activate dependabot.
The update is done manually till now e.g. [here](https://github.com/modelica/ModelicaStandardLibrary/commit/79e34fd2ce848477751a5e4b69455bff5053fc96)

MatthiasBSchaefer:
@AnHeuermann , @HansOlsson 
Thanks for your suggestions.
The latest commit includes:
- only testing PRs targeting master or maintainace branches
- trigger workflow also on labeling the PR (with ready_for_ci label)
- full sha for github_script workflow

.github/workflows/CI_ReSim_trigger.yml
MatthiasBSchaefer: see commit [1a266a0e](https://github.com/modelica/ModelicaStandardLibrary/pull/4805/commits/1a266a0ef303d3af39fe9285d5762c37894affce)

MatthiasBSchaefer: i will take care of dependabot, including the other workflows of MSL in a seperate PR

PR #4806, opened 2026-09-08
move reference results from MAP-LIB_ReferenceResults 

MatthiasBSchaefer:
move reference results from MAP-LIB_ReferenceResults  repo ( https://github.com/modelica/MAP-LIB_ReferenceResults/commit/c7509a7af1abf2492b38b6b95ff077cfe65f7047 )  into ModelicaStandardLibrary/ReferenceResults


AHaumer:
as discussed in today's MAP-Lib monthly meeting

HansOlsson (approved):
Looks good to me.
Since it is a separate directory it is possible to check out the rest and skip it.

AHaumer (approved):
ok it's about purely moving the ref.results in a separate top-level directory.
Can't check all lines (e.g. whether someting is missing!) but in general it seems ok to me.

PR #4810, opened 2026-09-09
Added missing Evaluate = true to sequence_start in Internal/InitAngle

### LINKED ISSUE #4809: Missing Evaluate = true in MultiBody.Joints.Internal.InitAngle
In the MultiBody library, all declarations of `Modelica.Mechanics.MultiBody.Types.RotationSequence sequence_start` have an `Evaluate = true` annotation. The reason is that evaluating this parameter allows for essential symbolic simplifications, while at the same time there is really no user need to change this parameter at runtime without recompiling the model.

Unfortunately, probably due to some oversight, `Evaluate = true` is actually missing in the declaration of `sequence_start` in one model, namely `Modelica.Mechanics.MultiBody.Joints.Internal.InitAngle`. 

This is discussed in depth in this [comment](https://github.com/OpenModelica/OpenModelica/issues/16360#issuecomment-5397237560). Long story short:
- `MultiBody.Joints.Internal.InitAngle` is only instantiated in one place in the library inside `Modelica.Mechanics.MultiBody.Joints.FreeMotionScalarInit`
https://github.com/modelica/ModelicaStandardLibrary/blob/99f004e92c43904e4c444383bf87db0c86595f54/Modelica/Mechanics/MultiBody/Joints/FreeMotionScalarInit.mo#L176
- the `sequence_start` parameter of `FreeMotionScalarInit` mode has `Evaluate = true` but is passed to the `InitAngle` model without a `final` modifier
- the `sequence_start` parameter of `InitAngle` does not have `Evaluate = true`
https://github.com/modelica/ModelicaStandardLibrary/blob/99f004e92c43904e4c444383bf87db0c86595f54/Modelica/Mechanics/MultiBody/Joints/Internal/InitAngle.mo#L8-L9

Hence, a compiler that wants to maximize the possibility for end-users to change parameters at runtime without recompiling will not evaluate `sequence_start` inside `InitStart`, because it doesn't have a final binding and it has no `Evaluate = true` annotation.

This is not a good idea, because it unecessarily leads to huge complex expressions that need to be handled symbolically.

We should then add `Evaluate = true` to the declaration of `sequence_start` also within `Internal.InitAngle` as we already do in all other similar cases. There is really no reason not to do it.

casella:
Fixes #4809.

### DIFF
--- a/Modelica/Mechanics/MultiBody/Joints/Internal/InitAngle.mo
+++ b/Modelica/Mechanics/MultiBody/Joints/Internal/InitAngle.mo
@@ -6,7 +6,8 @@ model InitAngle
   import Modelica.Mechanics.MultiBody.Frames;
 
   parameter Modelica.Mechanics.MultiBody.Types.RotationSequence sequence_start={1,2,3}
-   "Sequence of angle rotations";
+   "Sequence of angle rotations"
+   annotation(Evaluate = true);
 
   Interfaces.Frame_a frame_a
     annotation (Placement(transformation(extent={{-116,-16},{-84,16}})));


tobolar (approved):
Fine to me.

